PR #91, opened 2014-02-08
Add missing required property names to errors

primitive-type:
In cases where required properties are missing, the `property` keys in `$this->errors` should contain the names of the missing properties rather than an empty string. This would help a lot for parsing purposes.


bighappyface:
@primitive-type if you are still interested in this contribution, would you please rebase against the latest?

bighappyface:
Closed with https://github.com/justinrainbow/json-schema/pull/154

PR #96, opened 2014-03-07
Update minimum PHP version to 5.3.2

siwinski:
`bin/validate-json` uses `stream_resolve_include_path()` which is only available in PHP >= 5.3.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         }
     }],
     "require": {
-        "php": ">=5.3.0"
+        "php": ">=5.3.2"
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",


bighappyface:
We may as well bump it to 5.3.3

siwinski:
Bumped to 5.3.3 (note that this is the version used by RHEL 6 so hopefully min PHP version won't go beyond 5.3.3)

GrahamCampbell:
:-1: Composer depends on this package.

siwinski:
I'm fine with 5.3.2 but was requested to change it to 5.3.3.  Whichever the maintainers prefer I will make it so.

siwinski:
Rebased and repushed.

siwinski:
Re-based to current master and to reverted commits to original pull request of min PHP version of 5.3.2

GrahamCampbell:
:+1:

GrahamCampbell:
Also, composer itself depends on php 5.3.2+, so you literally can't install this package on php 5.3.0...

GrahamCampbell:
Could we get this merged?

bighappyface:
+1

PR #111, opened 2014-09-24
added property callback mechanism

daarond:
Registered properties, when encountered by the validator, fires the callback function.

Usage:
```
$validator->register_property('db_uniqueness',function($key, $element, $schema, $path, $i){
    // callback magick here

    // callback determines there are errors, so return them
    return array($key=>'random error 1', $key=>'specific error');
});
$validator->check($obj, $schema);
```

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -114,7 +114,7 @@ abstract class Constraint implements ConstraintInterface
      * @param $i
      */
     protected function call_user_properties($schema, $element, $path, $i){
-        if (count($GLOBALS['Validator_Properties'])>0){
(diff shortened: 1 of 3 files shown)


bighappyface:
-1 for use of `$GLOBALS` and breaking the method naming convention

daarond:
Your comment makes no sense. call_user_properties already uses $GLOBALS, so I used existing structures. I'm good with changing the method name to whatever naming convention.

bighappyface:
@daarond please consider your responses before you write them.

Here is the current master version of Constraint.php: https://github.com/justinrainbow/json-schema/blob/1235d2e2100545e775512f9c88cbb0b680245a24/src/JsonSchema/Constraints/Constraint.php

Here, you will notice `$GLOBALS` is nowhere to be found. That is because you wrote `call_user_properties` as part of your PR.

You did not use existing structures and you are breaking even basic OOP best practices, which is why I gave you a -1.

PR #117, opened 2014-10-23
Added missed symbol for query string.

pasho12:
Without it uri like "$ref":"http://hostname/some/path?some=param" resolves into "http://hostname/some/pathsome=param".


bighappyface:
@pavovs would you please include a unit test for this?

https://github.com/justinrainbow/json-schema/blob/master/tests/JsonSchema/Tests/Uri/UriResolverTest.php

bighappyface:
Closed due to age

PR #110, opened 2014-09-12
Resolving inner definitions on $ref's

stoiev:
Trying to resolve issue https://github.com/justinrainbow/json-schema/issues/93

PR #119, opened 2014-11-20
Fix #93 ($ref to local definition not working)

### LINKED ISSUE #93: $ref to local definition not working
Given the following reference:
```
(...)
"children": {
    "type": "array",
    "items": {
        "type": "object",
        "oneOf": [
            { "$ref": "#/definitions/Node" }
        ]
    }
}
(...)
```
The validator fails with:
```
Fragment "/definitions/Node" not found in file:///path/to/current/location#/definitions/Node
```

Expected: The validator looks for the definition of Node inside the same schema file, under the property "definitions".

co3k:
See also #110. This pull request is based @stoiev's branch. First, thanks for his great work!
(I skipped fc947a6c949af5302d81e8b3b409d5ead76bc0ca commit. Because it is not related with this problem and it contains some problems; composer.lock should not be ignored (see http://blog.naderman.de/2014/02/17/replace-conflict-forks-explained/) and netbeans project directories  should be ignored in $HOME/.gitignore)

And I've referenced some implementation of brandur/json_schema (it is Ruby JSON Schema validation library). I really appreciate to the library.

My fix resolves some problems of #110; e.g. the #110 implementation only resolves the top level of $ref but my implementation resolves any level of a path.

But, however, it doesn't pass json/tests/draft4/ref.json yet. We should effort resolve this problem too, but the inline dereferencing problem is needed to fix ASAP.

### DIFF
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -244,7 +244,7 @@ class RefResolver
             return $data;
         }
 
-        $key = array_shift($pathParts);
+        $key = $this->transformKey(array_shift($pathParts));
 
         return $this->resolveRefSegment(is_array($data) ? $data[$key] : $data->$key,  $pathParts);
     }
@@ -260,5 +260,13 @@ class RefResolver
         $this->uriRetriever = $retriever;
 
         return $this;
+    }
+
+    protected function transformKey($key)
+    {
+        return strtr($key, [
+            '~1' => '/',
+            '~0' => '~',
+        ]);
     }
 }
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -41,7 +41,7 @@ class RefResolver
     protected $uriRetriever = null;
     
     
-    protected $innerDefinitions = array();
+    protected $rootSchema;
     
 
     /**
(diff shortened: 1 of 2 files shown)


bighappyface:
@co3k Both "ref" tests provided by the [schema test suite](https://github.com/json-schema/JSON-Schema-Test-Suite/tree/develop/tests/draft4) are already included:

https://github.com/co3k/json-schema/blob/2450d219d374861f734ff559d876bdcb19367793/tests/JsonSchema/Tests/Drafts/Draft4Test.php#L21-L22

Is everything already passing without `testInnerDefinitions`? It would be preferable to rely on the test suite only for verification.

bighappyface:
@co3k I have removed `ref.json` from the `getSkippedTests` array and, even with this PR patch applied, the same 8 tests failing today are still failing. One of the failing tests should be passing if the inner refs are resolving correctly:

https://github.com/json-schema/JSON-Schema-Test-Suite/blob/1.1.0/tests/draft4/ref.json#L105-L127

I think you can remove `testInnerDefinitions` in favor of leveraging the official test suite so that we can confirm all expected draft04 functionality and behavior is present. This PR is important (I need functionality to work too) and focusing on that `ref.json` test suite will give us 8x more tests around `$ref` resolution.

araines:
@bighappyface: I'm pretty sure the reason the official test suite is not working at the moment is because the resolver isn't brought in to play at all during that official test run.  Therefore the tests will fail.  I'm going to keep digging and see what I can find out about this, because this is definitely an issue that needs resolving with this package.

bighappyface:
@araines thanks for the feedback and PR. When I tried out the tests locally, I setup the ref resolver as you described although I may not have made it clear. What I was getting at was the lack of comprehensive testing following the established test suites.

I'm checking out your PR now.

bighappyface:
Closing in favor of https://github.com/justinrainbow/json-schema/pull/182

PR #122, opened 2014-12-12
Show values in case of error (for some checks)

doits:
Too see what value is causing the error. Helps in debugging.

I added the values just for some constraints (I needed urgently). If this approach is OK I can add them to all/more constraints. (And add some tests etc.).

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -61,11 +61,12 @@ abstract class Constraint implements ConstraintInterface
     /**
      * {@inheritDoc}
      */
-    public function addError($path, $message)
+    public function addError($path, $message, $value = null)
     {
         $this->errors[] = array(
             'property' => $path,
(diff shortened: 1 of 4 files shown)


src/JsonSchema/Constraints/Constraint.php
```diff
@@ -61,11 +61,12 @@ public function setUriRetriever(UriRetriever $uriRetriever)
...
      * {@inheritDoc}
      */
-    public function addError($path, $message)
+    public function addError($path, $message, $value = null)
     {
         $this->errors[] = array(
             'property' => $path,
-            'message' => $message
+            'message' => $message,
+            'value' => print_r($value, true)
```
bighappyface: @doits Given that the property is provided in an error message, would using a tool like [jq](http://stedolan.github.io/jq/) be better for debugging the JSON file data?

bighappyface:
Closed due to age

PR #125, opened 2015-02-11
Do not throw errors when using unknown/custom format

lwiesel:


bighappyface:
@lwiesel would you please rebase and resolve conflicts? Also, would you mind reviewing #134 to see how this update could factor in?

Maks3w:
For the record I attach the specification fragment about this:

> The value of this keyword is called a format attribute. It MUST be a string. A format attribute can generally only validate a given set of instance types. **If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.**

Maks3w:
@bighappyface The PR is rebased. Travis-CI error status is due failure with composer install

Maks3w:
@bighappyface #134 (now #141) should be rebased again when this is merged or you could resolve the conflict

onlinesid:
I screwed up my PR so I've closed and re-created it as #142 

I've removed unknown format error, updated the tests and squashed my commits into one

lwiesel:
Ok, since github is fine again these days, the Travis-CI build passed after my rebase.

These modifications are indeed included in PR #142, but I'd recommend to merge this PR before and keep #142 focused on optional extra argument to keep atomicity in commits. Anyway it's your call!

bighappyface:
+1

PR #105, opened 2014-06-09
Fix URI generic syntax delimit path components by slash ("`/`")

hakre:
Parts of justinrainbow/json-schema are using the backslash ("`\`") in
error on occasion.

PR #132, opened 2015-03-06
Make the enum element comparison strict

loucho:
The current check is matching numeric strings that are not equal ( like "604.1" == "604.10" because of PHP's == behavior that converts numeric strings to numbers and then performs a numeric comparison), even if type is checked before the comparison, the value should be identic

### DIFF
--- a/src/JsonSchema/Constraints/EnumConstraint.php
+++ b/src/JsonSchema/Constraints/EnumConstraint.php
@@ -23,13 +23,21 @@ class EnumConstraint extends Constraint
     public function check($element, $schema = null, $path = null, $i = null)
     {
         // Only validate enum if the attribute exists
-        if ($element instanceof UndefinedConstraint && (!isset($schema->required) || !$schema->required))  {
+        if ($element instanceof UndefinedConstraint && (!isset($schema->required) || !$schema->required)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


bighappyface:
@loucho have you reviewed the failing unit test in the build for this PR?

loucho:
Oh, hadn't thought of that, i just checked it, seems my change fails when comparing an object (I assume that this is because objects are not === unless they are a reference to the same object in memory or something). 

Maybe add a check for objects before making the comparison? what do you think?

bighappyface:
@loucho sorry for the late response. Would you please rebase for a clean build?

loucho:
Rebased and made the change, waiting on the CI tests

bighappyface:
@loucho solid improvement. I am inclined to suggest wrapping the returns in curlies for consistency sake, and a good rebase would distill this PR down to a single new commit, but it's all good.

+1

I will merge this with another :+1: from the community.

Would you mind helping review some other PRs for their community :+1:? I am sure they will return the favor :smile: 

#105 needs review.

onlinesid:
This is a bug fix so +1 !

bighappyface:
Thanks for the review @onlinesid! Let's call this 1.4.1.

loucho:
Sure, will do

PR #135, opened 2015-03-19
updated vendors

### LINKED ISSUE #124: Consider relaxing phpunit version requirement?
I'm unable to add this project as it requires `phpunit` ~3.7, while other projects I'm using have a 4.* requirement.  Could that be updated to a >=3.7 or is there a reason why the tests are tied to this version?

lsv:
* Updated all vendors
* Updated phpunit to version ~4

Fixes #124

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",
-        "phpunit/phpunit": "~3.7",
+        "phpunit/phpunit": "~4",
         "phpdocumentor/phpdocumentor": "~2"
     },
     "autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -1,31 +1,30 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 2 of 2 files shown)


stof:
@justinrainbow should we remove the lock file from the repo to make sure we can install things on PHP 5.3.3 as well ? the current lock file requires 5.3.23+

bighappyface:
+1, we don't need to store the `composer.lock` file and can instead specify the versions of packages in `composer.json` itself. Pull it out and lets see how the build goes.

keradus:
ref #139

bighappyface:
@lsv please rebase per #139

composer.json
```diff
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",
-        "phpunit/phpunit": "~3.7",
+        "phpunit/phpunit": "~4",
```
GrahamCampbell: Should really be `~4.0`

composer.json
```diff
@@ -40,7 +40,7 @@
...
         "phpdocumentor/phpdocumentor": "~2"
```
GrahamCampbell: ~2.0

bighappyface:

PR #115, opened 2014-10-14
Add stream context to the FileGetContents retriever

willgarcia:
Minor change. Fixes #114

PR #138, opened 2015-03-27
Remove warning if file doesn't exist and send correct exception

alecsammon:


bighappyface:
How would this be affected considering https://github.com/justinrainbow/json-schema/pull/115?

I would like to reconcile these two PRs (#115 and #138) and bump to v1.4.2 along with #105.

alecsammon:
Hi, sorry hadn't really thought this pull request through!

The problem I was trying to solve was the php warning if the file doesn't exist, however if fetching from a url then my solution is always going to fail. Therefor I don't think it's going to play nice with #115.

Although I don't really like it as a solution, the easiest way I can see of getting this to work is to use the @ supressor - see http://stackoverflow.com/a/6463291

bighappyface:
I think the best next step would be to create a test for this retriever so we can try out different URI scenarios :wink:

alecsammon:
Closed this pull request for now. Will reopen with tests later

PR #139, opened 2015-03-27
remove composer.lock

keradus:


bighappyface:
What's up with the failing test?

keradus:
it looks like this project fails with lowest php and deps...

keradus:
Min version of PHP seems to be ok.
Min deps version looks messed up: https://travis-ci.org/justinrainbow/json-schema/builds/56147406

keradus:
Remove testing lowest deps from travis, please check what they should be and add it yourself.

Ready to merge.

bighappyface:
May we add the `self-update` and `--prefer-source` back in? It's noisy without them.

https://travis-ci.org/justinrainbow/json-schema/jobs/56150828#L217-L225

keradus:
machines without sudo have newer composer, but here you need sudo :(

keradus:
selfupdate back in there.
composer cache wont work with prefer-source

bighappyface:
+1

PR #141, opened 2015-03-29
Added optional args to addError

onlinesid:
My pull request #134 was done incorrectly. I've done it on my own separate branch instead of master now.

### DIFF
--- a/src/JsonSchema/Constraints/ConstraintInterface.php
+++ b/src/JsonSchema/Constraints/ConstraintInterface.php
@@ -33,10 +33,12 @@ interface ConstraintInterface
     /**
      * adds an error
      *
-     * @param $path
-     * @param $message
+     * @param string $path
+     * @param string $message
+     * @param string $constraint the constraint/rule that is broken, e.g.: 'minLength'
+     * @param array $more more array elements to add to the error
      */
-    public function addError($path, $message);
+    public function addError($path, $message, $constraint='', $more=null);
 
     /**
      * checks if the validator has not raised errors
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -29,13 +29,13 @@ class FormatConstraint extends Constraint
         switch ($schema->format) {
             case 'date':
                 if (!$date = $this->validateDateTime($element, 'Y-m-d')) {
-                    $this->addError($path, sprintf('Invalid date %s, expected format YYYY-MM-DD', json_encode($element)));
+                    $this->addError($path, sprintf('Invalid date %s, expected format YYYY-MM-DD', json_encode($element)), 'format');
                 }
                 break;
 
             case 'time':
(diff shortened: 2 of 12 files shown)


src/JsonSchema/Constraints/ConstraintInterface.php
```diff
@@ -33,10 +33,12 @@ public function addErrors(array $errors);
...
      *
-     * @param $path
-     * @param $message
+     * @param string $path
+     * @param string $message
+     * @param string $constraint the constraint/rule that is broken, e.g.: 'minLength'
+     * @param array $more more array elements to add to the error
      */
-    public function addError($path, $message);
+    public function addError($path, $message, $constraint='', $more=null);
```
keradus: typehinting

onlinesid: It's in ConstraintInterface.php the phpdoc here is just inheriting the interface class.

onlinesid:
I'm working on a packagist package at the moment: https://packagist.org/packages/onlinesid/json-schema-provider

It's to provide a Service Provider and a validator class for easy use from Silex.

At the moment I have to use an alias (see the part that says dev-sid) because this pull request has been hanging for too long. I might need to start my own or find another fork with more activity.

src/JsonSchema/Constraints/FormatConstraint.php
```diff
@@ -45,74 +45,74 @@ public function check($element, $schema = null, $path = null, $i = null)
...
             case 'hostname':
                 if (!$this->validateHostname($element)) {
-                    $this->addError($path, "Invalid hostname");
+                    $this->addError($path, "Invalid hostname", 'format', array('format' => $schema->format,));
                 }
                 break;
 
             default:
-                $this->addError($path, "Unknown format: " . json_encode($schema->format));
+                $this->addError($path, "Unknown format: " . json_encode($schema->format), 'format', array('format' => $schema->format,));
```
Maks3w: Following #125 discussion the default case should be empty

> The value of this keyword is called a format attribute. It MUST be a string. A format attribute can generally only validate a given set of instance types. **If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.**
http://json-schema.org/latest/json-schema-validation.html#anchor105

onlinesid: Hi @Maks3w , you mean there should not be "$this->addError ..." under "default:" ?

Maks3w: Yes the whole line must be eliminated. default must be empty

Maks3w: Just for clarify. This is not a bug in your PR. The bug already exists in the upstream code, the reason of modify the line is to prevent (supersede) a git conflict with #125

keradus:
SAD :(

onlinesid:
I closed this because I had to add few more commits and it screwed up this PR. I've reopened another PR #142

PR #142, opened 2015-04-05
Optional extra arguments for custom error messages

onlinesid:
And unknown format no longer generate error

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -61,12 +61,20 @@ abstract class Constraint implements ConstraintInterface
     /**
      * {@inheritDoc}
      */
-    public function addError($path, $message)
+    public function addError($path, $message, $constraint='', array $more=null)
     {
-        $this->errors[] = array(
+        $error = array(
             'property' => $path,
-            'message' => $message
+            'message' => $message,
+            'constraint' => $constraint,
         );
+
+        if (is_array($more) && count($more) > 0)
+        {
+            $error += $more;
+        }
+
+        $this->errors[] = $error;
     }
 
     /**
--- a/src/JsonSchema/Constraints/ConstraintInterface.php
+++ b/src/JsonSchema/Constraints/ConstraintInterface.php
@@ -33,10 +33,12 @@ interface ConstraintInterface
     /**
      * adds an error
      *
-     * @param $path
-     * @param $message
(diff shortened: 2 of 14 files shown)


src/JsonSchema/Constraints/Constraint.php
```diff
@@ -61,12 +61,20 @@ public function setUriRetriever(UriRetriever $uriRetriever)
     /**
      * {@inheritDoc}
      */
-    public function addError($path, $message)
+    public function addError($path, $message, $constraint='', $more=null)
     {
-        $this->errors[] = array(
+        $a = array(
```
keradus: please don't use such a meaningless names

src/JsonSchema/Constraints/ConstraintInterface.php
```diff
@@ -33,10 +33,12 @@ public function addErrors(array $errors);
...
      *
-     * @param $path
-     * @param $message
+     * @param string $path
+     * @param string $message
+     * @param string $constraint the constraint/rule that is broken, e.g.: 'minLength'
+     * @param array $more more array elements to add to the error
      */
-    public function addError($path, $message);
+    public function addError($path, $message, $constraint='', $more=null);
```
keradus: if `$more` is an array it should be typehinted as `array`

src/JsonSchema/Constraints/Constraint.php
onlinesid: Hmm, it's better to use meaningless very short names for variables that are being used temporarily and locally within such a small method, but fine I'll change the name.

keradus: you never know if this method will be larger in the future

Maks3w:
Needs rebase

.gitignore
```diff
@@ -7,3 +7,5 @@ coverage
 .settings
 composer.lock
 docs-api
+/.idea
+/composer.phar
```
bighappyface: Please move these to you [global ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

bighappyface:
@onlinesid thanks for the rebase. Would you please squash this down to a single commit?

onlinesid:
I've squashed into a single commit now

bighappyface:
@Maks3w what are your thoughts on the the version number to iterate to when this is merged?

PR #143, opened 2015-04-05
Add constraint factory

Maks3w:
This factory can be injected and allow extend built-in constraint types with custom ones.

This factory allows extend the constraint system with custom ones.

Example:

```php
<?php

class MyFactory extends JsonSchema\Constraints\Factory {
     public function factory($constraintName) {
           if ($constraintName === 'custom_constraint') {
               return new CustomConstraint($this->uriRetriever);
           }

           return super::factory($constraintName);
     }
}

$validator = new Validator(Constraint::CHECK_MODE_NORMAL, null, new MyFactory());
```

This especially usefull when you need extend `FormatConstraint` for to validate custom formats.

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -28,13 +28,20 @@ abstract class Constraint implements ConstraintInterface
     const CHECK_MODE_TYPE_CAST = 2;
 
     /**
+     * @var null|Factory
+     */
+    private $factory;
+
+    /**
      * @param int          $checkMode
      * @param UriRetriever $uriRetriever
+     * @param Factory      $factory
      */
-    public function __construct($checkMode = self::CHECK_MODE_NORMAL, UriRetriever $uriRetriever = null)
+    public function __construct($checkMode = self::CHECK_MODE_NORMAL, UriRetriever $uriRetriever = null, Factory $factory = null)
     {
         $this->checkMode    = $checkMode;
         $this->uriRetriever = $uriRetriever;
+        $this->factory = $factory;
     }
 
     /**
@@ -48,6 +55,18 @@ abstract class Constraint implements ConstraintInterface
         }
 
         return $this->uriRetriever;
+    }
+
+    /**
+     * @return Factory
+     */
+    public function getFactory()
+    {
+        if (!$this->factory) {
+            $this->factory = new Factory($this->getUriRetriever());
+        }
+
+        return $this->factory;
     }
 
     /**
@@ -137,7 +156,7 @@ abstract class Constraint implements ConstraintInterface
      */
     protected function checkArray($value, $schema = null, $path = null, $i = null)
     {
-        $validator = new CollectionConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->createInstanceFor('collection');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
@@ -154,7 +173,7 @@ abstract class Constraint implements ConstraintInterface
      */
     protected function checkObject($value, $schema = null, $path = null, $i = null, $patternProperties = null)
     {
-        $validator = new ObjectConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->createInstanceFor('object');
         $validator->check($value, $schema, $path, $i, $patternProperties);
 
         $this->addErrors($validator->getErrors());
@@ -170,7 +189,7 @@ abstract class Constraint implements ConstraintInterface
      */
     protected function checkType($value, $schema = null, $path = null, $i = null)
     {
-        $validator = new TypeConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->createInstanceFor('type');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
@@ -186,7 +205,7 @@ abstract class Constraint implements ConstraintInterface
      */
     protected function checkUndefined($value, $schema = null, $path = null, $i = null)
     {
-        $validator = new UndefinedConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->createInstanceFor('undefined');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
@@ -202,7 +221,7 @@ abstract class Constraint implements ConstraintInterface
      */
     protected function checkString($value, $schema = null, $path = null, $i = null)
     {
-        $validator = new StringConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->createInstanceFor('string');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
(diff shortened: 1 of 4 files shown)


Maks3w:
@bighappyface ping

bighappyface:
+1

@alecsammon would you mind giving this a review?

src/JsonSchema/Constraints/Factory.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Create a constraint instance for the given constraint name.
+     *
+     * @param string $constraintName
+     * @return ConstraintInterface|ObjectConstraint
+     * @throws InvalidArgumentException if is not possible create the constraint instance.
+     */
+    public function factory($constraintName)
```
alecsammon: I think this might be clearer if called `getConstraintValidator` or something similar. The class is a factory, this is a getter/builder.

src/JsonSchema/Constraints/Factory.php
```diff
@@ -0,0 +1,81 @@
...
+                return new EnumConstraint(Constraint::CHECK_MODE_NORMAL, $this->uriRetriever, $this);
+            case 'format':
+                return new FormatConstraint(Constraint::CHECK_MODE_NORMAL, $this->uriRetriever, $this);
+            case 'schema':
+                return new SchemaConstraint(Constraint::CHECK_MODE_NORMAL, $this->uriRetriever, $this);
+            case 'validator':
+                return new Validator(Constraint::CHECK_MODE_NORMAL, $this->uriRetriever, $this);
+        }
+
+        throw new InvalidArgumentException('Unknown constraint ' . $constraintName);
```
alecsammon: This line appears to be untested - here is a test :)

```php
    /**
     * @expectedException JsonSchema\Exception\InvalidArgumentException
     */
    public function testInvalidConstraint()
    {
        $this->factory->factory('badConstraintName');
    }
```

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -51,6 +58,18 @@ public function getUriRetriever()
     }
 
     /**
+     * @return Factory
+     */
+    public function getFactory()
+    {
+        if (!($this->factory instanceof Factory)) {
```
alecsammon: Does this need to be so strict - surely a ``` if (!$this->factory)``` would suffice?

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -242,7 +261,7 @@ protected function checkEnum($value, $schema = null, $path = null, $i = null)
 
     protected function checkFormat($value, $schema = null, $path = null, $i = null)
     {
-        $validator = new FormatConstraint($this->checkMode, $this->uriRetriever);
+        $validator = $this->getFactory()->factory('format');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
```
alecsammon: As a thought - some of the duplication of this class could now be moved into a private method. ```checkObject``` would need to remain as it is - as it also passes ```$patternProperties```

```php
    protected function checkFormat($value, $schema = null, $path = null, $i = null)
    {
        $this->checkByType('format', $value, $schema, $path, $i);
    }

    /**
     * Create a validator of type $type and validate value against it
     * 
     * @param string $type
     * @param mixed  $value
     * @param mixed  $schema
     * @param mixed  $path
     * @param mixed  $i
     */
    private function checkByType($type, $value, $schema, $path, $i)
    {
        $validator = $this->getFactory()->factory($type);

    }
```

src/JsonSchema/Constraints/Factory.php
```diff
@@ -0,0 +1,81 @@
...
+namespace JsonSchema\Constraints;
+
+use JsonSchema\Exception\InvalidArgumentException;
+use JsonSchema\Uri\UriRetriever;
+use JsonSchema\Validator;
+
+/**
+ * Factory for centralize constraint initialization.
+ */
+class Factory
```
alecsammon: This library appears to duplicate the namespace in the class name. 

For example:
``` JsonSchema\Uri\UriRetriever``` instead of ``` JsonSchema\Uri\Retriever```
```JsonSchema\Constraints\NumberConstraint``` instead of ```JsonSchema\Constraints\Number```

This should maybe be called ```ConstraintFactory```

File: src/JsonSchema/Constraints/Constraint.php (hunk 4 above)
Comment (Maks3w) [inline src/JsonSchema/Constraints/Constraint.php]: This could be something for a different PR

**Comment by alecsammon** [comment]
Hi @Maks3w. Looks like a good PR, is essential that custom validation formats are supported by this library.

I've put quite a few comments on, most of these are to do with my personal style, so happy if you disagree and say no! 

Thanks for you patience with this PR.

File: src/JsonSchema/Constraints/Constraint.php (hunk 4 above)
Comment (alecsammon) [inline src/JsonSchema/Constraints/Constraint.php]: agreed.

File: src/JsonSchema/Constraints/Factory.php (hunk 5 above)
Comment (Maks3w) [inline src/JsonSchema/Constraints/Factory.php]: This seems to be just a cosmetic thing. Prefer to hear other opinions

Comment (alecsammon) [inline src/JsonSchema/Constraints/Factory.php]: @bighappyface - one for you.

**Comment by Maks3w** [comment]
Done

**Comment by alecsammon** [comment]
Hi @Maks3w - great - thank you...

@bighappyface - one question for you - if the ```Factory``` should be renamed to ```ConstraintFactory```

Otherwise I'm happy - once answered then it's ready to merge

+1

**Comment by Maks3w** [comment]
About class name rules I have two:

1. If there are namespaces there is no need for add suffixes or preffixes except the required by PSR-1
2. If Utility classes will share the namespace and may could be name conflicts inside the namespace then non Utitily classes will be suffixed with the role.

**Comment by Maks3w** [comment]
Travis-CI error is not related to this changes.

File: src/JsonSchema/Constraints/Factory.php (hunk 5 above)
Comment (bighappyface) [inline src/JsonSchema/Constraints/Factory.php]: I am okay with the name, but I think we would be best served to stick with the convention established in release 1.4.0: https://github.com/justinrainbow/json-schema/releases/tag/1.4.0 (PR #136)

That being said, I think the suggestion @alecsammon put forth is good guidance.

**Comment by bighappyface** [comment]
@Maks3w would you please rebase this on master for the latest?

**Comment by Maks3w** [comment]
@bighappyface rebased and squashed

File: src/JsonSchema/Constraints/Factory.php (hunk 5 above)
Comment (Maks3w) [inline src/JsonSchema/Constraints/Factory.php]: Well there are lot of opinions and this could be a large discussion.

1. #136 It's a workaround for a technical issue.
2. Recently a PHP 7 RFC has been approved relaxing reserved words based in context.
3. The purpose of namespaces is IMO to have clean class names.

File: src/JsonSchema/Constraints/Constraint.php [hunk 6]
```diff
@@ -28,13 +28,20 @@
...
 
     /**
+     * @var null|Factory
+     */
+    private $factory;
+
+    /**
      * @param int          $checkMode
      * @param UriRetriever $uriRetriever
+     * @param Factory      $factory
```
GrahamCampbell: Missing |null

bighappyface:
@Maks3w sorry for the wait on this. I'll try to get this in with #182 as the next release.

PR #153, opened 2015-05-28
Fix warning on file_get_contents

alecsammon:
Sorry - no test on this. Little bit tricky to write.

If file_get_contents fails then it reports a warning as well as returning false. The code then checks for false and throws an exception. 

This pull request set's a custom error handler (and resets it) that suppresses the warning.

This should be safe to merge with #115.


cozylife:
I'm looking forward to having this request pulled.

ajacquelin:
+1

alecsammon:
Tests added - let me know if you're happy with the format.

alecsammon:
Awesome - thank you

PR #154, opened 2015-06-05
For missing properties, the key for that error in set to ""

samkeen:
Expected bahavior is that the key would be set to the poperty name.

This fix addressed the same issue PR'd by @primitive-type but is
rebased to current master

### DIFF
--- a/tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
+++ b/tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
@@ -9,8 +9,72 @@
 
 namespace JsonSchema\Tests\Constraints;
 
+use JsonSchema\Constraints\UndefinedConstraint;
+
 class RequiredPropertyTest extends BaseTestCase
 {
+
+    public function testErrorPropertyIsPopulatedForRequiredIfMissingInInput()
+    {
+        $validator = new UndefinedConstraint();
+        $document = json_decode(
+            '{
+            "bar": 42
+        }'
+        );
+        $schema = json_decode(
+            '{
+            "type": "object",
+            "properties": {
+                "foo": {"type": "number"},
+                "bar": {"type": "number"}
+            },
+            "required": ["foo"]
+        }'
+        );
+
+        $validator->check($document, $schema);
+        $error = $validator->getErrors();
+        $this->assertErrorHasExpectedPropertyValue($error, "foo");
+    }
+
+    public function testErrorPropertyIsPopulatedForRequiredIfEmptyValueInInput()
+    {
+        $validator = new UndefinedConstraint();
+        $document = json_decode(
+            '{
+            "bar": 42,
+            "foo": null
+        }'
+        );
+        $schema = json_decode(
+            '{
+            "type": "object",
+            "properties": {
+                "foo": {"type": "number"},
+                "bar": {"type": "number"}
+            },
+            "required": ["foo"]
+        }'
+        );
+
+        $validator->check($document, $schema);
+        $error = $validator->getErrors();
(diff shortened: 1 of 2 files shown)


bighappyface:
@samkeen good fix. Would you mind including a test for this?

samkeen:
added tests.  Let me know if they are adequate.   I can also rebase to a single commit if you prefer.

bighappyface:
@samkeen thank you for the update! I think a squash would be great.

I have restarted the Travis runs to see if the failures clear up.

samkeen:
rebased to current upstream/master, + 2 commits squashed to 1

bighappyface:
+1 looks good. @alecsammon @Maks3w would you mind to help review?

tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
```diff
@@ -9,8 +9,52 @@
...
+            "properties": {
+                "foo": {"type": "number"},
+                "bar": {"type": "number"}
+            },
+            "required": ["foo"]
+        }');
+
+        $validator->check($document, $schema);
+        $error = $validator->getErrors();
+        $this->assertEquals('foo', $error[0]['property']);
```
Maks3w: If `$error` is not well formatted, i.e. there is no index 0 or 'property` key does not exists test run will stop without an useful message.

Could you try to compare the whole `$error` variable or assert 0 and property exists?

tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
```diff
@@ -9,8 +9,52 @@
...
```
Maks3w: Same here

Maks3w:
@bighappyface, @samkeen has reveal invalid data provider is not enough without to check invalid data fails with the expected message.

i.e.  invalid data providers should be refactored for to check the error message too.

tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
samkeen: added check for unset `$error[0]['property']` in both tests

tests/JsonSchema/Tests/Constraints/RequiredPropertyTest.php
```diff
@@ -9,8 +9,60 @@
...
+            "properties": {
+                "foo": {"type": "number"},
+                "bar": {"type": "number"}
+            },
+            "required": ["foo"]
+        }');
+
+        $validator->check($document, $schema);
+        $error = $validator->getErrors();
+        if(!isset($error[0]['property'])) {
```
Maks3w: PHPUnit provides `assertArrayHasKey` and `assertArraySubset` try to use one of them.

Also this does not check if index `0` is set.

Maks3w: Sorry `assertArraySubset` requires ^4.4.0 /cc @bighappyface

samkeen: ah, sorry been away from PHP for a while.  Let me know if latest push works, not the most elegant code but I think it covers any malformed input with a helpful fail message.

samkeen:
sorry @Maks3w , i realized i didn't use `assertArrayHasKey` as you suggested.  I guess I thought it was better to have a single PHPUnit Assertion per test.  If you like though, I can change it.

bighappyface:
@Maks3w any last thoughts on this PR?

nsitbon:
Can you merge guys we are actively waiting for this PR making our tests failed.
Thx.

PR #155, opened 2015-06-13
chore(travis): remove phpdoc ext and ignore output

bighappyface:
This PR fixes a warning when running `phpdoc` on Travis, and ignores the `phpdoc` output.


Maks3w:
:+1:

PR #157, opened 2015-06-13
Scrutinizer-CI is not used

Maks3w:
Initial experiment is 2 years old

https://scrutinizer-ci.com/g/justinrainbow/json-schema/inspections

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,5 +1,0 @@
-before_commands:
-  - composer install --dev
-
-tools:
-  php_analyzer: true


.scrutinizer.yml
```diff
@@ -1,5 +0,0 @@
-before_commands:
```
GrahamCampbell: Scrutiniser doesn't allow you to run this command anyway anymore.

bighappyface: @GrahamCampbell this PR is to remove the Scrutinizer file from the repo.

bighappyface:
+1

.scrutinizer.yml
GrahamCampbell: I know. That's why I said "anyway anymore.".

GrahamCampbell: I was pointing out that this config is wrong anyway.

PR #159, opened 2015-06-14
Promote HHVM.

Maks3w:
This repo has a coverage bigger than 85% and tests works with HHVM. This means HHVM support is enough stable for to make it officially supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ php:
 matrix:
     allow_failures:
         - php: 7
-        - php: hhvm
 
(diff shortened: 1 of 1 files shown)


bighappyface:
+1 based on your reasoning. I am not an HHVM user so it makes no difference for me, but I do as that an HHVM user/supporter approve to have something more... official. Cool?

bighappyface:
@Maks3w I am going to move forward with merging this PR. Thanks for the help :smile_cat:

PR #164, opened 2015-07-13
Fix: Autoload test resources with Composer

localheinz:
This PR

* [x] leverages Composer's autoloading for dev resources and removes autoloading of tests from `tests/bootstrap.php`
* [x] uses `vendor/autoload.php` for bootstrapping tests, it suffices


localheinz:
@bighappyface 

I'd follow up on this one with switching from PSR-0 to PSR-4, what do you think?

bighappyface:
+1

@localheinz PSR-4 would be a good follow up.

PR #166, opened 2015-07-13
Fix: Encourage to use phpunit as installed with Composer

localheinz:
This PR

* [x] updates `README.md` and encourages to use `phpunit` as installed with Composer


bighappyface:
+1

PR #167, opened 2015-07-13
Fix: Constraint::addError() does not return anything

localheinz:
This PR

* [x] calls a method, then returns, rather than attempting to return the result of a method which does not return anything


bighappyface:
+1 makes sense.

PR #168, opened 2015-07-13
Enhancement: Speed up builds as much as possible

localheinz:
This PR

* [x] indents `.travis.yml` consistently (unrelated)
* [x] provides feedback quickly if it has failed, using the `fast_finish` option
* [x] moves update of Composer itself to `before_install` section (unrelated)
* [x] runs `composer install` instead of `composer update` (unrelated)
* [x] checks links on PHP 5.6 only 
* [x] builds docs on PHP 5.6 only

### Before

screen shot 2015-07-12 at 23 25 43

See https://travis-ci.org/justinrainbow/json-schema/builds/70676013.

### After

screen shot 2015-07-13 at 00 09 41

https://travis-ci.org/justinrainbow/json-schema/builds/70680917.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
 
 install:
   - sudo apt-get -y install pypy python-sphinx graphviz
-  - travis_retry composer update --no-interaction --prefer-source
+  - travis_retry composer install --no-interaction --prefer-source
 
 script:
   - cd docs && make linkcheck && cd ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ php:
   - hhvm
 
 matrix:
+  fast_finish: true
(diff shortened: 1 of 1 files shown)


localheinz:
Seems like most of the time is still spent on installing dependencies with Composer. If links wouldn't needed to be checked on Travis, container-based infrastructure could be enabled, cutting the whole build down to probably around 2 minutes.

bighappyface:
@localheinz would you push the link check and docs build below `phpunit` for the fastest fail on PHP 5.6? We probably should have done that a long time ago considering it's pointless to build the docs for broken code.

Overall I think this is a great improvement.

localheinz:
@bighappyface 

:+1:

bighappyface:
+1

PR #169, opened 2015-07-13
Enhancement: Switch to PSR-4 autoloading

localheinz:
This PR

* [x] switches to PSR-4 autoloading

Follows #164.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,10 +44,10 @@
         "phpdocumentor/phpdocumentor": "~2"
     },
     "autoload": {
-        "psr-0": { "JsonSchema": "src/" }
+        "psr-4": { "JsonSchema\\": "src/JsonSchema/" }
     },
     "autoload-dev": {
-        "psr-0": { "JsonSchema": "tests/" }
+        "psr-4": { "JsonSchema\\Tests\\": "tests/JsonSchema/Tests/" }
     },
     "bin": ["bin/validate-json"],
     "extra": {


GrahamCampbell:
> flattens the directory structure

:-1:

bighappyface:
@localheinz @GrahamCampbell I am concerned about this PR as well. While I am all for PSR-4, flattening the directories could be really problematic for forked variations with patches applied in lieu of the historically slow turnaround for PR merges on this repo. IMO, PSR-4 is not out, but we will need to spend more time on the structural impacts.

localheinz:
@bighappyface @GrahamCampbell 

Rebased and push-forced, so switching to PSR-4 now, but not flattening the directory structure!

composer.json
```diff
@@ -44,10 +44,10 @@
         "phpdocumentor/phpdocumentor": "~2"
     },
     "autoload": {
-        "psr-0": { "JsonSchema": "src/" }
+        "psr-4": { "JsonSchema\\": "src/JsonSchema" }
```
GrahamCampbell: Should end in a `/` for the dirnames in both cases.

bighappyface:
+1

@GrahamCampbell you good with this?

GrahamCampbell:
:+1:

bighappyface:
Derp! Just saw @GrahamCampbell responded on this and rolled v1.4.4 :disappointed: 

Will merge and account for this in 1.4.5 once another source code-related change is merged.

localheinz:
@bighappyface 

:+1:

PR #170, opened 2015-07-13
Fix: Start all messages with a capital letter

localheinz:
This PR

* [x] starts all messages with a capital letter

:information_desk_person: Some messages already do, others don't - it's just an opinion, but I feel it should be consistent. What do you think?

### DIFF
--- a/src/JsonSchema/Constraints/EnumConstraint.php
+++ b/src/JsonSchema/Constraints/EnumConstraint.php
@@ -41,6 +41,6 @@ class EnumConstraint extends Constraint
             }
         }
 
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
:+1:

bighappyface:
+1

My 2 cents: a good fast-follow would be to convert the string literals to single-quotes or add the various vars into the double-quoted string to take better advantage of variable interpolation, instead of concatenation operations. Probably micro-optimization when we have more important concerns, but hey...

localheinz:
@bighappyface 

:+1:

Actually, I was also thinking about consistently using `sprintf`, what do you think?

PR #171, opened 2015-07-13
Fix: Be more specific with error message

localheinz:
This PR

* [x] attempts to be more specific with error messages when requiring a minimum / maximum number of items in an array


bighappyface:
+1

PR #173, opened 2015-07-16
Fix path output for required properties

JustTSK:
Example:

Before:
   `day: "The property day is required"`
After
   `availability[0].hours[0].day: "The property day is required"`

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -121,11 +121,7 @@ class UndefinedConstraint extends Constraint
                 // Draft 4 - Required is an array of strings - e.g. "required": ["foo", ...]
                 foreach ($schema->required as $required) {
                     if (!property_exists($value, $required)) {
-<<<<<<< HEAD
-                        $this->addError($required, "The property " . $required . " is required", 'required');
-=======
                         $this->addError((!$path) ? $required : "$path.$required", "The property " . $required . " is required", 'required');
->>>>>>> 9bda789... add variable path
                     }
                 }
             } else if (isset($schema->required) && !is_array($schema->required)) {


bighappyface:
@JustTSK thanks for the PR.

Please remove the `.gitignore` addition in favor of your [global git ignore](http://stackoverflow.com/questions/7335420/global-git-ignore) for `.DS_Store`.

I think the outcome is good, but the implementation could be cleaner.

Perhaps something like:

```php
$required = (!$path) ?: "$path.$required";
```

This would allow you to save a variable, a full condition block, and use the more efficient string interpolation over concatenation.

@localheinz @samkeen thoughts?

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -122,7 +122,12 @@ protected function validateCommonProperties($value, $schema = null, $path = null
                 // Draft 4 - Required is an array of strings - e.g. "required": ["foo", ...]
                 foreach ($schema->required as $required) {
                     if (!property_exists($value, $required)) {
-                        $this->addError($required, "The property " . $required . " is required");
+                        if($path) {
```
GrahamCampbell: cs

JustTSK:
@bighappyface we cant override `$required` variable because its used in error description: `"The property " . $required . " is required"`

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -122,7 +122,12 @@ protected function validateCommonProperties($value, $schema = null, $path = null
...
+                            $property_path = $path .".". $required;
+                        } else {
+                            $property_path = $required;
+                        }
+                        $this->addError($property_path, "The property " . $required . " is required");
```
JustTSK: can be
```php
$this->addError( (!$path) ? "$required" : "$path.$required" , "The property " . $required . " is required");
```

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -122,7 +122,7 @@ protected function validateCommonProperties($value, $schema = null, $path = null
                 // Draft 4 - Required is an array of strings - e.g. "required": ["foo", ...]
                 foreach ($schema->required as $required) {
                     if (!property_exists($value, $required)) {
-                        $this->addError($required, "The property " . $required . " is required");
+                        $this->addError( (!$path) ? $required : "$path.$required" , "The property " . $required . " is required");
```
GrahamCampbell: cs

JustTSK: ?

bighappyface: Code style

localheinz:
@JustTSK 

Build is currently failing because of some weird error:

> [Composer\Repository\RepositorySecurityException]                            
The contents of http://packagist.org/p/provider-2015-04$09177ee9742784c520bc46745e536dd726464fd578db4be65f2924bf8c6f1c7c.json do not match its signature. This should indicate a man-in-the-middle attack. Try running composer again and report this if you think it is a mistake.

src/JsonSchema/Constraints/UndefinedConstraint.php
JustTSK: Ok, can i have more info what code style are you using? I cant find it anywhere in the docs

JustTSK: @GrahamCampbell ?

mwaterboehr:
I have not noticed this pull request before. So I just created another pull request (PR #190) that fixes this issue. What do you think about my solution there?

JustTSK:
Updated pull request

PR #174, opened 2015-07-17
Enhancement: Require more recent phpunit/phpunit

localheinz:
This PR

* [x] updates the `phpunit/phpunit` requirement to have a more recent version of it installed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",
-        "phpunit/phpunit": "~3.7",
+        "phpunit/phpunit": "^4.8.22",
         "phpdocumentor/phpdocumentor": "~2"
     },
     "autoload": {


bighappyface:
@localheinz thoughts on the TravisCI failures? Perhaps the legacy hardware is part of the issue?

composer.json
```diff
@@ -40,8 +40,8 @@
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",
-        "phpunit/phpunit": "~3.7",
-        "phpdocumentor/phpdocumentor": "~2"
+        "phpdocumentor/phpdocumentor": "~2",
```
GrahamCampbell: `^2.0` or something would be better here.

composer.json
```diff
@@ -40,8 +40,8 @@
...
+        "phpunit/phpunit": "^4.7.7"
```
GrahamCampbell: Why do we need that version specifically? Would `^4.0` not be sufficient?

composer.json
localheinz: @GrahamCampbell 

The change here is only related to requiring `phpunit/phpunit` with

```
$ composer require --dev --sort-packages phpunit/phpunut:^4.7.7
```

composer.json
localheinz: @GrahamCampbell 

It probably is - however, I've made it a habit when updating packages to use the caret operator, and furthermore, specify the most recent version. It doesn't hurt, and the eventual package installed will be the same.

composer.json
```diff
@@ -40,8 +40,8 @@
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",
-        "phpunit/phpunit": "~3.7",
-        "phpdocumentor/phpdocumentor": "~2"
+        "phpdocumentor/phpdocumentor": "~2",
+        "phpunit/phpunit": "^4.7"
```
localheinz: @GrahamCampbell 

Requiring `phpunit/phpunit` without any version constraint yields to what you can see here now!

localheinz:
@bighappyface 

Let's see what https://travis-ci.org/justinrainbow/json-schema/builds/71827504 says!

GrahamCampbell:
> Requiring phpunit/phpunit without any version constraint yields to what you can see here now!

I never suggested that...

localheinz:
@bighappyface 

Builds still fail trying to update composer. Workaround could be to check in `composer.phar` and use it for installing dependencies on Travis (given that still works).

bighappyface:
+1

It looks like the PHP 5.3.x Travis builds are the only ones failing now, and they are still exhibiting the man in the middle issue. I am good with the builds as is, and I am good with the upgrade for PHPUnit.

@GrahamCampbell are you okay with the latest?

.travis.yml
```diff
@@ -18,7 +18,7 @@ before_install:
   - composer selfupdate
```
GrahamCampbell: This command should be removed.

.travis.yml
```diff
@@ -18,7 +18,7 @@ before_install:
...
 
 install:
-  - travis_retry composer install --no-interaction --prefer-source
+  - composer install --no-interaction --prefer-dist
```
GrahamCampbell: prefer-dist is wrong

GrahamCampbell: You definely don't want that. Using that is just selfishly using api calls for that mashine.

composer.json
GrahamCampbell: > Requiring phpunit/phpunit without any version constraint yields to what you can see here now!

`^4.0` would be fine?

localheinz:
@bighappyface 

Do you want to take a look again?

bighappyface:
@localheinz thanks for the reminder. While I am all about the update I recently encountered a some issues with changing versions of dependencies, even dev dependencies. Are there major updates in PHPunit 4.x dependencies that we may need to be concerned with? I would hate for this update to conflict with user installs and we would probably need to consider a minor bump.

bighappyface:
@localheinz 

I have compared the `phpunit` dependencies between versions and I think we are probably safe:

https://packagist.org/packages/phpunit/phpunit#3.7.38
https://packagist.org/packages/phpunit/phpunit#4.8.23

I'm going to merge and see what people using the `dev` latest encounter.

davidwindell:
This seems to be causing us some problems with composer;

```
Fatal error: Uncaught exception 'UnexpectedValueException' with message 'Could not parse version constraint ^4.8.22: Invalid version string "^4.8.22"' in /var/www/vendor/composer/composer/src/Composer/Package/Version/VersionParser.php:378
Stack trace:
#0 /var/www/vendor/composer/composer/src/Composer/Package/Version/VersionParser.php(244): Composer\Package\Version\VersionParser->parseConstraint('^4.8.22')
#1 /var/www/vendor/composer/composer/src/Composer/Package/Version/VersionParser.php(207): Composer\Package\Version\VersionParser->parseConstraints('^4.8.22')
#2 /var/www/vendor/composer/composer/src/Composer/Package/Loader/ArrayLoader.php(123): Composer\Package\Version\VersionParser->parseLinks('justinrainbow/j...', 'dev-master', 'requires (for d...', Array)
#3 /var/www/vendor/composer/composer/src/Composer/Repository/FilesystemRepository.php(62): Composer\Package\Loader\ArrayLoader->load(Array)
#4 /var/www/vendor/composer/composer/src/Composer/Repository/ArrayRepository.php(170): Composer\Repository\FilesystemRepository in /var/www/vendor/composer/composer/src/Composer/Package/Version/VersionParser.php on line 378
```

localheinz:
@davidwindell 

Time to update composer itself, the `^` operator has been supported for a long, long time!

```
$ composer self-update
```

davidwindell:
Yep, already running latest release, still get the error.

PR #179, opened 2015-07-27
Change to allow MaxDepth to be set on construct of the RefResolver

s3than:
If required for complex schemas the ability to set the max depth on construct of the RefResolver

No B/C break.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ $schema = $retriever->retrieve('file://' . realpath('schema.json'));
 $data = json_decode(file_get_contents('data.json'));
 
 // If you use $ref or if you are unsure, resolve those references here
-// This modifies the $schema object
+// This modifies the $schema object, if the schema has a max depth greater than 7 
+// you can pass as an additional value in the RefResolver
 $refResolver = new JsonSchema\RefResolver($retriever);
 $refResolver->resolve($schema, 'file://' . __DIR__);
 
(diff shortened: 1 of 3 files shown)


s3than:
It looks like Travis timed out the composer update. Can you confirm or re-run. Thanks

bighappyface:
@s3than thanks for the help.

Would you please expand on the current test for this, or create another for testing the maximum depth?

https://github.com/justinrainbow/json-schema/blob/34ca930d9fbfb439ae8d3b87ca046b914e190295/tests/JsonSchema/Tests/RefResolverTest.php

GavinCS:
this fix doesn't fix the underlying problem faced with recursive referencing as in issue #180.
I ended up with the following fatal exception:
`Fatal error: Maximum function nesting level of '250' reached, aborting! in .../vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php on line 160`

GrahamCampbell:
Your nesting limit is probably still too low then.

GavinCS:
Thanks. After further investigation I found that this exception is being thrown by the nesting level set in xdebug. 

However no matter how high I set the maximum depth, recursion seems to be causing an infinite loop.
This is not the case when reverting to 1.4.3 pre maximum depth

bighappyface:
@GavinCS would you add a test or post a JSON document showcasing the issues you are encountering?

Regarding 1.4.3, the exception was added in 1.4.4 in place of a hard return. In 1.4.3, the validator isn't considering the nested schemas at all past the maximum depth, so I am not sure that 1.4.3 was actually working as much as just not throwing the exception.

bighappyface:

PR #182, opened 2015-08-15
Fix #93 ($ref to local definition not working)

### LINKED ISSUE #93: $ref to local definition not working

araines:
Another attempt at getting issue #93 sorted - thanks to @co3k and @stoiev for putting in the initial leg work.  Based my initial work on #119, but actually got the tests running properly with the resolver too and the official test pack.  This reduces the 8 failures mentioned by @bighappyface in #119 down to 3.

I'm not sure this is quite the right way to fix it from an architectural point of view, but it does improve the capability of this library, at least.

In order to get the remaining 3 failing tests to pass, I suspect there will need to be a fairly big rewrite.  The main problem is around that of recursion - self-referencing documents will simply cause the current design to fail.  But that should be the subject of another issue.

### DIFF
--- a/tests/JsonSchema/Tests/Constraints/BaseTestCase.php
+++ b/tests/JsonSchema/Tests/Constraints/BaseTestCase.php
@@ -9,6 +9,8 @@
 
 namespace JsonSchema\Tests\Constraints;
 
+use JsonSchema\RefResolver;
+use JsonSchema\Uri\UriRetriever;
 use JsonSchema\Validator;
 
 abstract class BaseTestCase extends \PHPUnit_Framework_TestCase
@@ -18,9 +20,14 @@ abstract class BaseTestCase extends \PHPUnit_Framework_TestCase
      */
     public function testInvalidCases($input, $schema, $checkMode = Validator::CHECK_MODE_NORMAL, $errors = array())
     {
+        $schema = json_decode($schema);
+
+        $refResolver = new RefResolver(new UriRetriever);
+        $refResolver->resolve($schema);
+
         $validator = new Validator($checkMode);
 
-        $validator->check(json_decode($input), json_decode($schema));
+        $validator->check(json_decode($input), $schema);
 
         if (array() !== $errors) {
             $this->assertEquals($errors, $validator->getErrors(), print_r($validator->getErrors(),true));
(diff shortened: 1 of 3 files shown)


bighappyface:
+1

This is the best start to resolving this (pun intended).

@araines thanks a ton for using the standard test suite and getting this most of the way there. With the high demand for this support, I think it is good to move forward.

@alecsammon @localheinz @GrahamCampbell would you please review?

graste:
It would be really nice to have this working.

bighappyface:
@graste may I have your help with code review? I prefer two peer reviews prior to merging a PR.

graste:
Okay, I checked out master and applied this PR. It brings down the draft4 ref test failures from 8 to 3 as advertised. What I noticed is, that issue #93 is working fine even without this PR – see [this gist](https://gist.github.com/graste/2fccfa710a375f93105d). After having a look at the code I must say, that I've got the same feeling as @araines has with his PR: The solution might not be ideal as it feels a bit like not reusing stuff like [`UriRetriever::resolvePointer`](https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Uri/UriRetriever.php#L95) – otoh it brings down the failing tests. My suggestion to you would be to decide whether it's worth the improvements and temporarily include it or take the time to refactor the whole `definitions` and `$ref` related stuff. Short term including the PR would probably be okay, long term you need a bit of thinking/refactoring to handle the skipped tests from the json schema suite anyways.

What I noticed though when testing a `$ref` use case is, that the `oneOf` doesn't handle objects with different content correctly. I'll create a short PR with succeeding and failing test to show that.

bighappyface:
@graste thanks for the review. I am merging and rolling v1.4.5

tests/JsonSchema/Tests/Constraints/BaseTestCase.php
```diff
@@ -33,13 +40,18 @@ public function testInvalidCases($input, $schema, $checkMode = Validator::CHECK_
...
     public function testValidCases($input, $schema, $checkMode = Validator::CHECK_MODE_NORMAL)
     {
+        $schema = json_decode($schema);
+
+        $refResolver = new RefResolver(new UriRetriever);
+        $refResolver->resolve($schema);
+
         $validator = new Validator($checkMode);
 
-        $validator->check(json_decode($input), json_decode($schema));
```
Maks3w: This is a clear sympton of a BC Break

bighappyface: I don't follow. Are you referring to the addition of `RefResolver` in the lines above or this specific line?

Issue #181, opened 2015-07-27
Configure RefResolver:$maxDepth?


eschwartz:
I kept getting `The maximum stack depth has been exceeded` errors, and I couldn't figure out what was going on. Finally, I found a public static `RefResolver$maxDepth` resolver property. I tried setting it to 20 (default was 7), and the schema validated fine.

I feel a little funny overriding a public static property on a test utility. What would you think about making this a configurable property on the `RefResolver` instance?

vdespa:
Same issue here.

micahgodbolt:
This is also required for our project. We either have to manually change the file after installation, or we'll need to fork the project. Would much rather just have a param we could set.

bighappyface:
@micahgodbolt why not submit a PR instead of working from a fork? If you are going to make a change that will help others, you should consider giving it back.

micahgodbolt:
@bighappyface oh I certainly would! I totally agree with contributing > complaining :) I was going to say that #184 already submitted one, but it seems you already closed that.

bighappyface:
I'll help with any PR that has support from the community. I don't think #184 accomplishes anything and closed it predicting that it would not pass community scrutiny.

PRs are very welcome and the exclusive source of maintenance on this project. My only ask is that you submit changes with suitable tests and consider helping review other PRs in need of a review.

mathroc:
`RefResolver:$maxDepth` is a public (static) property  so it's already configurable:

```php
RefResolver:$maxDepth = 42;
```

eschwartz:
Yes, you can do that. I guess my point was that I felt kind of dirty overriding a global static property, that is used by all instances of RefResolver.

bighappyface:
Closed with https://github.com/justinrainbow/json-schema/releases/tag/2.0.0

PR #184, opened 2015-08-20
#181 make RefSolver::$maxDepth able to configure

edisonthk:
I am having this #184 issue also. Due to low maxDepth, `JsonSchema\Exception\JsonDecodingException: The maximum stack depth has been exceeded` message is keep throwing.

Make it changeable so that it can be changed based on developer needs.

### DIFF
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -226,4 +226,9 @@ class RefResolver
 
         return $this;
     }
+
+    public function setMaxDepth($maxDepth)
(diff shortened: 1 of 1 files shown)


bighappyface:
@edisonthk you don't need a setter method to set the value of a public, static variable like `JsonSchema/RefResolver::$maxDepth`

```php
// e.g.
JsonSchema/RefResolver::$maxDepth = 20;
```

The setter method would only work using a concrete instance, and the static variable value is used across all instances.

PR #186, opened 2015-08-31
Fix additional properties validation

mrix:
This fixes error when `additionalProperties` is ignored if `properties` or `patternProperties` is not set.
Such definitions are used in http://json-schema.org/draft-04/schema


bighappyface:
+1

stefk:
+1. This is the correct behaviour. The added tests clearly match the spec, and they're all red without this fix.

bighappyface:
@stefk would this conflict with the work you did in #191?

stefk:
@bighappyface, yes, but I'll update that PR if needed.

PR #187, opened 2015-09-03
add oneOf object type tests – one of them fails

graste:
The resolution of the definitions of the objects for the `oneOf` constraint with `$ref` referencing something works, but the actual validation of the objects then fails for the second added test case with ```prop2 – failed to match exactly one schema```.

See https://gist.github.com/graste/968573d9c29280b66a7c as well for a small usecase/example.

DON'T JUST MERGE this PR as it contains a failing test case deliberately.


bighappyface:
Closed with https://github.com/justinrainbow/json-schema/pull/182

Please re-open if necessary.

graste:
Unfortunately testing will have to wait a month or two. Thanks for the work though! :+1:

Issue #144, opened 2015-04-06
uri format should allow protocol-relative urls


legoktm:
Per <https://tools.ietf.org/html/rfc3986#section-4.2> protocol-relative urls are valid (it claims they're "rarely used" but that's not true anymore).

Downstream MediaWiki bug is https://phabricator.wikimedia.org/T93010

bighappyface:
Closed with https://github.com/justinrainbow/json-schema/pull/174

legoktm:
Huh? What does #174 and phpunit have to do with this bug?

bighappyface:
@legoktm good catch! I meant https://github.com/justinrainbow/json-schema/pull/245

PR #189, opened 2015-09-27
Related to #144

siad007:
Would fix various uri behavior according to rfc 3986 instead of only support rfc 2396.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,10 +44,10 @@
         "phpdocumentor/phpdocumentor": "~2"
     },
     "autoload": {
-        "psr-0": { "JsonSchema": "src/" }
+        "psr-4": { "JsonSchema\\": "src/JsonSchema/" }
     },
     "autoload-dev": {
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -5,5 +5,6 @@ coverage
 .buildpath
 .project
 .settings
+.idea
```
bighappyface: This should be added to a [global ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

bighappyface:
I prefer a more descriptive commit message

.gitignore
siad007: @bighappyface I would suggest to remove
```
.buildpath
.project
.settings
```
as well.

bighappyface:
@siad007 thanks for updates. Next steps are to rebase and squash this down to a single commit.

bighappyface:
Closed due to age

PR #191, opened 2015-10-11
Unskip official test + various test suite enhancements

stefk:
Hi,

The main goal of this PR is to unskip the [zero terminated floats](https://github.com/json-schema/JSON-Schema-Test-Suite/blob/develop/tests/draft4/optional/zeroTerminatedFloats.json) test from the official suite. 

It turns out that the issue doesn't come from the validator itself, but from the way test cases are loaded, i.e. using a `json_encode` extra step ([here](https://github.com/justinrainbow/json-schema/blob/master/tests/JsonSchema/Tests/Drafts/BaseDraftTestCase.php#L24)) which casts zero terminated floats to integers. The encoding option `JSON_PRESERVE_ZERO_FRACTION` could help, but it's only available as of PHP 5.6.6.

So... long story short, I took the liberty to refactor the test suite in order to address this issue, and while I was at it, to make improvements where I could in the testing setup. Here are the results:

- All the tests in `Tests\Constraints` which previously used PHP strings to store JSON data are now based entirely on independent JSON files, containing the data needed to execute them, just like in the official JSON Schema suite.
- A new test case class, `ValidatorTestCase`, provides the main test method as well as helpers for all the tests based on JSON files (i.e. both constraint and draft tests).
- The `getSkippedTests` method is replaced by a *whitelist*/*blacklist* functionality, allowing to quickly include/exclude tests during development.
- Default failure messages and diffs from PHPUnit are minimized and errors now appear in a clean and readable format. Example:
```
**********************************************************************

File: arrays.json

Test: valid test #2 (check mode: 1)

Schema: {
    "type": "object",
    ...
}

Instance: {
    "data": [
        1,
        2,
       ...
    ]
}

Expected: no error

Actual: [
    {
        "property": "data[3]",
        "message": "string value found, but an integer is required"
    }
]

**********************************************************************
```

 -  There're now 4 more tests in the suite (zero terminated floats for draft3 and 4).Obviously, all the previous tests are still working.


.gitignore
```diff
@@ -3,6 +3,7 @@
 !/bin/validate-json
 coverage
 .buildpath
+.idea
```
bighappyface: This should be in the [global ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

bighappyface:
@stefk thanks a ton for this work. The improved consistency with the full JSON Schema test suite is very welcome.

I must admit that it is a large PR and difficult to review considering the amount of PHP string to JSON conversion, but it looks like we are good to go.

All that I ask now is for another member in the community to approve and I will move forward with a merge and rolling another version.

+1

bighappyface:
@stefk I have merged #186 and you can rework this PR upon it. I'll merge this in when ready.

stefk:
@bighappyface, done and rebased.

bighappyface:
@stefk oops, this has gone into conflict again. Please rebase once more and I'll get this through.

bighappyface:
Closed due to age

PR #192, opened 2015-10-15
Create .gitattributes

gsouf:
git attributes allows to reduce size of dependency, very usefull when you create a phar! More explanations: https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,5 @@
-/doc                export-ignore
+/docs               export-ignore
 /tests              export-ignore
 .gitignore          export-ignore
 .travis.yml         export-ignore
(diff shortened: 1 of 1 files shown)


ktomk:
Why exclude the `README.md ? It wouldn't contain much info for the zip package *but* not having one seem unfriendly to me. Maybe better update it?

bighappyface:
@gsouf I agree with @ktomk and the first line should be `docs`

I'll merge with those updates. Good contribution.

gsouf:
I did put README.md because he only files you need in production are the source files

bighappyface:
+1

PR #194, opened 2015-10-22
bugfix: patternProperties raised errors when the pattern has slashes

ribeiropaulor:
Now the delimiter to evaluate the regex is chosen dynamically.

### DIFF
--- a/src/JsonSchema/Constraints/ObjectConstraint.php
+++ b/src/JsonSchema/Constraints/ObjectConstraint.php
@@ -42,15 +42,23 @@ class ObjectConstraint extends Constraint
 
     public function validatePatternProperties($element, $path, $patternProperties)
     {
+        $try = array('/','#','+','~','%');
         $matches = array();
         foreach ($patternProperties as $pregex => $schema) {
+            // Choose delimiter. Necessary for patterns like ^/ , otherwise you get error
+            foreach ($try as $delimiter) {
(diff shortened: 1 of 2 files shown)


bighappyface:
@ribeiropaulor please rebase on master. Thanks for adding the tests.

ribeiropaulor:
rebased!

src/JsonSchema/Constraints/ObjectConstraint.php
```diff
@@ -42,15 +42,23 @@ function check($element, $definition = null, $path = null, $additionalProp = nul
 
     public function validatePatternProperties($element, $path, $patternProperties)
     {
+        $try = array('/','#','+','~','%');
         $matches = array();
         foreach ($patternProperties as $pregex => $schema) {
+            // Choose delimiter. Necessary for patterns like ^/ , otherwise you get error
+            foreach ($try as $delimiter) {
```
bighappyface: `$delimiter` should be defined and defaulted to a sensible value before this loop. Otherwise, `$delimiter` is only defined per the loop and will be set the last value if nothing breaks the loop.

ribeiropaulor: done!

ribeiropaulor:
should be good to merge

bighappyface:
+1

PR #196, opened 2015-11-05
Better EnumConstraint error message

wpillar:
Currently `EnumConstraint` uses `print_r` which produces messages like:

```
Does not have a value in the enumeration Array\n(\n    [0] => GBP\n    [1] => USD\n)
```

Which is not the most friendly messaging in the world. I'm changed this to use `implode` to make it more readable:

```
value is not in enumeration: [GBP, USD]
```

(There's probably broken tests as a result of this, I'll fix these if we're happy to accept the change)

Thoughts?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,5 @@
-/doc                export-ignore
+/docs               export-ignore
 /tests              export-ignore
 .gitignore          export-ignore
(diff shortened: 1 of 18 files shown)


bighappyface:
@wpillar please see the TravisCI console for failures.

PR #197, opened 2015-11-09
Better enum constraint error message

wpillar:
Currently `EnumConstraint` uses `print_r` which produces messages like:

```
Does not have a value in the enumeration Array\n(\n    [0] => GBP\n    [1] => USD\n)
```

Which is not the most friendly messaging in the world. I've changed this to handle more like this:

```
value is not in enumeration: [GBP, USD]
```

It makes it much easier to read for simple enums, for complex enums involving objects they are stilled handled with `print_r`.

### DIFF
--- a/src/JsonSchema/Constraints/EnumConstraint.php
+++ b/src/JsonSchema/Constraints/EnumConstraint.php
@@ -41,6 +41,31 @@ class EnumConstraint extends Constraint
             }
         }
 
-        $this->addError($path, "Does not have a value in the enumeration " . print_r($schema->enum, true), 'enum', array('enum' => $schema->enum,));
+        $this->addError($path, "value is not in enumeration: [" . $this->enumToString($schema->enum) . "]", 'enum', array('enum' => $schema->enum,));
+    }
+
+    /**
+     * @param array $enum
+     * @return string
+     */
+    private function enumToString(array $enum)
+    {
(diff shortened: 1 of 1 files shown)


src/JsonSchema/Constraints/EnumConstraint.php
```diff
@@ -41,6 +41,31 @@ public function check($element, $schema = null, $path = null, $i = null)
...
+    {
+        $enumString = '';
+
+        foreach ($enum as $value) {
+            if (is_array($value)) {
+                $enumString += '[' . $this->enumToString($value) . ']' . ', ';
+                continue;
+            }
+
+            if (is_object($value)) {
```
bighappyface: Why not cast `$value` to an array instead?

wpillar: Updated.

bighappyface: @wpillar to be honest, I don't really like this approach. I feel like the code could be better, using arrays and `implode` instead of lots of less-than-efficient concatenations and that funny `rtrim` at the end. That being said, if you can rally a few approvals from the community, I won't hold it up.

bighappyface:
@wpillar thanks for the adjustments. I am still on the fence, but I'll merge with "+1" from the community.

PR #198, opened 2015-11-11
Custom constraints

AubreyHewes:
Allows adding custom constraints via

    $validator->addConstraint('name', \Callable|className|instance)

**Background**
I ran in to the problem of custom constraints and via pr #143 by @Maks3w (the factory) the possiblity was available to easily accomplish adding custom constraints..  (this is moved from a @Maks3w fork)

**Implementation**
This allows applying custom constraints, by name, as either a callable/className/instance.
_I am not saying this is how it should be_ --- all input welcome.

i.e. schema

	{ 'myCustomCheck': value }

Can be constrained via `$validator->addConstraint('myCustomCheck', $constraint)`

#### Add a callable constraint

	$validator->addConstraint('test', \Callable);

 * Inherits _current/default_ ctr params (mode, uriRetriever, factory) - what the factory does internally
 * The callable is the `ConstraintInterface->check` signature `function ($value, $schema = null, $path = null, $i = null)`

#### Add by Constraint class-name

	$validator->addConstraint('test', 'FQCN');

 * Inherits _current/default_ ctr params (mode, uriRetriever, factory) - what the factory does internally
 * `FQCN` must be of type `JsonSchema\Constraints\ConstraintInterface`

#### Add by Constraint instance
	
	$validator->addConstraint('test', new MyCustomConstraint(...));

 * Possibly requires adding the correct ctr params (mode, uriRetriever, factory et al) - what the factory does internally; If not the mode will be the default and the uriRetriever/factory will be another instance (if this is a problem remains to be seen)
 * `MyCustomConstraint` must be of type `JsonSchema\Constraints\ConstraintInterface`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,32 @@ if ($validator->isValid()) {
 }
 ```
 
+## Custom Constraints
+Add custom constraints via `$validator->addConstraint($name, $constraint)`;
+
+The given `$constraint` is applied when `$name` is found within the current evaluated schema path.
+
+### Add a callable constraint
+
+	$validator->addConstraint('test', \Callable);
+
+ * Inherits _current_ ctr params (uriRetriever, factory)
+ * The callable is the `ConstraintInterface->check` signature `function check($value, $schema = null, $path = null, $i = null)`
+
+### Add by custom Constraint instance
+	
+	$validator->addConstraint('test', new MyCustomConstraint(...));
+
+ * Requires adding the correct ctr params (uriRetriever, factory et al)
+ * `MyCustomConstraint` must be of type `JsonSchema\Constraints\ConstraintInterface`
+
+### Add by custom Constraint class-name
+
+	$validator->addConstraint('test', 'FQCN');
+
+ * Inherits _current_ ctr params (uriRetriever, factory)
+ * `FQCN` must be of type `JsonSchema\Constraints\ConstraintInterface`
+
 ## Running the tests
 
     $ vendor/bin/phpunit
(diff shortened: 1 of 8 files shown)


bighappyface:
Please explain the need for custom constraints outside of the JSON Schema spec?

Wouldn't a custom constraint in a given schema only work for this validator and not others?

AubreyHewes:
The official spec is too concrete for normal usage; the change is mostly based on (fe) https://github.com/acornejo/jjv#custom-checks

 * support same schema in fe/be

Also slightly based on the "format" spec: 
 * Implementations MAY add custom format attributes. Save for agreement between parties, schema authors SHALL NOT expect a peer implementation to support this keyword and/or custom format attributes. (http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.2)

For optimal schema usage you should be able to define own constraints, outside of the spec;

    { 'myCustomCheck': expectation }

And be able to test your constraint ... this allows this. Even if it is not spec.. it just opens up the possibilities.

Yes you need to create a validator for the required schema validation... This may be different than a previous global validator..(if this is what you meant)? So also enhances creating a new validator specific to the requirement.

AubreyHewes:
The PR is slightly superfluous as you can add your own factory per validator instance.

bighappyface:
@AubreyHewes thanks for your patience and insight. Would you please rebase and squash this PR down to a single commit?

AubreyHewes:
@bighappyface I am leaning towards this being superfluous. As via the factory you can do your own stuff anyway.. Also directly supporting custom attributes/stuff does not help the json spec. 
Own business/domain spec should not be part of a spec validation. I.e. apply jsonschema and then apply own business/domain spec? Will close this.

PR #202, opened 2015-12-02
Fix CollectionConstraint to allow uniqueItems to be false

dicexq:
According to the latest [json-schema-validation](http://json-schema.org/latest/json-schema-validation.html#anchor49).
uniqueItems should be either a boolean or if not present, set to false


bighappyface:
+1

PR #204, opened 2015-12-14
Fix path output for required properties

JustTSK:
Example:

Before:
```day: "The property day is required"```
After
```availability[0].hours[0].day: "The property day is required"```


bighappyface:
+1

Issue #145, opened 2015-04-16
Format constraint and date-time with fractions


jayS-de:
The date-time is not handling the fraction part of a datetime string correctly

To the FormatTest::getValidFormats should be added following cases:

            array('2000-05-01T12:12:12.123Z', 'date-time'),
            array('2000-05-01T12:12:12.123+01:00', 'date-time'),
            array('2000-05-01T12:12:12.123+0100', 'date-time'),
            array('2000-05-01T12:12:12.123456+01:00', 'date-time'),
            array('2000-05-01T12:12:12.123456+0100', 'date-time'),

And in the function FormatConstraint::check the case 'date-time' should be changed to:

            case 'date-time':
                if (!$this->validateDateTime($element, 'Y-m-d\TH:i:sT') &&
                    !$this->validateDateTime($element, 'Y-m-d\TH:i:s.uT') &&
                    !$this->validateDateTime($element, 'Y-m-d\TH:i:sP') &&
                    !$this->validateDateTime($element, 'Y-m-d\TH:i:s.uP') &&
                    !$this->validateDateTime($element, 'Y-m-d\TH:i:sO') &&
                    !$this->validateDateTime($element, 'Y-m-d\TH:i:s.uO')
                ) {
                    $this->addError($path, sprintf('Invalid date-time %s, expected format YYYY-MM-DDThh:mm:ssZ or YYYY-MM-DDThh:mm:ss+hh:mm', json_encode($element)));
                }
                break;

The T is better than \Z because this way the timezone part from the string will be correctly read. Important if the script is not running with UTC timezone. And the FormatConstraint::validateDateTime can now be changed to this, so that the tests will work:

    protected function validateDateTime($datetime, $format)
    {
        $dt = \DateTime::createFromFormat($format, $datetime);

        if (!$dt) {
            return false;
        }
        if (strpos($format, '.u') > 0 ) {
            $dtTest = new \DateTime($datetime);
            return $dt == $dtTest;
        }
        return $datetime === $dt->format($format);
    }

So that when a fraction is included in the datetime string, it will be checked if reading from format and just a plain DateTime instance will have the same value.

bighappyface:
Closed with https://github.com/justinrainbow/json-schema/pull/238

Please re-open if necessary.

PR #207, opened 2015-12-15
Fix non-6 digit microsecond date time formats

johnnoel:
RFC3339 [allows](http://tools.ietf.org/html/rfc3339#section-5.8) for second fractions to be any length, however PHP's date/time formatting ALWAYS prints them out as 6 digits. This means if a date-time is passed with a < 6 digit second fractional, the FormatConstraint::validateDateTime will fail as the formatted date will contain 6 digits.

E.g. '2000-05-01T12:12:12.**123**Z' is passed as a date time and is valid. However after parsing, format() will produce '2000-05-01T12:12:12:12.**123000**Z'.

Fixes issue #145.

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -130,7 +130,19 @@ class FormatConstraint extends Constraint
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


bighappyface:
+1

PR #210, opened 2015-12-25
Fully implement $ref compliance with draft 4

araines:
**This makes public interface breaking changes - should consider bumping the library major version when taking this change**

I tried to make these changes without making too many changes to the existing library, but unfortunately the end result is more changes than I'd really like (including changing some of the unit tests).  However, we now have (local) $ref compatibility and it passes all the tests provided for JSON Schema Draft 4 $refs.

https://github.com/justinrainbow/json-schema/pull/210/files#diff-c1e4a623a503c7f9de75304aae24ae77L21

The major changes here are:
* We can do away with the pesky maximum depth, as now we use object references to achieve recursion where necessary within the schema
* The JSON Pointer resolution has been abstracted in to its own class, with hopefully a clearer and cleaner implementation
* "definitions" is now resolved, although have not been formally fully tested, it is certainly more functional than it once was

I've tried to be respectful to the original library, but appreciate my changes may not be to everyone's taste - comments and improvements are obviously very welcome.

### DIFF
--- a/src/JsonSchema/Pointer.php
+++ b/src/JsonSchema/Pointer.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace JsonSchema;
+
+use JsonSchema\Exception\InvalidPointerException;
+use JsonSchema\Exception\ResourceNotFoundException;
+
+/**
+ * Resolve JSON Pointers (RFC 6901)
+ */
+class Pointer
+{
+    const LAST_ELEMENT = '-';
+    const SEPARATOR = '/';
+
+    /**
+     * @var object
+     */
+    private $json;
+
+    /**
+     * @var string
+     */
+    private $pointer;
+
+    /**
+     * @param object $json The json document to resolve within.
+     */
+    public function __construct($json)
+    {
+        $this->json = $json;
+    }
+
+    /**
+     * Get the part of the document which the pointer points to.
+     *
+     * @param string $pointer The Json Pointer.
+     *
+     * @throws ResourceNotFoundException
+     *
+     * @return mixed
+     */
+    public function get($pointer)
+    {
+        if ($pointer === '') {
+            return $this->json;
+        }
+
+        $this->validatePointer($pointer);
+        $this->pointer = $pointer;
+
+        $parts = array_slice(array_map('urldecode', explode('/', $pointer)), 1);
+
+        return $this->resolve($this->json, $this->decodeParts($parts));
+    }
+
+    /**
+     * Decode any escaped sequences.
+     *
+     * @param array $parts The json pointer parts.
+     *
+     * @return array
+     */
+    private function decodeParts(array $parts)
+    {
+        $mappings = array(
+            '~1' => '/',
+            '~0' => '~',
+        );
+
+        foreach ($parts as &$part) {
+            $part = strtr($part, $mappings);
+        }
+
+        return $parts;
+    }
+
+    /**
+     * Recurse through $json until location described by $parts is found.
+     *
+     * @param mixed $json  The json document.
+     * @param array $parts The parts of the pointer.
+     *
+     * @throws ResourceNotFoundException
+     *
+     * @return mixed
+     */
+    private function resolve($json, array $parts)
+    {
+        // Check for completion
+        if (count($parts) === 0) {
+            return $json;
+        }
+
+        $part = array_shift($parts);
+
+        if (is_object($json) && isset($json->$part)) {
+            return $this->resolve($json->$part, $parts);
+        } elseif (is_array($json)) {
+            if ($part === self::LAST_ELEMENT) {
+                return $this->resolve(end($json));
+            }
+            if (filter_var($part, FILTER_VALIDATE_INT) !== false &&
+                array_key_exists($part, $json)
+            ) {
+                return $this->resolve($json[$part], $parts);
+            }
+        }
+
+        throw new ResourceNotFoundException('Failed to resolve pointer ' . $this->pointer);
+    }
+
+    /**
+     * Validate a pointer string.
+     *
+     * @param string $pointer The pointer to validate.
+     *
+     * @throws InvalidPointerException
+     */
+    private function validatePointer($pointer)
+    {
+        if ($pointer !== '' && !is_string($pointer)) {
+            throw new InvalidPointerException('Pointer is not a string');
+        }
+
+        $firstCharacter = substr($pointer, 0, 1);
+
+        if ($firstCharacter !== self::SEPARATOR) {
+            throw new InvalidPointerException('Pointer starts with invalid character');
+        }
+    }
+
+}
--- a/src/JsonSchema/Pointer.php
+++ b/src/JsonSchema/Pointer.php
@@ -10,6 +10,7 @@ use JsonSchema\Exception\ResourceNotFoundException;
  */
 class Pointer
 {
+    const EMPTY_ELEMENT = '_empty_';
     const LAST_ELEMENT = '-';
     const SEPARATOR = '/';
 
@@ -36,6 +37,7 @@ class Pointer
      *
      * @param string $pointer The Json Pointer.
      *
+     * @throws InvalidPointerException
      * @throws ResourceNotFoundException
      *
      * @return mixed
@@ -94,11 +96,16 @@ class Pointer
 
         $part = array_shift($parts);
 
-        if (is_object($json) && isset($json->$part)) {
+        // Ensure we deal with empty keys the same way as json_decode does
+        if ($part === '') {
+            $part = self::EMPTY_ELEMENT;
+        }
+
+        if (is_object($json) && property_exists($json, $part)) {
             return $this->resolve($json->$part, $parts);
         } elseif (is_array($json)) {
             if ($part === self::LAST_ELEMENT) {
-                return $this->resolve(end($json));
+                return $this->resolve(end($json), $parts);
             }
             if (filter_var($part, FILTER_VALIDATE_INT) !== false &&
                 array_key_exists($part, $json)
@@ -107,7 +114,9 @@ class Pointer
             }
         }
 
-        throw new ResourceNotFoundException('Failed to resolve pointer ' . $this->pointer);
+        $message = 'Failed to resolve pointer ' . $this->pointer .
+            ' from document id ' . (isset($json->id) ? $json->id : '');
+        throw new ResourceNotFoundException($message);
     }
 
     /**
--- a/src/JsonSchema/PointerResolver.php
+++ b/src/JsonSchema/PointerResolver.php
@@ -8,16 +8,11 @@ use JsonSchema\Exception\ResourceNotFoundException;
 /**
  * Resolve JSON Pointers (RFC 6901)
  */
-class Pointer
+class PointerResolver
 {
     const EMPTY_ELEMENT = '_empty_';
     const LAST_ELEMENT = '-';
     const SEPARATOR = '/';
-
-    /**
-     * @var object
-     */
-    private $json;
 
     /**
      * @var string
@@ -25,16 +20,9 @@ class Pointer
     private $pointer;
 
     /**
-     * @param object $json The json document to resolve within.
-     */
-    public function __construct($json)
-    {
-        $this->json = $json;
-    }
-
-    /**
      * Get the part of the document which the pointer points to.
      *
+     * @param object $json    The json document to resolve within.
      * @param string $pointer The Json Pointer.
      *
      * @throws InvalidPointerException
@@ -42,10 +30,10 @@ class Pointer
      *
      * @return mixed
      */
-    public function get($pointer)
+    public function resolvePointer($json, $pointer)
     {
         if ($pointer === '') {
-            return $this->json;
+            return $json;
         }
 
         $this->validatePointer($pointer);
@@ -53,7 +41,7 @@ class Pointer
 
         $parts = array_slice(array_map('urldecode', explode('/', $pointer)), 1);
 
-        return $this->resolve($this->json, $this->decodeParts($parts));
+        return $this->resolve($json, $this->decodeParts($parts));
     }
 
     /**
--- a/src/JsonSchema/PointerResolver.php
+++ b/src/JsonSchema/PointerResolver.php
@@ -15,11 +15,6 @@ class PointerResolver
     const SEPARATOR = '/';
 
     /**
-     * @var string
-     */
-    private $pointer;
-
-    /**
      * Get the part of the document which the pointer points to.
      *
      * @param object $json    The json document to resolve within.
@@ -37,11 +32,10 @@ class PointerResolver
         }
 
         $this->validatePointer($pointer);
-        $this->pointer = $pointer;
 
         $parts = array_slice(array_map('urldecode', explode('/', $pointer)), 1);
 
-        return $this->resolve($json, $this->decodeParts($parts));
+        return $this->resolve($json, $pointer, $this->decodeParts($parts));
     }
 
     /**
@@ -68,14 +62,15 @@ class PointerResolver
     /**
      * Recurse through $json until location described by $parts is found.
      *
-     * @param mixed $json  The json document.
-     * @param array $parts The parts of the pointer.
+     * @param mixed  $json  The json document.
+     * @param string $json  The original json pointer.
+     * @param array  $parts The (remaining) parts of the pointer.
      *
      * @throws ResourceNotFoundException
      *
      * @return mixed
      */
-    private function resolve($json, array $parts)
+    private function resolve($json, $pointer, array $parts)
     {
         // Check for completion
         if (count($parts) === 0) {
@@ -90,20 +85,20 @@ class PointerResolver
         }
 
         if (is_object($json) && property_exists($json, $part)) {
-            return $this->resolve($json->$part, $parts);
+            return $this->resolve($json->$part, $pointer, $parts);
         } elseif (is_array($json)) {
             if ($part === self::LAST_ELEMENT) {
-                return $this->resolve(end($json), $parts);
+                return $this->resolve(end($json), $pointer, $parts);
             }
             if (filter_var($part, FILTER_VALIDATE_INT) !== false &&
                 array_key_exists($part, $json)
             ) {
-                return $this->resolve($json[$part], $parts);
+                return $this->resolve($json[$part], $pointer, $parts);
             }
         }
 
-        $message = 'Failed to resolve pointer ' . $this->pointer .
-            ' from document id ' . (isset($json->id) ? $json->id : '');
+        $message = "Failed to resolve pointer $pointer from document id"
+            . (isset($json->id) ? $json->id : '');
         throw new ResourceNotFoundException($message);
     }
 
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -22,6 +22,8 @@ use JsonSchema\Uri\UriResolver;
  */
 class RefResolver
 {
+    const SELF_REF_LOCATION = '#';
+
     /**
      * @var UriRetrieverInterface
      */
@@ -64,7 +66,7 @@ class RefResolver
 
         // Retrieve dereferenced schema
         if ($location == null) {
-            $schema = end($this->schemas);
+            $schema = $this->schemas[self::SELF_REF_LOCATION];
         } elseif (array_key_exists($location, $this->schemas)) {
             $schema = $this->schemas[$location];
         } else {
@@ -166,8 +168,8 @@ class RefResolver
     private function enterResolutionScope($schemaPartial, $sourceUri)
     {
         if (count($this->scopes) === 0) {
-            $this->scopes[] = '#';
-            $this->schemas[] = $schemaPartial;
+            $this->scopes[] = self::SELF_REF_LOCATION;
+            $this->schemas[self::SELF_REF_LOCATION] = $schemaPartial;
         }
 
         if (!empty($schemaPartial->id)) {
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -127,7 +127,7 @@ class RefResolver
         }
 
         // First determine our resolution scope
-        $scope = $this->getResolutionScope($schema, $sourceUri);
+        $scope = $this->enterResolutionScope($schema, $sourceUri);
 
         // These properties are just schemas
         // eg.  items can be a schema or an array of schemas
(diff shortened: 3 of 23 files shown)


bighappyface:
@araines thank you SO MUCH for tackling this, and thanks for your patience. Please rebase this PR on the latest master and squash your commits down. The travis build should be fixed now and it will be good to see this PR pass so that we can focus in on review.

araines:
@bighappyface rebased and squashed

Cheater912:
Awesome, thanks for this.

src/JsonSchema/Pointer.php
```diff
@@ -0,0 +1,133 @@
...
+                return $this->resolve(end($json));
+            }
+            if (filter_var($part, FILTER_VALIDATE_INT) !== false &&
+                array_key_exists($part, $json)
+            ) {
+                return $this->resolve($json[$part], $parts);
+            }
+        }
+
+        throw new ResourceNotFoundException('Failed to resolve pointer ' . $this->pointer);
```
mathroc: would be nice to add `$json->id` (if present) here

mathroc:
Looks very nice. Do you think support for `$ref` in [JSON Hyper-Schema](http://json-schema.org/latest/json-schema-hypermedia.html) could be added as well ? 

I think those are the only possibilities added by JSON Hyper-Schema:
```json
{
	"links" : [
		{
			"schema": {
				"$ref": "/a/schema"
			},
			"targetSchema": {
				"$ref": "/another/schema"
			}
		}
	]
}
```

mathroc:
Hi,

I tried to support the json hyper schema. it seems to be enough to add this :

```php

    public function resolve($schema, $sourceUri = null)
    {
        // [snip]
  
        // Resolve links
        $this->resolveLinks($schema, $scope);

        // Resolve $ref
        $this->resolveRef($schema, $scope);

        // Pop back out of our scope
        array_pop($this->scopes);
    }

    /**
     * Look for the links property in the object. If found, look for schema
     * and targetSchema and resolves them.
     *
     * @param object $schema    JSON Schema to flesh out
     * @param string $sourceUri URI where this schema was located
     */
    public function resolveLinks($schema, $sourceUri)
    {
        if (! isset($schema->links)) {
            return;
        }

        foreach ($schema->links as $link) {
            $this->resolveProperty($link, 'schema', $sourceUri);
            $this->resolveProperty($link, 'targetSchema', $sourceUri);
        }
    }
```

@araines do you think you can add this in your PR ?

mathroc:
and another topic, I have a problem resolving this schema:

```json
{
  "type":"object",
  "definitions": {
    "schema": {
      "$ref": "http://json-schema.org/draft-04/hyper-schema"
    }
  },
  "properties": {
    "schema": {
      "$ref": "#/definitions/schema"
    }
  },
  "required": ["schema"]
}
```

here is an extract of the stacktrace : 

```
Fatal error: Maximum function nesting level of '256' reached, aborting! in /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/Uri/UriResolver.php on line 143

Call Stack:
    0.0007     462216   1. {main}() /scripts/vendor/justinrainbow/json-schema/bin/validate-json:0
    0.0041     677568   2. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/bin/validate-json:218
    0.6655     798864   3. JsonSchema\RefResolver->resolveObjectOfSchemas() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:147
    0.6655     799176   4. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:219
    0.6655     799400   5. JsonSchema\RefResolver->resolveRef() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:154
    0.6656     799448   6. JsonSchema\RefResolver->fetchRef() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:277
    0.6657     799968   7. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:75
    0.6657     800816   8. JsonSchema\RefResolver->resolveObjectOfSchemas() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:147
    0.6658     804280  13. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:219
    0.6659     805128  14. JsonSchema\RefResolver->resolveProperty() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:135
    0.6659     805128  15. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:237
    0.6659     805872  16. JsonSchema\RefResolver->resolveObjectOfSchemas() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:147
[snip]
    0.6720    1082456 253. JsonSchema\RefResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:219
    0.6720    1082456 254. JsonSchema\RefResolver->getResolutionScope() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:130
    0.6720    1082552 255. JsonSchema\Uri\UriResolver->resolve() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/RefResolver.php:177
    0.6720    1082552 256. JsonSchema\Uri\UriResolver->parse() /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/Uri/UriResolver.php:37
```

but it works with this one : 

```json
{
  "type":"object",
  "properties": {
    "schema": {
      "$ref": "http://json-schema.org/draft-04/hyper-schema"
    }
  },
  "required": ["schema"]
}
```

not sure where the problem comes from

src/JsonSchema/RefResolver.php
```diff
@@ -101,47 +117,66 @@ public function getUriRetriever()
...
         // These are all objects containing properties whose values are schemas
-        foreach (array('dependencies', 'patternProperties', 'properties') as $propertyName) {
-            $this->resolveObjectOfSchemas($schema, $propertyName, $sourceUri);
+        foreach (array('definitions', 'dependencies', 'patternProperties', 'properties') as $propertyName) {
+            $this->resolveObjectOfSchemas($schema, $propertyName, $scope);
         }
 
-        --self::$depth;
+        // Resolve $ref
+        $this->resolveRef($schema, $scope);
```
Markard: I guess you should resolveRef before other resolutions. Before 134 line.

webmozart:
What's the state of this?

bighappyface:
@webmozart good question. Would you mind reviewing the PR and feedback from @Markard and please post your thoughts?

Once the PR gains a couple of +1s I can proceed with the merge and bump.

src/JsonSchema/Pointer.php
```diff
@@ -0,0 +1,133 @@
...
+
+namespace JsonSchema;
+
+use JsonSchema\Exception\InvalidPointerException;
+use JsonSchema\Exception\ResourceNotFoundException;
+
+/**
+ * Resolve JSON Pointers (RFC 6901)
+ */
+class Pointer
```
webmozart: Is there any test case for this class?

src/JsonSchema/RefResolver.php
```diff
@@ -101,47 +117,66 @@ public function getUriRetriever()
...
+
+    /**
+     * Returns the resolution scope for the given schema. Inspects the partial
+     * for the presence of 'id' and then returns that as a absolute uri.
+     *
+     * @param  object $schemaPartial JSON Schema to get the resolution scope for
+     * @param  string $sourceUri     URI where this schema was located
+     * @return string
+     */
+    private function getResolutionScope($schemaPartial, $sourceUri)
```
webmozart: I don't think this "getter" should change internal state. Should this method maybe be split into separate methods for writing and reading?

src/JsonSchema/Uri/UriResolver.php
```diff
@@ -81,77 +34,124 @@ public function resolve($uri, $baseUri = null)
...
+        }
+        if (isset($parts['fragment'])) {
+            $uri['fragment'] = $parts['fragment'];
         }
 
-        $basePathSegments = explode('/', $basePath);
+        return $this->build($uri);
+    }
+
+    public function extractLocation($uri)
```
webmozart: Should this really be public?

src/JsonSchema/Uri/UriResolver.php
```diff
@@ -81,77 +34,124 @@ public function resolve($uri, $baseUri = null)
...
+    {
+        $parts = $this->parse($uri);
+        unset($parts['fragment']);
+
+        return $this->build($parts);
+    }
+
+    public function extractFragment($uri)
```
webmozart: Same here

src/JsonSchema/Uri/UriResolver.php
```diff
@@ -81,77 +34,124 @@ public function resolve($uri, $baseUri = null)
...
-        $path = preg_replace('|^/?(\.\./(\./)*)*|', '', $relativePath);
+        return $fragment;
+    }
 
-        return implode('/', $basePathSegments) . '/' . $path;
+    /**
+     * @param string $uri
+     * @return boolean
+     */
+    public function isValid($uri)
```
webmozart: Same here

tests/JsonSchema/Tests/Drafts/Draft4Test.php
```diff
@@ -18,7 +18,6 @@ protected function getSkippedTests()
             // Not Yet Implemented
             'definitions.json',
             // Partially Implemented
-            'ref.json',
```
webmozart: Is this line supposed to be removed?

araines:
Sorry guys, been v busy recently and haven't had a chance to come back to comment on and fix up review comments.  I'll try to find some time over the next few days to go through everything and tidy up.

bighappyface:
@araines no worries. @Markard and @webmozart  have been super helpful and I think we can get this one across the finish line soon.

tests/JsonSchema/Tests/Drafts/Draft4Test.php
araines: Yes - this is reenabling the tests for ref.json - proving that the changes are tested and working correctly.

webmozart: Ah I didn't see this is in `getSkippedTests()`. Makes sense, thanks. :)

src/JsonSchema/RefResolver.php
araines: The reason for doing it afterward is to ensure we don't get stuck in an infinite loop whilst attempting to dereference everything.

src/JsonSchema/RefResolver.php
araines: I'm not sure I understand what you mean?

Just in case this helps - the internal state is being used as a means by which to track what references and schemas we have seen before so that we can maintain object references (pointers) so that we can fully resolve recursive schema.

src/JsonSchema/Uri/UriResolver.php
araines: It is used by the RefResolver

src/JsonSchema/Uri/UriResolver.php
araines: It is used by the RefResolver

src/JsonSchema/Uri/UriResolver.php
araines: It was public before, I was trying to avoid making too many public interface changes where possible.

src/JsonSchema/Pointer.php
araines: Very good point, in putting the tests together I've identified a number of bugs!

src/JsonSchema/Pointer.php
araines: done

araines:
@mathroc Regarding HyperSchema, I think it might be best left to a different PR.  If I add the code in for it as suggested, I'll have to write some tests for it as well.  I'm not flush with time at the moment and I don't need the HyperSchema stuff to work myself - but perhaps if you could tackle that one?

In terms of your failing ref resolution - that is resolving a remote ref, which I haven't been looking at directly.  They fall under the tests which are 'not fully implemented', so its possible that is where the problem lies.

araines:
@bighappyface / @mathroc / @webmozart - I think I've addressed all the comments now (or at least responded), just let me know if there is more that wants doing!  Cheers.

mathroc:
@araines thank you, this PR is a huge improvement already. I'll submit different PRs for the HyperSchema and `refRemote.json` (maybe `definitions.json` too as I think it should work too once remote refs are OK)
:+1:

src/JsonSchema/Pointer.php
webmozart: Cool :) I think it's a bit weird that this code lives in a class named `Pointer`. Objects of this class do not actually represent a pointer (e.g. `/data/2`), but the JSON data. Also, I don't understand why this class stores any state (`$json`).

In my opinion, this should be a stateless service class with a method `resolvePointer($json, $pointer)`. Following this thought, I think that this code could be just as well merged back into `UriRetriever`.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -140,10 +125,10 @@ public function retrieve($uri, $baseUri = null)
         $resolvedUri = $fetchUri = $resolver->resolve($uri, $baseUri);
```
webmozart: I think that support for `$baseUri` should be dropped from this method. Always passing URIs makes the code much clearer and removes the need to construct a `UriResolver` upon every call.

The `UriResolver` can be created (and cached) in the calling code.

src/JsonSchema/Pointer.php
webmozart: Alternatively, we could call this `PointerResolver` with a single method `resolvePointer($json, $pointer)`. I'd actually prefer this since the URI retriever's responsibility is loading schemas, not dealing with pointers.

src/JsonSchema/RefResolver.php
webmozart: Yes exactly :) The problem is that the code does not convey that intent. Compare:

~~~php
$scope1 = $this->getResolutionScope(...);
$scope2 = $this->getResolutionScope(...);
// huh, scope changed twice?
// $scope1 !== $scope?
~~~

with

~~~php
$this->enterScope(...);
$scope1 = $this->getScope();

$this->enterScope(...);
$scope2 = $this->getScope();
$this->leaveScope(...);

$this->leaveScope(...);
~~~

This is just an example, but it should make clear how separating methods that *change* state from methods that *read* state can greatly improve readability (under "readability" I understand ease of understanding what's going on, not utmost conciseness).

webmozart:
*If* we bump the major version, I suggest cleaning up a few more things that are a bit weird in this library:

* There's duplicate code in `UriRetriever` and `UriResolver` (`parse()`, `generate()`, `resolve()`, `isValid()`). The duplicate code should be removed by calling the methods of one class in the other class. I suggest dropping them from `UriRetriever` whose responsibility is something else (retrieving schemas).

* The `$baseUri` should be dropped from `UriRetriever::retrieve()`. The caller should make sure to resolve URIs before calling this method.

* `UriRetriever::resolvePointer()` should not be public or even removed if we add a `PointerResolver`.

* `UriRetriever::confirmMediaType()` should not be public. The first argument `$uriRetriever` can probably be removed too.

* The fact that `UriRetriever` has a method `setUriRetriever()` which accepts something else is irritating.

Fixing issues like these should make contributing and changes a bit easier than they are right now.

webmozart:
I think it might also be worthwhile to replace `UriResolver` by a new final utility class `Uri` with static methods. `UriResolver` is always created inline (and never injected, which also wouldn't make sense) and its methods are stateless, so there's no reason in creating a new object every time. Also, the name `UriResolver` is very easy to confuse with `UriRetriever` (happened to me often enough).

`Uri::parse()` and `Uri::generate()` reads much better. `resolve()` could be named `Uri::makeAbsolute()`.

src/JsonSchema/Pointer.php
araines: As you identified - the intention of this class was to keep all pointer resolution code encapsulated in one place whose only responsibility is to resolve pointers.  My choice of naming is definitely not ideal, in retrospect!

The other part of the thinking was to have the JSON document as part of the state of the class as that allows to resolve multiple pointers from a single document, but there clearly isn't a need for that, and it does make for a slightly confusing usage pattern.  I'll rename and make the changes as you've described.

bighappyface:
@webmozart thanks for your feedback. As this is a community-driven project, all PRs are welcome, so the various recommendations you have would be best conveyed as individual contributions as there is no official "maintainer" to put it on any given roadmap (I just happen to have push access to this repo).

Regarding those various recommendations, I think it is good to focus on one iteration at a time. Not every PR need be a masterpiece. Provided that we bump to `2.0.0` with the merge of this PR (which is the tentative plan) we can then iterate further based on your recommendations and those from others. This will help folks like @araines have their contributions accepted in timely fashions and a keep this package progressing.

A lot of people have been asking for better `$ref` handling for awhile and this work gets us distinctly closer, so I want to work with all of us to get to a happier place, and reviews like this go a long way toward that goal. That being said, what are the minimum additional changes to this PR that you would like to see to obtain your "+1" so that we can move forward?

webmozart:
@bighappyface Yes, I was mainly dropping my thoughts here not to loose them. :) Let's keep this PR focused!

src/JsonSchema/RefResolver.php
araines: I see what you mean, totally agree.

src/JsonSchema/Uri/UriResolver.php
webmozart: What do you think about moving this and `extractFragment()` to a private method of `RefResolver`? I don't see much reusability here.

araines:
I've added changes to aid with code clarity which were directly related to the changes I've made, but as already discussed, I've left the rest for another PR :)

src/JsonSchema/Uri/UriResolver.php
araines: Honestly, I think the whole library could do with a good reshuffling and cleaning up.  There were actually big chunks of code in there which look like copy-paste jobs, I think slightly better since my changes.

Ultimately the validation side of the library seems to be solid and well laid out, but the resolution side could do with some real love.  As I was making these changes, I did consider radical alterations but decided against in favour of trying to get something which would work and would require the minimum number of test changes so that everyone could have some confidence that it worked!

src/JsonSchema/PointerResolver.php
```diff
@@ -0,0 +1,130 @@
...
+     * @return mixed
+     */
+    public function resolvePointer($json, $pointer)
+    {
+        if ($pointer === '') {
+            return $json;
+        }
+
+        $this->validatePointer($pointer);
+        $this->pointer = $pointer;
```
webmozart: I think it would be better to pass this pointer to `resolve()` instead of storing it. We don't really need to store this for any other reason but showing an exception message.

src/JsonSchema/RefResolver.php
```diff
@@ -101,47 +117,75 @@ public function getUriRetriever()
...
-            $this->rootSchema = $schema;
+        // Fill in id property
+        if ($sourceUri) {
+            $schema->id = $sourceUri;
         }
 
-        // Resolve $ref first
-        $this->resolveRef($schema, $sourceUri);
+        // First determine our resolution scope
+        $scope = $this->enterResolutionScope($schema, $sourceUri);
```
webmozart: :+1: :)

src/JsonSchema/RefResolver.php
```diff
@@ -63,11 +54,36 @@ public function __construct($retriever = null)
...
+        $resolver = new UriResolver();
+        $uri = $resolver->resolve($ref, $sourceUri);
+
+        // Split in to location and fragment
+        $location = $resolver->extractLocation($uri);
+        $fragment = $resolver->extractFragment($uri);
+
+        // Retrieve dereferenced schema
+        if ($location == null) {
+            $schema = end($this->schemas);
```
webmozart: I would use a separate variable for storing this. Currently, `$schemas` is used as associative array (map) from Uri => schema object. Using the same data structure also as stack is prone to errors.

We could simply keep `$rootSchema` for this.

src/JsonSchema/Uri/UriResolver.php
webmozart: You are right :) But quite for that reason I suggest not to introduce any more public methods unless really needed.

bighappyface:
@araines @webmozart @mathroc 

Hey folks, are we good on this or are there still pending changes?

I'm going to get caught up on this soon but want to ask if we know anything is outstanding before I dive in once again.

bighappyface:
@rtucek would you mind reviewing this PR as a return favor to your recently accepted work in #232 ?

mathroc:
I'm ok with the state of this PR. but I'd suggest not to make a new release immediately after merging. It could give some time to make some BC improvement as suggested by @webmozart
@webmozart woudl you mind creating a few issue based on your earlier comments ? (maybe they could be put into a 2.0 roadmap). I'll be happy to work on some of them

rtucek:
@bighappyface yeah, sure no problem.
But this PR is huge and it takes some time to pick the pieces. I'll look into this during weekend.

webmozart:
I would rather see this PR finished instead of merged + fixes later on. Otherwise you run the risk of publishing undesired code with the next release and having to maintain BC for that code.

bighappyface:
@rtucek thank you for the help!

@webmozart I would like to see this PR finished as well.

araines:
It looks like there are still some comments here to be addressed.  I'm happy to do so, but I'm afraid my time gets torn between a lot of different things!  I'll try to get on to some of the remaining comments as soon as I can.

Assuming everyone is happy with taking on the few remaining comments from @webmozart which involve making further BC breaking changes (although ultimately for the better I believe), then I should be able to get them sorted out in the next few days.

src/JsonSchema/RefResolver.php
araines: I thought this was a reasonably elegant way to deal with this problem, but you are probably right - when it comes to extending it in future it does require a fairly good understanding of what is going on - which will inevitably lead to bugs.

I'm opting to use `$schemas` purely as an associative lookup, and rather than keeping track of another state variable (`$rootSchema`), I think it keeps the code a bit cleaner to simply setup the root point within the associative array.  Let me know if you think otherwise, though.

src/JsonSchema/Uri/UriResolver.php
araines: I've just been pondering over this again - it seems like it makes sense to keep the URI resolution within one class away from reference resolution (as a separation of concerns).

In addition, if we were to move these to the RefResolver then we'd either have to move the `build` and `parse` methods over too, or make them public.

I'm not saying I'm totally against the idea, but I'm trying to consider how best to organise it :)

TerjeBr:
Have a look at the test file I added in this commit https://github.com/justinrainbow/json-schema/pull/243/commits/e3f8e1176c438c773e783ee275fc80284590c27c

I will be working to rectify these errors.

araines:
I'm at a bit of a loss at the moment as to why these tests are failing - I've got them passing on my local machine.  Will have to try to understand what is going on better, then get these issues resolved.

TerjeBr:
The tests that are failing is from `vendor/json-schema/JSON-Schema-Test-Suite/tests/draft4/ref.json`
This is the new test file that was included.

It is the test at lines 129-142 that fails.

The test tries to validate against the schema at http://json-schema.org/draft-04/schema#
and this fails because of a bug in `JsonSchema/Constraints/ObjectConstraint.php` at line163

jojo1981:
@araines I think this PR can be closed?

bighappyface:

PR #213, opened 2016-01-06
Fix exclusive min/max strictness

### LINKED ISSUE #199: NumberConstraint allows 0.00 for an exclusive minimum of 0
When validating this:

```json
{
  "amount": 0.00
}
```

Against this:

```json
{
  "properties": {
    "amount": {
      "type": "number",
      "minimum": 0,
      "exclusiveMinimum": true
    }
  }
}
```

It passes, when it should fail according to this json schema linter:

image

http://jsonschemalint.com/draft4/

This is because of the strict equality check here: https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/NumberConstraint.php#L28

wpillar:
Fixes #199.


bighappyface:
+1

PR #216, opened 2016-01-14
improve RefResolver

mathroc:
improve a few case resolving `$ref`s

not done as well as #210 but does not introduce BC so it could be useful before having a new major version

### DIFF
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -99,7 +99,7 @@ class RefResolver
      * @param object $schema    JSON Schema to flesh out
      * @param string $sourceUri URI where this schema was located
      */
-    public function resolve($schema, $sourceUri = null)
+    public function resolve($schema, $sourceUri = null, $rootSchema = null)
     {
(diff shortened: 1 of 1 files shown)


bighappyface:
Hey @mathroc thanks for the contribution! While I understand the value in continuing to improve `RefResolver` while maintaining backward compatibility, I think we are better served by focusing on #210 and giving this package a shot at a comprehensive fix, even if that means a major version bump.

May @araines and I have your assistance reviewing #210 to ensure it is ready to rock so that I can proceed with the next version? I very much value your work here and want to try to get your help while you are willing to give it to help this package move in the right direction sooner rather than later.

mathroc:
@bighappyface I understand, it's just that made thoses changes before seeing #210 and submitted this in case you'd rather not have a major version immediately. I'm actually using @araines fork and my use cases are also working with his PR :)
I'm gooing to take a closer look at #210 and will tell you if I have comments

(closing this one then)

bighappyface:
@mathroc :+1:

PR #218, opened 2016-01-18
feat(composer): require PHP 5.3.29 (PHP 5.3.x EOL version) minimum

bighappyface:
With PHP 5.3.x EOL as of 8/14/2014 it is time to remove support for PHP 5.3.x versions prior to the last version.

https://secure.php.net/releases/5_3_29.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: php
 matrix:
   fast_finish: true
   include:
-    - php: 5.3.3
     - php: 5.3
     - php: 5.4
     - php: 5.5
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         }
     }],
     "require": {
-        "php": ">=5.3.2"
+        "php": ">=5.3.29"
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.1.0",


stof:
given that composer defines the platform php version to 5.3.9 (its min version), it means it will never use 1.6.1+ versions and so won't have bugfixes.
So I suggest reverting the requirement change (there is nothing actually incompatible with older PHP versions in the library AFAICT)

bighappyface:
Hey folks,

PHP 5.3 is long gone and anything less than 5.3.30 isn't recommended for anyone.

Considering that you, the users, are consuming this package on older platforms, I recommend simply skipping the platform requirements during install:

http://stackoverflow.com/questions/26165101/skip-composer-php-requirement

stof:
@bighappyface I'm talking about the building of the Composer dependencies themselves, which target the older version.
and the composer requirement is not about a recommended version. It is about a required version. Most projects I maintained are compatible with 5.3+ or 5.4+, but the recommended version is still to use the latest (I would recommend 5.6 or 7 for now, at the latest patch version of course).

bighappyface:
@stof Checkout https://github.com/justinrainbow/json-schema/pull/231

We are good to go.

PR #221, opened 2016-01-24
feat(Factory): add setConstraintClass

bighappyface:
Adds support for setting arbitrary constraint classes by constraint name to the constraint factory.

Similar to https://github.com/justinrainbow/json-schema/pull/198 the user can assign custom constraints or override existing constraints. The goal is to allow users the ability to customize behavior a bit further without having to fork this repo and hack up the code to customize things for their needs.


bighappyface:
See conversation in #220

PR #222, opened 2016-01-24
feat(travis): pivot to container-based infrastructure

bighappyface:
https://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure

PR #223, opened 2016-01-25
Fix: Remove conditions never evaluating to true

localheinz:
This PR

* [x] removes conditions from `.travis.yml` which will never evaluate to `true`

Follows #222.


bighappyface:
+1

Issue #215, opened 2016-01-13
Bug: incorrect handling of minProperties and maxProperties in optional attribute


rtucek:
Given schema:

```json
{
  "type": "object",
  "additionalProperties": false,
  "required": [
    "session",
    "include_hotel",
    "arrival",
    "departure",
    "ip",
    "travellers",
    "language_iso_code"
  ],
  "properties": {
    "session": {
      "type": "string"
    },
    "campaign": {
      "type": "integer"
    },
    "service_provider": {
      "type": "integer"
    },
    "include_hotel": {
      "type": "boolean"
    },
    "arrival": {
      "type": "string",
      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
    },
    "departure": {
      "type": "string",
      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
    },
    "ip": {
      "type": "string"
    },
    "offerId": {
      "type": "integer",
      "minimum": 1
    },
    "quiet": {
      "type": "boolean"
    },
    "language_iso_code": {
      "type": "string"
    },
    "travellers": {
      "type": "array",
      "minItems": 1,
      "maxItems": 16,
      "items": {
        "type": "integer"
      }
    },
    "rooms": {
      "type": "object",
      "additionalProperties": false,
      "minProperties": 1,
      "maxProperties": 2,
      "properties": {
        "size_1": {
          "type": "number"
        },
        "size_2": {
          "type": "number"
        }
      }
    },
    "offset": {
      "type": "integer"
    }
  }
}
```
And this data:
```json
{
  "campaign": 1,
  "offset": 0,
  "service_provider": 3,
  "include_hotel": false,
  "arrival": "2016-01-20",
  "departure": "2016-01-27",
  "session": "foo",
  "ip": "bar",
  "language_iso_code": "de",
  "travellers": [
    18,
    18,
    18
  ]
}
```

Proofed by http://jsonschemalint.com/draft4/ the json is valid.
However the validator is returning the following error:
```php
[
    0 =>  [
        "property" => "rooms",
        "message" => "Must contain no more than 2 properties",
        "constraint" => "maxProperties",
        "maxProperties" => 2
    ]
];
```

This should not happen because the property rooms is not given.
Thus their is no reason for a check on maxProperties constraint.

bighappyface:
@rtucek good find. Pull requests welcome :wink:

rtucek:
Here we got #224:
[...]

PR #224, opened 2016-02-01
Bugfix for #215

rtucek:
Fixed bug, discovered in #215. 
Also added regression test.

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -166,7 +166,7 @@ class UndefinedConstraint extends Constraint
         }
 
         // Verify minimum and maximum number of properties
-        if (is_object($value)) {
+        if (is_object($value) && !($value instanceof UndefinedConstraint)) {
             if (isset($schema->minProperties)) {
                 if (count(get_object_vars($value)) < $schema->minProperties) {
                     $this->addError($path, "Must contain a minimum of " . $schema->minProperties . " properties", 'minProperties', array('minProperties' => $schema->minProperties,));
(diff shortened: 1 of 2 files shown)


bighappyface:
@rtucek thanks for the contribution!

A couple of things:

# Would you please squash this PR down to a single commit?
# Would you please simplify the test schema and document to focus only on demonstrating the code change? Please review the other test cases for this constraint and/or tests in the official test suite (https://github.com/json-schema/JSON-Schema-Test-Suite)

rtucek:
Hi @bighappyface,

thank your for your quick reply.

1) Yeah, that should have been done in the first place - sry.
2) Well, I personally don't agree with simplifying the test. It was a complex schema and JSON that uncovered the bug...

Anyhow, I did the rebase as requested and it passed the test.

bighappyface:
@rtucek thanks for the update.

Upon deeper review of the changes, I think we have misplaced the `minProperties/maxProperties` validation logic in `UndefinedConstraint` when it should be within `ObjectConstraint` as those are validation keywords of objects (http://json-schema.org/latest/json-schema-validation.html#anchor53).

If the validation were located in the appropriate place then the negation added in this PR would probably be unnecessary. Further, this package already validates all of the `minProperties/maxProperties` testcases provided by the JSON Schema Test Suite:

https://github.com/json-schema/JSON-Schema-Test-Suite/blob/1.1.0/tests/draft4/minProperties.json
https://github.com/json-schema/JSON-Schema-Test-Suite/blob/1.1.0/tests/draft4/maxProperties.json

Those test cases demonstrate the constraint validation specifically. That is what I meant by " please simplify the test schema and document to focus only on demonstrating the code change." The latest update to the new test case seems a bit odd to me because it's redundant and the "document" is an empty object literal (`{}`) that doesn't demonstrate the schema provided afterward.

Would you please look at migrating the `minProperties/maxProperties` validation logic to the `ObjectConstraint` class to see if it works are expected with your original schema and JSON document?

rtucek:
Yeah, it would definitely a true fix for the issue - rather than just a dirty one-line quicky :smile: 
Probably I get this done until the end of the week.

PR #225, opened 2016-02-01
Fixed references to schemas that contain relative references

webmozart:
This PR fixes references to schemas that contain relative references themselves. Currently, only the root schema was used to resolve relative references. Now, the current schema is remembered when fetching other schemas and used as context for relative references.

### DIFF
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -11,6 +11,7 @@ namespace JsonSchema;
 
 use JsonSchema\Exception\JsonDecodingException;
 use JsonSchema\Uri\Retrievers\UriRetrieverInterface;
+use JsonSchema\Uri\UriResolver;
 use JsonSchema\Uri\UriRetriever;
 
 /**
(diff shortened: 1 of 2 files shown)


bighappyface:
@webmozart thanks for your contribution!

The Draft4 tests include some ref tests that are currently skipped:

https://github.com/justinrainbow/json-schema/blob/f3380ebccd4cd5d17a25dd249d13029aaf210606/tests/JsonSchema/Tests/Drafts/Draft4Test.php#L21-L22

Do these changes help fix failing tests in those test suites?

Forgive me for not checking myself, but the way I would review is:

1. Checkout `master`
2. Comment out the two Draft4 ref tests
3. Run `phpunit` to see the current failures,
4. Bring in the changes to `RefResolver`
5. Run `phpunit` again to see if any tests start passing

Currently there is an open PR that overhauls all `$ref` handling and brings all tests to passing: https://github.com/justinrainbow/json-schema/pull/210

Because that PR came in first, I would prefer to give that work priority, but if this work solves many of outstanding `$ref` issues then I would prefer to go with changes that are backward compatible.

webmozart:
#210 looks like a better implementation. No, I did not fix all tests, but I'm very happy if the ref-handling gets improved! :)

PR #226, opened 2016-02-02
Fix: Attempt to cache dependencies installed with composer

localheinz:
This PR

* [x] attempts to cache dependencies installed with composer between builds (faster builds)
* [ ] authenticates with the GitHub API (requires work by you, @bighappyface)

Follows https://github.com/justinrainbow/json-schema/pull/222#issuecomment-174639086.

:information_desk_person: This should cut down the build time significantly; after all, more than 95% of the build time are spent on cloning the dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: false
 language: php
 
+env:
+  global:
+    - secure: "foo"
+
 matrix:
   fast_finish: true
   include:
@@ -19,6 +23,7 @@ cache:
 
 before_install:
   - composer selfupdate
+  - composer config github-oauth.github.com $GITHUB_TOKEN
 
 install:
   - travis_retry composer install --no-interaction --prefer-dist
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,15 @@ matrix:
   allow_failures:
     - php: 7
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,10 @@
 sudo: false
 language: php
 
+env:
+  global:
+    - secure: "foo"
```
localheinz: @bighappyface 

You actually need to create GitHub token and ideally encrypt it, then replace the result here:

```
$ travis encrypt GITHUB_TOKEN=foo

Please add the following to your .travis.yml file:

  secure: "..."

Pro Tip: You can add it automatically by running with --add.
```

:information_desk_person: Do not use `--add`, it just messes up `.travis.yml` by rewriting it.

localheinz: @bighappyface 

Alternatively, you can - of course - specify a `GITHUB_TOKEN` in Travis!

bighappyface:
@localheinz I am no the fence about this one.

First, I am a collaborator but not the owner of the repo, so I can't configure repo settings. Push/pull only.

Second, I am wondering if we could achieve essentially the same (or better) performance improvement by bringing `composer.lock` back into the repo. I know on many other projects that I work with having `composer.lock` makes the biggest difference that I can feel.

localheinz:
@bighappyface 

Both make a lot of sense!

bighappyface:
Closing for now

PR #227, opened 2016-02-02
Fix: Generate coverage on one PHP version only

localheinz:
This PR

* [x] generates coverage on PHP5.6 only and disables xdebug otherwise

Follows https://github.com/justinrainbow/json-schema/pull/222#issuecomment-174639086.

:person_with_pouting_face: This should further cut down build time. Of course, this only makes sense when there aren't different execution paths on different PHP versions.


bighappyface:
+1 pretty minimal

PR #231, opened 2016-02-05
fix(composer): reduce minimum php to 5.3.3

bighappyface:
Reduce minimum php to 5.3.3 in `composer.json` for more inclusive PHP 5.3.x support.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         }
     }],
     "require": {
-        "php": ">=5.3.29"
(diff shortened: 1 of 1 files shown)


Seldaek:
:+1: - I think this is fine really, obviously people shouldn't use older versions than the latest patch, but with the reality of the world, and LTS linux distros, a lot of people still run 5.3.9 for example even though they run custom distro versions that contain most security fixes, so not too bad.

PR #233, opened 2016-02-15
Fixing issue #215 and restoring proposed PR #224.

rtucek:
Sry, I accidentally deleted the original PR #224 during rebase and I couldn't restore it anymore.

As you suggested I've submitted your requested changes to #232 but this was very tricky and the solution is incomprehensible from my point of view.
So even if the min-/maxProperties check is misplaced in the UndefinedConstraint, I'would still prefer my original PR.

Anyhow, both PRs are fixing #215 and it's up to you which PR you prefer.


bighappyface:
Closed with #232

PR #234, opened 2016-02-18
Fix: allow explicit 0 secfracs in datetime format

ehimen:


src/JsonSchema/Constraints/FormatConstraint.php
```diff
@@ -138,7 +138,7 @@ protected function validateDateTime($datetime, $format)
         // which will fail the above string comparison because the passed
         // $datetime may be '2000-05-01T12:12:12.123Z' but format() will return
         // '2000-05-01T12:12:12.123000Z'
-        if ((strpos('u', $format) !== -1) && (intval($dt->format('u')) > 0)) {
+        if ((strpos('u', $format) !== -1) && (preg_match('/\.\d+Z/', $datetime))) {
```
bighappyface: This pattern will produce a match on the invalid test case yet the invalid test case passes because of the absence of the "u" check in this condition.

I have a refiddle where I added the end-of-string anchor that seems to make it a bit more clear: http://refiddle.com/3nf2

bighappyface:
@mathroc what do you think?

src/JsonSchema/Constraints/FormatConstraint.php
ehimen: Ah yes, thanks. Also, I hadn't considered `2000-05-01T12:12:12.000000+01:00`, which this will fail on.

ehimen: I have updated with your suggestion.

mathroc:
that's working for me :)
:+1:

bighappyface:
+1

erayd:
I'm trying to add unit tests to cover this (#366), as the return line is never being hit by the current testsuite. However, I have a few questions about this code:

 1) What is the purpose of the `strpos` conditional? As written, this will *always* return true. Unless this is intended to work around some kind of bug in the PHP core, `strpos` should always return an integer offset >= 0, or boolean `false`.
 2) The case cited in the comment is already in the testsuite, but never actually hits this - the relevant test case seems to be handled by #238, which was merged several months prior to this PR. I assume that means there's a reason for this that isn't covered by the rfc3339 helper thing, but I can't figure out what that might be.
 3) This code appears to be intended for checking `date-time` constraints, but that constraint is entirely handled by the rfc3339 helper. Does this code perhaps need to be moved and integrated with that?

I would appreciate it if someone could clarify what this code is actually for, and whether it still serves a purpose, needs moving, has been superseded, or was simply merged in error after the problem was already solved.

```PHP
        // handles the case where a non-6 digit microsecond datetime is passed
        // which will fail the above string comparison because the passed
        // $datetime may be '2000-05-01T12:12:12.123Z' but format() will return
        // '2000-05-01T12:12:12.123000Z'
        if ((strpos('u', $format) !== -1) && (preg_match('/\.\d+Z$/', $datetime))) {
            return true;
        }
```

PR #236, opened 2016-02-25
Fixup resolve URI from anchored sourceURI

bburnichon:
When using resolver with a base URI which has an anchored, the resolved URI keep the base URI anchor.

This patch ensures frament part of baseURI is removed before generating URI.

```php
use JsonSchema\Uri\UriResolver;

$resolver = new UriResolver();

echo $resolver->resolve('baz.json', 'http://example.org/foo/bar.json#baz');

// expected: http://example.org/foo/baz.json
// actual:   http://example.org/foo/baz.json#baz
```


bburnichon:
ping @justinrainbow @bighappyface

bighappyface:

PR #237, opened 2016-02-27
Aligning EnumConstraint error output to use json_encode()

tinyroy:
json_encode() is also used in FormatConstraint.

Example output
`
["message"]=>string(70) "Does not have a value in the enumeration ["Abacate","Manga","Pitanga"]"
`


bighappyface:
+1

PR #238, opened 2016-03-03
add rfc3339 helper class

bburnichon:
Using this helper class, it make it easy for clients of library to retrieve a proper instance of `DateTime` from their code.

```php
$element = '2016-03-03T13:40:42Z';

$constraint = new JsonSchema\Constraints\FormatConstraint();
$constraint->check($element, (object)array('format' => 'date-time'));

if (!$constraint->isValid()) {
  throw new RuntimeException('Error in element format');
}

$dateTime = JsonSchema\Rfc3339::createFromString($element);
```

If you do not want to fully check constraint, `JsonSchema\Rfc3339::createFromString()` will return `null` for invalid strings

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -8,6 +8,7 @@
  */
 
 namespace JsonSchema\Constraints;
+use JsonSchema\Rfc3339;
 
 /**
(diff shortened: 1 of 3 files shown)


bburnichon:
ping @justinrainbow @bighappyface

bighappyface:
+1

PR #242, opened 2016-03-20
Fix UriResolver::generate

zeggel:
Add '?' sign to query part.
Change array_key_exists to empty to avoid insertion '?' and '#' signs when corresponding parts are empty.


bighappyface:
@zeggel would you mind squashing this down to a single commit?

bighappyface:
Closed due to inactivity. Can be re-opened if necessary.

PR #243, opened 2016-03-20
Implement $ref compliance

TerjeBr:
I have added some tests to #210 
These tests shows that there are still work do be done to make this code work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,16 +8,18 @@ matrix:
     - php: 5.4
     - php: 5.5
     - php: 5.6
+      env: WITH_COVERAGE=true
     - php: 7
     - php: hhvm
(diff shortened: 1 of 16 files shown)


bighappyface:
@TerjeBr you would want to open a PR against @araines fork and branch to expand on his work

TerjeBr:
This was forked off @araines branch.

Do you mean I should open the PR at https://github.com/araines/json-schema instead of here at https://github.com/justinrainbow/json-schema ?

bighappyface:
Correct. That will allow your work to come in on the open PR. Otherwise, your open PR would be bringing in his work and I think he is still making adjustments. I would prefer you work together on that branch of changes. Make sense?

Issue #240, opened 2016-03-16
Fix RefResolver and make it compatible with draft-04


jojo1981:
The current implementation of the RefResolver can only handle JSON schema definitions which result in a tree and not a graph and only to a certain depth.

The RefResolver needs to be replaced or fixed so it can handle JSON schema definitions which contains cycles.

I'm still working on a solution for the RefResolver and I'm almost done. When I've added some regression tests I will submit my code and create a pull request.

mfn:
This would be much appreciated. We tried to add swagger json schema validation do our pipeline in PHP using this library but ultimately failed because of the draft-04 requirement.

jojo1981:
OK, then I will try to speed things up.

jojo1981:
Trying to get it also working with older PHP versions (I'm using PHP 5.6.14) and make sure the travis build: https://travis-ci.org/jojo1981/json-schema will be OK before creating a PR. So for now it's still work in progress.

jojo1981:
@mfn please join the review.

bighappyface:
@jojo1981 please go ahead and open a PR for your work. I basically agree that the public can decide.

bighappyface:
Derp... nvm: https://github.com/justinrainbow/json-schema/pull/245

mfn:
> @mfn please join the review.

@jojo1981 I tried but embarrassingly I actually don't  know how to do it. Maybe I'm wrong here, but I've this setup:
- downloaded json draft 04 schema
- downloaded swagger schema
- and then I've my own `swagger.json`

I know how to valdiate a single json file against a single schema, but I'm stuck how to actually use the API to have them use both. The swagger schema is built on top of the json schema (referencing things like `positiveInteger`, etc.).

Here's a sample code I'm using for "one schema, one json" to validate (in that case it's about the "jsonapi schema" which works fine). Sorry to ask for such support here but I feel like I'm missing something ...

```PHP
<?php
$data = new stdClass(); // assume some actual data I want to validate

$schemaFile = base_path('docs/jsonapi/jsonapi-schema-1.0.json');
$schema = json_decode(file_get_contents($schemaFile));

$refResolver = new RefResolver();
# Minimum depth required to work with jsonapi schema
$refResolver::$maxDepth = 11;
$refResolver->resolve($schema);

$validator = new Validator();
$validator->check($data, $schema);

if (!$validator->isValid()) {
    $msg = "Invalid jsonapi reponse:\n";
[...]

PR #245, opened 2016-03-23
#240 Fix RefResolver and make it compatible with draft-04

jojo1981:
The public API of the RefResolver will be changed. Suggesting to take this PR in a new Major version of this library.

### DIFF
--- a/src/JsonSchema/RefResolver.php
+++ b/src/JsonSchema/RefResolver.php
@@ -9,269 +9,177 @@
 
 namespace JsonSchema;
 
-use JsonSchema\Exception\JsonDecodingException;
-use JsonSchema\Uri\Retrievers\UriRetrieverInterface;
-use JsonSchema\Uri\UriRetriever;
+use JsonSchema\Exception\UnresolvableJsonPointerException;
+use JsonSchema\Iterator\ObjectIterator;
+use JsonSchema\Entity\JsonPointer;
 
 /**
- * Take in an object that's a JSON schema and take care of all $ref references
+ * Take in a source uri to locate a JSON schema and retrieve it and take care of all $ref references.
+ * Try to update the resolved schema which looks like a tree, but can be a graph. (so cyclic schema's are allowed).
+ * This way the current validator does not need to be changed and can work as well with the updated schema.
  *
- * @author Tyler Akins <fidian@rumkin.com>
- * @see    README.md
+ * @package JsonSchema
+ * @author Joost Nijhuis <jnijhuis81@gmail.com>
+ * @author Rik Jansen <rikjansen@gmail.com>
  */
 class RefResolver
 {
-    /**
-     * HACK to prevent too many recursive expansions.
-     * Happens e.g. when you want to validate a schema against the schema
-     * definition.
-     *
-     * @var integer
-     */
-    protected static $depth = 0;
+    /** @var UriRetrieverInterface */
+    private $uriRetriever;
+
+    /** @var UriResolverInterface */
+    private $uriResolver;
 
     /**
-     * maximum references depth
-     * @var integer
+     * @param UriRetrieverInterface $retriever
+     * @param UriResolverInterface $uriResolver
      */
-    public static $maxDepth = 7;
-
-    /**
-     * @var UriRetrieverInterface
-     */
-    protected $uriRetriever = null;
-
-    /**
-     * @var object
-     */
-    protected $rootSchema = null;
-
-    /**
-     * @param UriRetriever $retriever
-     */
-    public function __construct($retriever = null)
+    public function __construct(UriRetrieverInterface $retriever, UriResolverInterface $uriResolver)
     {
         $this->uriRetriever = $retriever;
+        $this->uriResolver = $uriResolver;
     }
 
     /**
-     * Retrieves a given schema given a ref and a source URI
+     * Resolves all schema and all $ref references for the give $sourceUri. Recurse through the object to resolve
+     * references of any child schemas and return the schema.
      *
-     * @param  string $ref       Reference from schema
-     * @param  string $sourceUri URI where original schema was located
-     * @return object            Schema
-     */
-    public function fetchRef($ref, $sourceUri)
-    {
-        $retriever  = $this->getUriRetriever();
-        $jsonSchema = $retriever->retrieve($ref, $sourceUri);
-        $this->resolve($jsonSchema);
-
-        return $jsonSchema;
-    }
-
-    /**
-     * Return the URI Retriever, defaulting to making a new one if one
-     * was not yet set.
-     *
-     * @return UriRetriever
-     */
-    public function getUriRetriever()
-    {
-        if (is_null($this->uriRetriever)) {
-            $this->setUriRetriever(new UriRetriever);
-        }
-
-        return $this->uriRetriever;
-    }
-
-    /**
-     * Resolves all $ref references for a given schema.  Recurses through
-     * the object to resolve references of any child schemas.
-     *
-     * The 'format' property is omitted because it isn't required for
-     * validation.  Theoretically, this class could be extended to look
-     * for URIs in formats: "These custom formats MAY be expressed as
-     * an URI, and this URI MAY reference a schema of that format."
-     *
-     * The 'id' property is not filled in, but that could be made to happen.
-     *
-     * @param object $schema    JSON Schema to flesh out
      * @param string $sourceUri URI where this schema was located
+     * @return object
      */
-    public function resolve($schema, $sourceUri = null)
+    public function resolve($sourceUri)
     {
-        if (self::$depth > self::$maxDepth) {
-            self::$depth = 0;
-            throw new JsonDecodingException(JSON_ERROR_DEPTH);
-        }
-        ++self::$depth;
-
-        if (! is_object($schema)) {
-            --self::$depth;
-            return;
-        }
-
-        if (null === $sourceUri && ! empty($schema->id)) {
-            $sourceUri = $schema->id;
-        }
-
-        if (null === $this->rootSchema) {
-            $this->rootSchema = $schema;
-        }
-
-        // Resolve $ref first
-        $this->resolveRef($schema, $sourceUri);
-
-        // These properties are just schemas
-        // eg.  items can be a schema or an array of schemas
-        foreach (array('additionalItems', 'additionalProperties', 'extends', 'items') as $propertyName) {
-            $this->resolveProperty($schema, $propertyName, $sourceUri);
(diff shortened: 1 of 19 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ coverage
 .settings
 composer.lock
 docs-api
+.idea
```
bighappyface: You should add this to a global ignore: http://stackoverflow.com/questions/7335420/global-git-ignore

bighappyface:
@araines what do you think of this alternative approach to resolving the ref issue (pun intended)?

.gitignore
jojo1981: I agree. I will update the code and revert this change!

jojo1981:
@bighappyface last weekend a good friend of mine also a great programmer was at my home. We've both have studied again on my solution and have improved the code. I have squashed the commits and did a new push. Can you please re-check this code?

bighappyface:
@jojo1981 sorry I didn't chime in on this sooner.

There is a set of ref test provided by the JSON Schema test suite that are currently skipped because of the missing functionality.

You can enable those tests by omitting this line:

https://github.com/justinrainbow/json-schema/blob/61a95a151aac86eaa9bbd31311d3d12a78c971b8/tests/JsonSchema/Tests/Drafts/Draft4Test.php#L21

I pulled this PR down and ran the test suite with those ref tests enabled and ran into a lot of failures. I would expect those tests to pass first and foremost to improve our draft 4 coverage.

araines:
@bighappyface I'm happy to review this once its working with the tests, also I'm not precious about which solution gets used moving forward - so long as the library is improving, thats all that matters!

jojo1981:
@bighappyface Good point I will try to fix the code so the draft 4 tests will also work.

jojo1981:
@bighappyface You're right, those tests also needed to be executed and of course pass successfully. I will fix it and push it again.

jojo1981:
@all Updated the code and enabled test.

jojo1981:
@bighappyface I have ommitted more than only this line: https://github.com/justinrainbow/json-schema/blob/61a95a151aac86eaa9bbd31311d3d12a78c971b8/tests/JsonSchema/Tests/Drafts/Draft4Test.php#L21

jojo1981:
Rebased from `master` branch.

jojo1981:
@araines I agree! Same for me I don't care which solution is moved forward as long as the library improves.

araines:
Picking it up and taking a look :)

araines:
@jojo1981 - Interesting how we independently have approached the problem in a very similar manner - I guess that is probably a positive thing!  However, I think this solution has the edge - it feels a bit cleaner and neater than the one I did in #210 .  The only downside to this one is that we are changing the documented way of using the library, which means it will break for more people (rather than those who had custom-extended it).

However, having looked over this and weighed it up, I think this represents the better solution, even though it will cause slightly more pain for people using this library elsewhere.  @bighappyface I'd personally recommend taking this PR over #210

bighappyface:
@mirfilip would you mind taking a look at this?

jojo1981:
@araines I agree it's not backwards compatible and can only be used in the next major (I think this is a good step for this library). Thanks for your positive remarks! For now I will leave this PR and do no forced pushed anymore. I'm waiting for review comment. So people let's start reviewing!

src/JsonSchema/RefResolver.php
```diff
@@ -9,269 +9,175 @@
...
+    private function getRefSchema(JsonPointer $jsonPointer, $schema)
     {
-        if (! isset($schema->$propertyName) || ! is_object($schema->$propertyName)) {
-            return;
+        /** @var object|array $refSchema */
+        $refSchema = $schema;
+        foreach ($jsonPointer->getPropertyPaths() as $path) {
+            if (is_object($refSchema) && property_exists($refSchema, $path)) {
+                $refSchema = $refSchema->{$path};
+            } elseif (is_array($refSchema) && array_key_exists($path, $refSchema)) {
```
araines: How does this work with getting the last array element (a tilde)?  Spec here: https://tools.ietf.org/html/rfc6901#section-4

I couldn't see a test for it, and I'm trying to wrap my head around how this bit works.

araines:
@jojo1981 given the Interfaces are now defined and are nice and simple, does that mean the other public methods of UriResolver and UriRetriever should potentially be made protected/private - or perhaps the interfaces should be extended to include these if they are necessary?  If my memory serves me correctly, there is a fair bit of duplication between the UriResolver and UriRetriever which we can perhaps slim out whilst you are making public interface breaking changes anyway?

src/JsonSchema/RefResolver.php
jojo1981: The JsonSchema\Entity\JsonPointer should implement the logic as describe in: https://tools.ietf.org/html/rfc6901#section-4

jojo1981:
@araines I agree, but suggest doing that in a separate issue. This PR is unintentional getting bigger than I at first expected. Because this is a library it's always a good idea to use simple and clean interfaces. Only have some public methods who has a clean purpose. Maybe after the complete refactoring we should use annotations on public methods like: `@api` when it's a stable public method and `@deprecated` for methods which will removed in the next major.

src/JsonSchema/RefResolver.php
araines: I see that is the intention, but as I understand it the `JsonPointer` only splits the path elements, meaning you'd end up with a list of the various parts.  That is fine, and is used here, but it looks like at that point the `$refSchema` would be an array and the code highlighted would be executed, looking for the path of `~` in the reference schema - I don't understand how that would ensure we end up with the last element of the array (I may well be missing something).

Either way, I would suggest adding a test for this particular test case to ensure it works and in future the functionality is not broken when further work is done.

araines:
@jojo1981 - fair enough, i'm sure it can be handled as a separate matter.

src/JsonSchema/RefResolver.php
jojo1981: I agree I have to re-check this part. I have re-runned all currently existing PHP Units tests and see this part is not fully covered (lines 108, 109 and 110 are not hit by any test).

Also I think It's better to throw an exception when `$refSchema` can not be fully resolved. Now the current `$schema` will be returned (see line 110).

screen shot 2016-04-06 at 21 12 23

Also I will try to provide a test which coverage line 108.

jojo1981: After some more investigation is see line 108 is dead code. So I will fix this.

jojo1981:
Updated the code but with an extra commit. When this review is finished and can be merged I will squash all the commits and force push again. For now while active in the review process I think is handier to have separate commits.

jojo1981:
@mirfilip would you mind also taking a look at this, just as @bighappyface asked?

jojo1981:
@bighappyface did you re-check this PR?

bighappyface:
@jojo1981 sorry, I have been monitoring the thread but haven't jumped into the full review quite yet. I'll jump on it this weekend for sure.

jojo1981:
@bighappyface No problemo!

src/JsonSchema/RefResolver.php
```diff
@@ -9,269 +9,177 @@
...
      */
-    public function setUriRetriever(UriRetriever $retriever)
-    {
-        $this->uriRetriever = $retriever;
-
-        return $this;
-    }
-
-    protected function resolveRefSegment($data, $pathParts)
+    private function getReservedKeysWhichAreInFactSubSchemas()
```
bighappyface: Interesting name here. Very matter-of-fact. I don't have a problem with it, just commenting on the choice of words/phrasing.

tests/JsonSchema/Tests/Constraints/BaseTestCase.php
```diff
@@ -10,24 +10,39 @@
...
-        $refResolver = new RefResolver(new UriRetriever);
-        $refResolver->resolve($schema);
+        $value = json_decode($input);
 
         $validator = new Validator($checkMode);
+        $validator->check($value, $schema);
 
-        $validator->check(json_decode($input), $schema);
+        if ($validator->isValid()) {
+            $a = 2;
```
bighappyface: Not sure what is going on here.

tests/JsonSchema/Tests/Drafts/Draft3Test.php
```diff
@@ -12,11 +25,12 @@ protected function getFilePaths()
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
     protected function getSkippedTests()
     {
         return array(
-            'ref.json',
-            'refRemote.json',
```
bighappyface: BOOM! :boom:

tests/JsonSchema/Tests/Drafts/Draft4Test.php
```diff
@@ -12,18 +25,15 @@ protected function getFilePaths()
...
+     * {@inheritdoc}
+     */
     protected function getSkippedTests()
     {
         return array(
-            // Not Yet Implemented
-            'definitions.json',
-            // Partially Implemented
-            'ref.json',
-            'refRemote.json',
```
bighappyface: BOOM! :boom: 

Really stoked about this expansion of features for this package.

bighappyface:
@jojo1981 @araines this looks pretty solid to me. I'll admit that I didn't get too deep into the logic of these changes, but I tend to trust developers, particularly when they write tests to support their claims :)

As far as I am concerned this is good to go and merits a major bump. Do you concur?

tests/JsonSchema/Tests/Constraints/BaseTestCase.php
jojo1981: Ugly... just some stupid debug code. I will remove this.

src/JsonSchema/RefResolver.php
jojo1981: :-) Yeah I know. A better name is welcome. So if you have a suggestion?

jojo1981:
Processed the remarks and added an extra commit.

jojo1981:
@bighappyface When we good to go just let me know so I can squash the commits.

bighappyface:
@jojo1981 go ahead and squash the commits. I'm slammed today but I will work on the merge and bump this evening.

jojo1981:
@bighappyface I've squashed the commits as you can see.

bighappyface:
+1

bighappyface:
@jojo1981 @araines we are at 2.0.0:

https://github.com/justinrainbow/json-schema/releases/tag/2.0.0

jojo1981:
@bighappyface Thx!

Issue #241, opened 2016-03-16
Switch to a standard open source license


DicsyDel:
Hi Guys,

Would it be possible to re-license this lib under GPL or Apache or any other standard open source license?

Thanks,
Igor

bighappyface:
GPLv3 works for me. Anyone else have any feedback before I update the license?

charmeleon:
@bighappyface How about the MIT license? According to [this SO thread](http://stackoverflow.com/questions/3902754/mit-vs-gpl-license), GPL license taints the entire project and is incompatible with MIT. Some quotes:

> You can add GPL code into a MIT product, but the resulting application can only be distributed under the GPL licence. 
>  A project that was originally MIT-licensed can no longer be distributed as a whole under the MIT license once it contains code that is only available under GPL. 
> While you can combine GPL and MIT code, the GPL is tainting. Which means the package as a whole gets the limitations of the GPL. As that is more restrictive you can no longer use it in commercial (or rather closed source) software

Personally, I never have reservations about MIT licensed code, but do shy away from GPL licensed projects.

bighappyface:
@charmeleon MIT works for me too and I am open to the most permissive license.

DicsyDel:
MIT licence would be great.

bighappyface:
#246 has the license update and is ready to rock. If I can get a couple "+1" I'll merge and we can close these issues.

Thanks folks!

DicsyDel:
Thanks a lot for this change!

legoktm:
I only saw this now due to the release, so sorry for chiming in late, but can someone explain how the 3-clause BSD license is not a "standard open source license"?

@charmeleon calling GPL "tainting" is just FUD, and statements like "As that is more restrictive you can no longer use it in commercial (or rather closed source) software" are untrue.

PR #246, opened 2016-03-24
chore(LICENSE): switch to MIT

bighappyface:
See #241

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,29 +1,21 @@
-Copyright (c) 2008, Gradua Networks
-Author: Bruno Prieto Reis
-All rights reserved.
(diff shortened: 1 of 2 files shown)


rtucek:
Yeah, like most composer packages, this should be MIT.
Please update the composer.json too!

bighappyface:
@rtucek good catch! Updated.

jojo1981:
@bighappyface I think this one can be merged? Nobody has made any objection...

bighappyface:
@jojo1981 thanks for the review

jojo1981:
Nice, thx!

PR #255, opened 2016-04-19
Fixing issue with non-object values

### LINKED ISSUE #254: Issue with Required Validation
In `UndefinedConstraint::validatedCommonProperties()` if `$value` is not an object, it will assume that the value is valid.  I've run into multiple scenarios where the value is erroneously an array and the schema appears to be valid, even though it actually isn't.

natebrunette:
If the value is not an object when checked by the UndefinedConstraint, it will
fail to check any required properties, making the value appear to be valid when
it isn't.

This fixes #254 

Signed-off-by: Nate Brunette <n@tebru.net>

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -123,6 +123,11 @@ class UndefinedConstraint extends Constraint
                     $this->addError($path, "Is missing and it is required", 'required');
                 }
             }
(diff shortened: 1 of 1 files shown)


natebrunette:
As mentioned in the ticket, this does not work if the parent field is not required.  It still marks each child field as required.  Unfortunately, I was not familiar enough with how the library processed the json to fix it myself.

jojo1981:
@natebrunette I think this PR can be closed?

jojo1981:
@natebrunette Thanks!

PR #262, opened 2016-05-04
Do not override error handler

dmitry-varennikov-eventbase:
### What does this PR Do?
Make use of [error control operator](http://php.net/manual/en/language.operators.errorcontrol.php) to suppress `E_WARNING` for `file_get_contents` if JSON schema not found. 

### Where should the reviewer start?
Check `JsonSchema\Uri\Retrievers\FileGetContents::retrieve` method and run unit test `vendor/bin/phpunit tests/JsonSchema/Tests/Uri/Retrievers/FileGetContentsTest.php`

### Any background context you want to provide?
It's a nice way to suppress php errors using `set_error_handler` function and make use of exceptions instead but it comes with one architectural flaw: exceptions thrown from overriden error handler bypass application stack trace. 
In my code I use a middleware which takes care of thrown exceptions on the top of my application stack and outputs nice response to a client. In case of overriding error handler (and even restoring it back right away) the error thrown from it bypasses my error handler thus resulting in `PHP Fatal error:  Uncaught exception 'JsonSchema\\Exception\\ResourceNotFoundException' with message 'JSON schema not found at file:...`
Use of error control operator allows to preserve functionality intact and gives a caller code ability to catch this exception within application stack. 

**UPD:** I know I can wrap a call to resolver in `try-catch` but I expect a more predictable behaviour. 
```
try {
    $schema = $refResolver->resolve('file://'.$events_create_schema);
} catch (ResourceNotFoundException $e) {
    // @TODO: take care of exception here
}
```

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/FileGetContents.php
+++ b/src/JsonSchema/Uri/Retrievers/FileGetContents.php
@@ -14,34 +14,29 @@ use JsonSchema\Validator;
 
 /**
  * Tries to retrieve JSON schemas from a URI using file_get_contents()
- * 
- * @author Sander Coolen <sander@jibber.nl> 
+ *
+ * @author Sander Coolen <sander@jibber.nl>
  */
 class FileGetContents extends AbstractRetriever
 {
     protected $messageBody;
-    
+
     /**
      * {@inheritDoc}
      * @see \JsonSchema\Uri\Retrievers\UriRetrieverInterface::retrieve()
      */
     public function retrieve($uri)
     {
-        $context = stream_context_create(array(
-            'http' => array(
-                'method' => 'GET',
-                'header' => "Accept: " . Validator::SCHEMA_MEDIA_TYPE
-            )));
-
-        set_error_handler(function() use ($uri) {
-            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
-        });
-        $response = file_get_contents($uri);
-        restore_error_handler();
-
-        if (false === $response) {
+        if (!file_exists($uri)) {
             throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
         }
+
+        $response = file_get_contents($uri);
+
+        if (false === $response) {
+            throw new ResourceNotFoundException('JSON schema was not retrieved at ' . $uri);
+        }
+
         if ($response == ''
             && substr($uri, 0, 7) == 'file://' && substr($uri, -1) == '/'
         ) {
@@ -55,10 +50,10 @@ class FileGetContents extends AbstractRetriever
             // Could be a "file://" url or something else - fake up the response
             $this->contentType = null;
         }
-        
+
         return $this->messageBody;
     }
-    
+
     /**
      * @param array $headers HTTP Response Headers
      * @return boolean Whether the Content-Type header was found or not
@@ -70,10 +65,10 @@ class FileGetContents extends AbstractRetriever
                 return true;
             }
         }
-        
+
         return false;
     }
-    
+
     /**
      * @param string $header
      * @return string|null


jojo1981:
@dmitry-varennikov-eventbase Suppressing errors is not a good thing to do. A better solution would be to check a certain condition and throw exception an exception.

src/JsonSchema/Uri/Retrievers/FileGetContents.php
```diff
@@ -33,11 +33,7 @@ public function retrieve($uri)
                 'header' => "Accept: " . Validator::SCHEMA_MEDIA_TYPE
             )));
 
-        set_error_handler(function() use ($uri) {
-            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
-        });
-        $response = file_get_contents($uri);
-        restore_error_handler();
+        $response = @file_get_contents($uri);
```
jojo1981: Why not:

        if (!file_exists($uri)) {
            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
        }

        $response = file_get_contents($uri);

dmitry-varennikov-eventbase:
That also will do

mirfilip:
Agree with @jojo1981. Nevertheless, the problem is that `\JsonSchema\UriRetrieverInterface` doesn't state what every particular resolver implementation should do when it cannot retrieve. The behaviour should absolutely be the same for every retriever (or one would have to adjust error handling if a retriever is changed).

@dmitry-varennikov-eventbase I don't want to derail your PR, so I suggest you add `@throws \ResourceNotFoundException` to the interface and implement the behaviour to just `FileGetContents`. That'd give us the solid base for further refinements.

dmitry-varennikov-eventbase:
I took a second look at `JsonSchema\Uri\Retrievers\FileGetContents::retrieve` method. There is unused stream `$context`. Just curious if it has to be used at all.

mirfilip:
@dmitry-varennikov-eventbase It looks like I has been prepared to be passed to `file_get_contents` as a stream. The question is whether `FileGetContents` should work with external URIs - there is `Curl` for that. It's important because `Validator::SCHEMA_MEDIA_TYPE` needs to be honoured or not. Also, `$context` is bound to `HTTP` protocol. @jojo1981 any ideas?

jojo1981:
@mirfilip I think `$context` can be removed. It's not used and it has no purpose in the `JsonSchema\Uri\Retrievers\FileGetContents` class.

I have no idea why there are 3 classes which implement the `JsonSchema\Uri\Retrievers\UriRetrieverInterface`. By default always the `JsonSchema\Uri\Retrievers\FileGetContents` will be used and this one will do it's job for local and remote files. See: https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Uri/UriRetriever.php#L77

dmitry-varennikov-eventbase:
I removed stream context and used `file_exists` suggested by @jojo1981

jojo1981:
@bighappyface I think this one can be merged.

bighappyface:
@dmitry-varennikov-eventbase would you mind squashing this down into a single commit?

dmitry-varennikov-eventbase:
Done

bighappyface:
+1

jojo1981:
+1

mirfilip:
@jojo1981 It wasn't clear if `FileGetContents` is for local files or remote ones. This `$context` looked like it was meant for both. I think we should keep `FileGetContents` for local files only, `Curl` for remote files.

Maks3w:
Another BC Break PR merged.  Another version semantic violation.

This should be released as v3.0.0 not as v2.0.3. Please revert

bighappyface:
@Maks3w please explain why this refactor is a major release. Accusers carry the burden of proof and you aren't presenting anything.

Maks3w:
@bighappyface What is the purpose of FileGetContents? What are their his responsabilities?

What kind of resources is able to manage `file_exists` and what kind of resources manage `file_get_contents`.


The magic word is HTTP

mirfilip:
@bighappyface Unfortunately, indeed there is a BC break. [Early file_exists check](https://github.com/justinrainbow/json-schema/commit/780dd6f45f86eadfcebee13d22a3771391bbb2e1#diff-3bcbe093a26c682f3edc77bb6ff3060fR30). It went unnoticed because majority of people use it in a local file manner.

@Maks3w Complaining cause you had patch up mistakes of others people *is* understandable, going back to the maintainers and bitching about is not. You contributed to ZF2 so you've probably seen it break BC couple of times. The point is, no one did it on purpose. If @bighappyface had noticed the change makes `FileGetContents` useless HTTP-wise, he'd have probably had it reworked.

Anyway, back to technical stuff, have you seen [FileGetContentsTest](https://github.com/justinrainbow/json-schema/blob/2.0.3/tests/JsonSchema/Tests/Uri/Retrievers/FileGetContentsTest.php)? You cannot tell what the responsibility of this class is, the tests are crap. This has been the state of matters when you started to use this project. To be honest, there is a [CurlRetriever](https://github.com/justinrainbow/json-schema/blob/2.0.2/src/JsonSchema/Uri/Retrievers/Curl.php) that I'd argue you should've used in the first place. The fact `FileGetContents` used to handle external links could've been as well considered a bug and removed. Fixing such bugs wouldn't have been considered a BC. Seen that everywhere.

Just to fresh the air, I have started a major rework in Retrievers department, it will clear up such things.

mirfilip:
@bighappyface I'd be awesome to have the project tested in such a way we can blindly rely on whether tests pass to accept the PR. The fact is tests are really bad here. I see no other option but to be extra careful and actually push the PR creator to cover the missing ground. E.g. in this case, we all should stop and document `FileGetContents` responsibility. I know, it hard to ask external people to dig but ¯\_(ツ)_/¯

@bighappyface I'd rewrite your analogy to: you're working with a gun, it shoots. Some day you discover it also scratches your bag well. An awesome, undocumented feature. You keep doing that with all the models of the pistol. One day you buy a new model and it's all round, sucks at scratching. The conclusion being: you were using a wrong thing to do the job, but it was undocumented that you cannot use it for scratching. There is no other way but to refer to common sense here.

I've seen similar situations with ZF2 and Symfony 2.x. Almost always the explanations were "this wasn't supposed to be used this way" and the thing wasn't considered a BC break.

PR #264, opened 2016-05-06
Bump json-schema/JSON-Schema-Test-Suite to 1.1.2

mirfilip:
/cc @bighappyface 

No effort move, all tests still passing. I'm currently looking at using `1.2.0` (seen [that](https://github.com/bighappyface/json-schema/pull/1))

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,11 +28,11 @@
         "type": "package",
         "package": {
             "name": "json-schema/JSON-Schema-Test-Suite",
-            "version": "1.1.0",
+            "version": "1.1.2",
             "source": {
                 "url": "https://github.com/json-schema/JSON-Schema-Test-Suite",
                 "type": "git",
-                "reference": "1.1.0"
+                "reference": "1.1.2"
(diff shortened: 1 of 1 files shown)


bighappyface:
@mirfilip please use a [global ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for the `.idea` directory. That is a user-specific ignore as not everyone uses the Jetbrains products.

Otherwise +1

mirfilip:
@bighappyface I cannot use a global ignore as some projects I work with do keep this under version control. Anyway, it was rather a thing for every other contributor that uses Jetbrains stuff so that the IDE config doesn't get committed.

I've checked some popular repos and they indeed refrain from specifying IDE specific stuff. Reverting that then.

bighappyface:
@mirfilip thanks for the update. Last request: please squash this down.

mirfilip:
@bighappyface Sure! Also, please see #265 as it supersedes this PR.

PR #265, opened 2016-05-09
Bump json-schema/JSON-Schema-Test-Suite to 1.2.0 + some tests improvements

mirfilip:
## Things done:
- Simplify tests directory structure. Namespace is still unchanged.

- Change drafts tests that we read from json schema test suite to be an associative array where a key is a `file.json / suite description / test case description`

Failing tests output before the change:
```
1) JsonSchema\Tests\Drafts\Draft4Test::testValidCases with data set #75 ('{"bar":1,"baz":2}', '{"properties":{"foo":{"not":{}}}}')
```
Failing tests output after the change:
```
1) JsonSchema\Tests\Drafts\Draft4Test::testValidCases with data set "not.json / forbidden property / property absent" ('{"bar":1,"baz":2}', '{"properties":{"foo":{"not":{}}}}')
```
It uses [description](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/develop/tests/draft4/additionalItems.json#L3) fields from suite and test case. This may seem an overkill, but it really simplifies hunting for the test to fix when you have a readable path.

- Update `json-schema/JSON-Schema-Test-Suite` to `1.2.0`. **Pay attention here, important**:
If we really want to state which version of the draft the validator follows, we **need** to compare against the newest available version. I presume everyone agrees ;) So, `1.2.0` bring three failure specs:
 - `format.json` / `Draft 3` / optional
 - `format.json` / `Draft 4` / optional
 - `not.json` / `Draft 4` / required

Sad thing, there is no mechanism to skip one test case from `not.json`. For the time being, I've skipped the whole file as it's now my priority to fix that case as it's required by spec. Still, I think it doesn't hurt us as I doubt people are going to dig tests to see the compliance with spec (I'm working on documenting that).

/cc @bighappyface


bighappyface:
+1

I looked into this in the past and stuffed it into my ever-growing to-do list... Thanks @mirfilip for bringing us up to the latest in terms of both the test site and a more sane PSR-4 directory structure.

mirfilip:
@bighappyface Thx for the check. This will inevitably cause some PRs to need of rebase but I figured there are not too many PRs so it's a good time for such change. I'm thinking of doing the same for `src`, what do you think?

bighappyface:
@mirfilip I think squashed PRs are for the best. We need to add a CONTRIBUTING file to help clarify our expectations.

Yes, I say do the same for `src` while you are at it, and thanks ;)

mirfilip:
@bighappyface squashed. This is the final version. I will tackle `src` structure separately.

jojo1981:
+1

bighappyface:
+1

I'm going to let this one sit in master and bump at a later time.

PR #267, opened 2016-05-19
TypeConstraint checks if a value is really a numeric array or it's an object

Alex-PK:
As JSON objects are represented as PHP arrays once decoded, the type check didn't work when importing json or yaml files.
I improved the check for both arrays and objects.

### DIFF
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -144,9 +144,9 @@ class TypeConstraint extends Constraint
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Alex-PK:
I added a check for empty arrays for both array and object check.
Most tests pass, but 2 from schema-v3 don't, as they test empty arrays and objects which, in PHP, are equivalent.

The alternative would be to go back to check only for is_object() and is_array(), but this would force to use stdClass everywhere a json object is needed.

I am open to suggestions as to how to manage this.

bighappyface:
@Alex-PK any movement on this regarding testing?

PR #268, opened 2016-05-19
Check headers for web-downloaded schemas instead of using file_exists

Alex-PK:
file_exists() doesn't work for URLs. This patch checks the headers for http(s) URIs and uses the old file_exists() for anything else.

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/FileGetContents.php
+++ b/src/JsonSchema/Uri/Retrievers/FileGetContents.php
@@ -27,8 +27,15 @@ class FileGetContents extends AbstractRetriever
      */
     public function retrieve($uri)
     {
-        if (!file_exists($uri)) {
-            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
+        if (preg_match('~https?://~i', $uri)) {
+            $headers = @get_headers($uri);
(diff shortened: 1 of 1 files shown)


bighappyface:
@Maks3w @mirfilip thoughts on this as a fix per our discussion in #262?

bighappyface:
@Alex-PK I haven't spent much time thinking about testing this, but test would be nice 😉

Maks3w:
IMO the code present in v2.2 was totally correct

This do two network connections or two local file access when everything is OK however when the resource does not exists only do one.

So just restore the previous version and don't add performance penalties In special the side related to network connection.

mirfilip:
This is not the way to go. First of all, I don't see a reason to do a existence check before you do `file_get_contents`. Especially when:
```php
// HTTP protocol
var_export(@file_get_contents("http://example.org/not-existing-resource"); // false
// FILE protocol
var_export(@file_get_contents("file:///tmp/foo"); // false
```
whatever the protocol, it returns false when silenced. That's all we care about.

Secondly, I agree with @Maks3w here. The best is just to revert #262 until the confusion is cleared up (with refactoring).

bighappyface:
Closed with #273

PR #270, opened 2016-05-23
feat(FileGetContents): remove file_exists check

bighappyface:
This PR removes the `file_exists` check in `JsonSchema\Uri\Retrievers\FileGetContents` per #262

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/FileGetContents.php
+++ b/src/JsonSchema/Uri/Retrievers/FileGetContents.php
@@ -27,10 +27,6 @@ class FileGetContents extends AbstractRetriever
      */
     public function retrieve($uri)
     {
-        if (!file_exists($uri)) {
-            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
-        }
-
         $response = file_get_contents($uri);
 
         if (false === $response) {


bighappyface:
@Maks3w @mirfilip this is a quick PR to remove the `file_exists` check and showcases how we had at least some test coverage despite the tests not being considered that great...

Frankly I didn't like the `set_error_handler` approach but with situations like this it looks more and more like reverting is the best option.

Maks3w:
I understand you don't like and for my use case this PR is fine. However don't throw an exception where previously did still to be considered a BC Break.

Just revert the change and discuss why the error handler is not fine for you. The only issue I know is for nested error handlers.

As far as I read in other issues the retriever subcomponent doesn't like to anyone and a refactor is on the roadmap. Just restore previous code and focus in the refactor instead patch the code you don't like.

bighappyface:
How about we just delete the 2.0.3 tag and leave the current code in dev-master? Wouldn't that solve your issue provided you don't have 2.0.3 set in your `composer.lock` file?

I would prefer to not revert code people took time to contribute and that made it through the review process.

Maks3w:
Three Gold rule: Don't rewrite history, don't remove tags, don't introduce bc breaks.

IIRC the only good the original PR had was some code style and remove the unused context var. Whatever if the PR is wrong, is wrong, there is no much to discuss. 

Do whatever you think its best. For the next time suspect from any change, kill the "Tests would be nice" tag and reject any PR does not follow the contribution guidelines.

bighappyface:
@Maks3w there aren't any contribution guidelines. With such strong opinions you should consider contributing them and/or a fix for this as well. You are the only one complaining after all.

Maks3w:
I'm contributing right now guiding you to become a better reviewer.

bighappyface:
sure...

PR #271, opened 2016-05-23
Revert "Do not override error handler"

bighappyface:
Reverts justinrainbow/json-schema#262

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/FileGetContents.php
+++ b/src/JsonSchema/Uri/Retrievers/FileGetContents.php
@@ -14,29 +14,34 @@ use JsonSchema\Validator;
 
 /**
  * Tries to retrieve JSON schemas from a URI using file_get_contents()
- *
- * @author Sander Coolen <sander@jibber.nl>
+ * 
+ * @author Sander Coolen <sander@jibber.nl> 
  */
 class FileGetContents extends AbstractRetriever
 {
     protected $messageBody;
-
+    
     /**
      * {@inheritDoc}
      * @see \JsonSchema\Uri\Retrievers\UriRetrieverInterface::retrieve()
      */
     public function retrieve($uri)
     {
-        if (!file_exists($uri)) {
-            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
-        }
+        $context = stream_context_create(array(
+            'http' => array(
+                'method' => 'GET',
+                'header' => "Accept: " . Validator::SCHEMA_MEDIA_TYPE
+            )));
 
+        set_error_handler(function() use ($uri) {
+            throw new ResourceNotFoundException('JSON schema not found at ' . $uri);
+        });
         $response = file_get_contents($uri);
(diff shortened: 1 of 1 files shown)


bighappyface:
@Maks3w should I bump to 2.0.4 or is that something else that is wrong to you?

Maks3w:
IIRC semver said in this case a major version should be released. But knowing how Composer work it's better to be released as minor or patch. 

We could consider this like a bug instead of HTTP support drop so release as 2.0.4

dmitry-varennikov-eventbase:
Can anybody point me to the place where `$context` variable used as we reverted to the original code?

dmitry-varennikov-eventbase:
Also would you mind to use my original proposal to use error control operator `@` to suppress errors for `file_get_contents` instead of redefining error handler?

bighappyface:
@dmitry-varennikov-eventbase how about the Curl retriever?

https://github.com/justinrainbow/json-schema/blob/5c133a3e336c2b35a4233f5da4cc15f083ea128d/src/JsonSchema/Uri/Retrievers/Curl.php

dmitry-varennikov-eventbase:
I didn't find any errors mentioning for [curl-exec](http://php.net/manual/en/function.curl-exec.php). So I'm not sure what your question is.

Also I believe `Curl` and `FileGetContents` retrievers should behave the same and both should NOT override php error handler but just throw `ResourceNotFoundException`

Maks3w:
@dmitry-varennikov-eventbase Error supression with `@` its almost not allowed by any project. One option is allow inject a custom error handler.

public $errorHandlersCallables = ['start' => callable, 'end' => callable]

$this->errorHandlersCallables['start']()
....
$this->errorHandlersCallables['end']()

dmitry-varennikov-eventbase:
@Maks3w don't you think it requires too much code for every caller to retrieve a JSON schema file?

Maks3w:
custom error handlers are optional. By default current behavior (set_error_handler) is used.

IMO let this as is and come back when you have an specific real case.

dmitry-varennikov-eventbase:
@Maks3w 
Here's a [specific real case](https://github.com/justinrainbow/json-schema/pull/262) I described in my original PR and suggested a solution to use error control operator `@` instead of overriding error handler. Your answer was 
> Error supression with @ its almost not allowed by any project.

Is it the only argument NOT to use my suggestion?

Maks3w:
May this could work.

```php
$error = false;
set_error_handler(function ($a, $b, ...) use (&$error) { $error = [$a, $b, ...];}
file_get_contents
restore_error_handler
if ($error) {throw Exception($error['a'], $error['b'])}

dmitry-varennikov-eventbase:
Yeah as long as an exception is thrown outside of the re-defined error handler I'm happy with it

dmitry-varennikov-eventbase:
[Throw `ResourceNotFoundException` outside of re-defined error-handler](https://github.com/justinrainbow/json-schema/pull/273)

PR #276, opened 2016-06-03
use JsonPointer for path handling

steffkes:
crafting a new API, trying to follow the jsonapi-specs, i've realized that  [error-objects](http://jsonapi.org/format/upcoming/#error-objects) require a json pointer:

> * `source`: an object containing references to the source of the error, optionally including any of the following members:
>   * `pointer`: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. `/data` for a primary data object, or `/data/attributes/title` for a specific attribute].

before making more changes, first thing to do was to get all tests working again, especially bringing back all the existing property paths since i don't intend to introduce a large BC which i probably still do? :/

having had a look at RFC6901 again, i'm curious why `JsonPointer#getPropertyPathAsString` does include a leading hashmark @jojo1981 - isn't that actually related to the filename rather than the path? checking other existing JsonPointer implementations in PHP, all use path expressions starting with a slash instead.

as a nice side-effect, using dots in keys wouldn't result in path that is rather hard to understand like `tld.company.project[0].content` (instead of `/tld.company.project/0/content`)

@mirfilip @bighappyface WDYT?

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -265,7 +265,7 @@ abstract class Constraint implements ConstraintInterface
     }
 
     /**
-     * Checks format of a element
+     * Checks format of an element
      *
      * @param mixed            $value
      * @param mixed            $schema
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -11,6 +11,7 @@ namespace JsonSchema\Constraints;
 
 use JsonSchema\Uri\UriRetriever;
 use JsonSchema\Validator;
+use JsonSchema\Entity\JsonPointer;
 
 /**
  * The Base Constraints, all Validators should extend this class
@@ -80,10 +81,11 @@ abstract class Constraint implements ConstraintInterface
     /**
      * {@inheritDoc}
      */
-    public function addError($path, $message, $constraint='', array $more=null)
+    public function addError(JsonPointer $path = null, $message, $constraint='', array $more=null)
     {
         $error = array(
-            'property' => $path,
+            'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
+            'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
             'message' => $message,
             'constraint' => $constraint,
         );
@@ -132,37 +134,32 @@ abstract class Constraint implements ConstraintInterface
     /**
      * Bubble down the path
      *
-     * @param string $path Current path
-     * @param mixed  $i    What to append to the path
+     * @param JsonPointer|null $path Current path
+     * @param mixed            $i    What to append to the path
      *
-     * @return string
+     * @return JsonPointer;
      */
-    protected function incrementPath($path, $i)
+    protected function incrementPath(JsonPointer $path = null, $i)
     {
-        if ($path !== '') {
-            if (is_int($i)) {
-                $path .= '[' . $i . ']';
-            } elseif ($i == '') {
-                $path .= '';
-            } else {
-                $path .= '.' . $i;
-            }
-        } else {
-            $path = $i;
-        }
-
+        $path = $path ?: new JsonPointer('');
+        $path = $path->withPropertyPaths(
+            array_merge(
+                $path->getPropertyPaths(),
+                array_filter(array($i), 'strlen')
+            )
+        );
         return $path;
     }
 
     /**
      * Validates an array
      *
-     * @param mixed $value
-     * @param mixed $schema
-     * @param mixed $path
-     * @param mixed $i
+     * @param mixed            $value
+     * @param mixed            $schema
+     * @param JsonPointer|null $path
+     * @param mixed            $i
      */
-    protected function checkArray($value, $schema = null, $path = null, $i = null)
+    protected function checkArray($value, $schema = null, JsonPointer $path = null, $i = null)
     {
         $validator = $this->getFactory()->createInstanceFor('collection');
         $validator->check($value, $schema, $path, $i);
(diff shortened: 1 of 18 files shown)


steffkes:
bummer, first commit and already broke the 5.3 build - didn't check the library requirements beforehand .. sorry for that, already pushed a fix using the old array syntax

jojo1981:
@steffkes According the RFC6901 a **#** evaluate to the the whole document.

tests/Entity/JsonPointerTest.php
```diff
@@ -92,4 +92,16 @@ public function getTestData()
 
         );
     }
+
+    public function testJsonPointerWithPropertyPaths()
+    {
+        $initial = new JsonPointer('#/definitions/date');
+        $modified = $initial->withPropertyPaths(array('~definitions/general', '%custom%'));
```
jojo1981: To prove **withPropertyPaths** is immutable. The code needs to be changed here into:

    $this->assertEquals(array('definitions', 'date'), $initial->getPropertyPaths());
    $this->assertEquals('#/definitions/date', $initial->getPropertyPathAsString());
    $modified = $initial->withPropertyPaths(array('~definitions/general', '%custom%'));
    $this->assertEquals(array('definitions', 'date'), $initial->getPropertyPaths());
    $this->assertEquals('#/definitions/date', $initial->getPropertyPathAsString());

steffkes: @jojo1981 i'm fine adding those two lines - would you mind explaining why you think they are necessary? i can't get my head around, what are those lines asserting that the current test doesn't? it's rather a duplication of the existing tests - which obviously doesn't hurt, but it still doesn't relate to the question if the method returns a new object or modified the existing

if we want to have additional checks, what about [assertNotSame](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertSame) for `$initial` and `$modified` ?

jojo1981: @steffkes It depends on which behavior should be tested. Asserting a new object is returned is 1 thing. asserting this has no side effect is another thing (so the original object has **NOT** be changed). I think both can be tested (so proved).

steffkes: @jojo1981 that still doesn't answer my question for more explanation about your intention. the second assert block for `$initial` already implies that it was okay in the first place - i don't see what kind of errors we could avoid by testing that (again). i'm not opposed to your suggestions, not at all, just want to understand if there is a real improvement that i'm just not seeing right now

steffkes:
i went ahead an included a `pointer` property in all error instances. since the jsonapi-specs said the pointer was supposed to look like `/data/attributes/title` i do call `ltrim` to get rid of the leading hashmark we do have - not entirely sure if this is the right thing to do here? probably have to re-read the specs (RFC as well as jsonapi) again to get a better idea how to handle it.

steffkes:
anyone like to have a second look? @bighappyface, probably @mirfilip or @jojo1981 (again, if you didn't already?)

the newly added `PointerTest` should make it clear how it's supposed to look and work. if there are any other places where we should add it, please let me know.

bighappyface:
@steffkes I'll schedule some time to review this in the coming days. I'm super slammed at the moment but very grateful for your work.

steffkes:
no worries @bighappyface, no need to rush it - thanks for your feedback. let me know if you have questions and/or suggestions, i'll wait for it. spare time is sometimes rare, been there too (:

bighappyface:
@steffkes would you please rebase this on master and squash the commits down to one?

steffkes:
@bighappyface i'm not sure that it went entirely as planned - didn't need rebasing/squashing that often in the past. if it doesn't look like it should - please do let me know. if you can i would appreciate it if you could tell me where i got it wrong.

i expected the other commits to disappear from this PR after doing all the steps i found described on [How-to-Rebase-a-Pull-Request](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request). at least this is what i looks like in my local git history ..

steffkes:
ah, looks like it's working - maybe some sort of cache on github that kept the older commits for the time being. the only commit i can see now is the latest that was created during the whole rebase/squash operation.

bighappyface:
Squash looks good

steffkes:
gentle reminder, no rush here - if you need anything else from me, drop me a note @bighappyface

bighappyface:
@steffkes my apologies for yet another rebase on master but some changes in another PR have put this into conflict. Would you mind rebasing on master once more? You shouldn't have to worry about squashing anymore.

steffkes:
@bighappyface i should have checked that before - thanks for the heads up. did another rebase - the diff looks good again. if you need anything else, let me know

bighappyface:
@steffkes no worries.

Would you mind reviewing the test failures?

steffkes:
argh, should have done that before as well .. will do it now

steffkes:
working again @bighappyface

bighappyface:
@jojo1981 @mirfilip @steffkes @FlorianSW would you mind taking a look at this PR?

I am thinking this would get bundled into a major version release to avoid any BC issues.

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -254,20 +251,28 @@ protected function checkNumber($value, $schema = null, $path = null, $i = null)
...
     {
         $validator = $this->getFactory()->createInstanceFor('enum');
         $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
     }
 
-    protected function checkFormat($value, $schema = null, $path = null, $i = null)
+    /**
+     * Checks format of a element
```
FlorianSW: an element

steffkes:
anyone? anything else? :)

bighappyface:
+1

PR #277, opened 2016-06-04
New way of handling references

psafarov:
I had issues compiling schemas containing circular references with RefResolver, so the way of handling references was changed, now Validator contains SchemaStorage object inside, which contains all schemas and can resolve references when it's needed. The pros of this approach are:
- lazy loading, if some schemas are not used, they won't be loaded
- no infinite loops, thanks to lazy evaluation again
- api becomes kind of simpler
```php
<?php
// You can pass reference which will be handled by SchemaStorage::resolveRef
$validator->check($data, (object)['$ref' => 'file://path/to/your/shcema'])
```


.gitignore
```diff
@@ -7,3 +7,4 @@ coverage
 .settings
 composer.lock
 docs-api
+.idea/
```
bighappyface: Please omit this change and add this to a global `.gitignore` https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

bighappyface:
@psafarov please rebase this on master

src/JsonSchema/SchemaStorage.php
```diff
@@ -27,29 +31,13 @@ public function getUriRetriever()
 	}
 
 	/**
-	 * @param UriRetrieverInterface $uriRetriever
-	 */
-	public function setUriRetriever($uriRetriever)
```
bighappyface: This approach prevents reuse of an instance of `SchemaStorage` by only allowing a user to set the retriever or resolver only during instantiation. The goal of simplifying instantiation `SchemaStorage` is achieved but I am not sure that is a problem nor am I sure it is germane to the main goal of handling circular references.

I am not sure I think this is an improvement but I won't hold up this PR for this if the community supports this move.

psafarov: @bighappyface you right, it's not only about circular references. I also wanted to simplify the interface. Now there is no RefResolvers you need to know about, until you have to

ElectroLutz:
Any info on that one? I would like to see this in a release.

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -33,37 +37,54 @@
...
-    {
-        $this->checkMode    = $checkMode;
-        $this->uriRetriever = $uriRetriever;
-        $this->factory = $factory;
+    public function __construct(
+        $checkMode = self::CHECK_MODE_NORMAL,
+        SchemaStorage $schemaStorage = null,
+        UriRetriever $uriRetriever = null,
+        Factory $factory = null
+    ) {
```
steffkes: wouldn't it be kinda easier to move `$schemaStorage` to the end of the list? if we'd append it to the existing list, we wouldn't introduce a BC-Break at this point.

src/JsonSchema/Constraints/Factory.php
```diff
@@ -41,15 +46,17 @@ class Factory
     );
 
     /**
+     * @param SchemaStorage $schemaStorage
      * @param UriRetriever $uriRetriever
      */
-    public function __construct(UriRetriever $uriRetriever = null)
+    public function __construct(SchemaStorage $schemaStorage = null, UriRetriever $uriRetriever = null)
```
steffkes: Same here, if `$schemaStorage` would go second/last - it would behave like it currently does.

src/JsonSchema/Constraints/Factory.php
```diff
@@ -41,15 +46,17 @@ class Factory
...
     {
         if (!$uriRetriever) {
             $uriRetriever = new UriRetriever();
         }
 
         $this->uriRetriever = $uriRetriever;
+        $this->schemaStorage = $schemaStorage;
```
steffkes: since `$schemaStorage` defaults to `null` - shouldn't we initiate a default `SchemaStorage` instance?

steffkes:
just had a first glance over the changes - did add a comment where i thought it's need. will have a more detailed look during this week - given this PR is still open and not already merged

src/JsonSchema/Constraints/Constraint.php
FlorianSW: +1 This seems like an unnecessary breaking change. However, if this is already a breaking change (and in this core functionality), it probably doesn't make sense to reorder arguments, if it makes sense. Unfortunately, I don't see a good reason to do it here? :)

src/JsonSchema/Constraints/Factory.php
FlorianSW: Yep, the behaviour should be identical to UriRetriever if possible.

src/JsonSchema/Constraints/Factory.php
```diff
@@ -10,15 +10,20 @@
...
 
 /**
  * Factory for centralize constraint initialization.
  */
 class Factory
 {
     /**
+     * @var SchemaStorage
+     */
+    protected $schemaStorage;
```
FlorianSW: We've a getter method for $uriRetriever, wouldn't it make sense to have one for `$schemStorage`, too?

src/JsonSchema/Constraints/ObjectConstraint.php
```diff
@@ -125,7 +125,7 @@ public function validateElement($element, $matches, $objectDefinition = null, $p
     public function validateDefinition($element, $objectDefinition = null, $path = null)
     {
         foreach ($objectDefinition as $i => $value) {
-            $property = $this->getProperty($element, $i, new UndefinedConstraint());
+            $property = $this->getProperty($element, $i, $this->getFactory()->createInstanceFor('undefined'));
```
FlorianSW: Looks good, however, is an unrelated change according to the discussion.

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -48,7 +48,7 @@ public function check($value = null, $schema = null, $path = null, $i = null)
             $validatedOneType = false;
             $errors = array();
             foreach ($type as $tp) {
-                $validator = new static($this->checkMode);
+                $validator = $this->getFactory()->createInstanceFor('type');
```
FlorianSW: I support this. However, `checkMode` can be set in the constructor (and you probably don't need to use the Factory, even if you should). So I'm not sure, if there was a reason why the constructor with the explicit checkMode was used or not (another contributor probably knows more about that).

src/JsonSchema/SchemaStorage.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @return UriResolverInterface
+     */
+    public function getUriResolver()
+    {
+        return $this->uriResolver;
+    }
+
+    public function addSchema($id, $schema = null)
```
FlorianSW: `$id`? I'm not sure, if this is as informative as it could be. The parameter name should indicate what the parameter represents. Actually I would think about a unique identifier, like a number/integer. But if I understand it correctly, it's an Uri, that resolves to a json schema. So probably you want to change the parameter name.

Also: missing phpdoc block.

src/JsonSchema/SchemaStorage.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (is_null($schema)) {
+            $schema = $this->uriRetriever->retrieve($id);
+        }
+        $objectIterator = new ObjectIterator($schema);
+        foreach ($objectIterator as $toResolveSchema) {
+            if (property_exists($toResolveSchema, '$ref') && is_string($toResolveSchema->{'$ref'})) {
```
FlorianSW: Is there a reason why `$ref` has a "$" and isn't called `ref`? While we're already here, we could rename it to `ref` as it currently is no variable and instead is a property. At least I would appreciate it :P

FlorianSW:
Added some comments. For now it looks sane and reasonable, but someone (:P) should test it in a real life example, just to make sure it's working like expected :)

src/JsonSchema/Constraints/Constraint.php
psafarov: I think, you're right, it's better to be in the end

src/JsonSchema/Constraints/ObjectConstraint.php
psafarov: It's a necessary change, the factory passes existing schemaStorage to the new constaint

src/JsonSchema/SchemaStorage.php
psafarov: @FlorianSW 'id' and '$ref' are property names from jsonschema documentation, so I'm not sure we can rename them, not the '$ref' at least

src/JsonSchema/Constraints/Factory.php
psafarov: Good point

psafarov:
@FlorianSW I can't say It has no bugs for sure, but this code is used in production in a company, no problems so far

src/JsonSchema/Constraints/TypeConstraint.php
psafarov: @FlorianSW parent constraint passes `checkMode` to the factory which in turn passes `checkMode` to the new constraint, so i don't see any problems here

bighappyface:
@psafarov I have merged a few PRs this morning and I put this PR out of sync. Would you please rebase on master?

psafarov:
@bighappyface done

bighappyface:
@psafarov sorry for the confusion. It looks like you merged master in okay but rebasing is different than merging master in. There should only be one commit in this PR.

psafarov:
@bighappyface fixed. But you're not going to merge it to master now, right? This pr contains breaking changes

bighappyface:
With community support I'll merge it into master. The plan is to merge it and then bump to 3.0.0 so indicate BC-breaking changes.

Other PRs I have merged have probably already put us in a position to bump to 3.0.0 anyway.

bighappyface:
Based on the passing tests, the account of this code being in production, the need to verified handling of circular references, and what feels like community approval and support, I am going to move this along and am going to cut release 3.0.0.

+1

neoascetic:
👍

Issue #266, opened 2016-05-17
Plugin-System for Validation of formats?


steffkes:
i'm using this library again for one of my projects, never had to extend it in any way since the included features were always enough - until this time ;>

there is already a part of `FormatConstraint` that validates `format=phone` but it only does a very, very simple check on US phone numbers. i did need to perform a check on a larger range including internal numbers. `libphonenumber-for-php` is a perfect fit for it, integrating that into `json-schema` and extending the `FormatConstraint` feels a bit ugly.

just for reference, one could or would probably:

```
class FormatConstraint extends \JsonSchema\Constraints\FormatConstraint
{
  public function check($element, $schema = null, $path = null, $i = null)
  {
    if($schema->format && $schema->format === 'phone') {
      try {
        PhoneNumberUtil::getInstance()->parse($element, NULL);
      } catch(NumberParseException $e) {
        $this->addError(
          $path,
          sprintf('Invalid phone number: %s', $e->getMessage()),
          'format',
          ['format' => $schema->format]
        );
      }
    }
  }

  protected function validatePhone($phoneNumber) {
    return true;
  }
}
```

since we extend `FormatConstraint` we'd have to give `validatePhone` a pass since it will be called from the original `FormatConstraint#check` method which is executed anyway since we have to ensure that all the other/existing format constraints still work.

the other thing is, since i'd like to include additional information in the response i need access to the exception - not a thing that all the existing `validate*` methods do right now. obviously a personal choice, but nevertheless. 

before i start digging in such a plugin-system, i'm curious WDYT? is it even worth it? am i the only one that needed something like this? is it probably the only extension one could ever want?

let me know!

mirfilip:
@steffkes Since `2.0.0` [there is a possibility to override constraints](https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/Factory.php#L68) by setting your own `FormatConstraint`. It's not quite exactly what you say, but theoretically it allows anyone to extend `FormatConstraint` and change `check` method so that it uses your new fancy way of `phone` format validation and passes all the other format types to original implementation.

steffkes:
@mirfilip probably should have been clearer in my initial description - that is exactly what i did, hence the sample `FormatConstraint` i've posted above.

it's *kinda okay*, but especially for existing formats you are going to overwrite you have to ensure that they pass the initial implementation since it's still executed - at least in my case when calling `parent::check` because i do need the existing checks on `date`, `email` and so on.

[...]

PR #280, opened 2016-06-15
[WIP] refactor FormatConstraint, use factory approach

steffkes:
Based on our discussion at #266 a very quick & dirty shot to show the approach i'd thought of.

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -8,7 +8,9 @@
  */
 
 namespace JsonSchema\Constraints;
+
 use JsonSchema\Rfc3339;
+use InvalidArgumentException;
 
 /**
  * Validates against the "format" property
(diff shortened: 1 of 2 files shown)


steffkes:
even w/o waiting for the build to end, i'm pretty sure it will fail on at least one version - i've put that together on a system that runs PHP 7. which allows a bit more / newer feature than the 5.3 release we need per our requirements

steffkes:
didn't rerun the tests after the latest change, which explains why it failed entirely -.- my bad, fixed pushed already.

narcoticfresh:
@steffkes sorry for the late checking.. 

yea, that looks feasable.. the only thing is that you create many throw-away instances, don't know if this could have much impact and how GC handles it.. 

maybe make a local `$instanceCache` array where you instanciate (and put them there) the first time, then always call `validate()` if it exists (which then needs an `Interface` for the Validator class) in the instance in the array..

We have call graphs that even call `ObjectConstraint->check()` around 2000 times - `Format` checks more by factors then of course - so small stuff like this quickly sums up.

digitalkaoz:
@narcoticfresh check again with the latest PRs merged...should be much better now

bighappyface:
@steffkes is this PR still in play?

steffkes:
@bighappyface thanks for asking, that's a valid question: i guess not. not actively working on a project anymore that uses json-schema that much. so there's currently not much time to work on this PR - i'm sorry for that.

bighappyface:
@steffkes no worries. Thanks for the update.

PR #285, opened 2016-06-24
Added I18n to validation messages

MarcosMenegazzo:
This PR adds I18n to validation messages, initially available in en_US and pt_BR. For additional languages, add it to `config/locales`, following the previous conventions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ language: php
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
-    - php: 5.4
     - php: 5.5
     - php: 5.6
       env: WITH_COVERAGE=true
@@ -14,8 +12,6 @@ matrix:
 
 before_install:
   - if [[ "$WITH_COVERAGE" != "true" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
-  - printf "\n" | pecl install yaml
-
   - composer selfupdate
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ matrix:
 
 before_install:
   - if [[ "$WITH_COVERAGE" != "true" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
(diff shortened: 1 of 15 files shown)


MarcosMenegazzo:
Seems that YAML is a bit tricky to setup in all environments listed at travis.yml, in particular PHP7  and HHVM

bighappyface:
@MarcosMenegazzo thank you for your contribution and hard work, but I do not think this PR is going to make it.

First, I am not sure I agree with the implementation of nor the need for human-readable translations in this library. I am far more in favor of a language-agnostic means of identifying specific validation failure causes as such an approach would eliminate the need to maintain languages in a package that is supposed to be focused on JSON schema facilities.

Second, cutting backward compatibility for the sake of making your changes easier doesn't sit very well with me. It is already difficult enough maintaining BC in PRs that focus on the core functionality of this package.

If you can garner enough community support for this PR then I am always ready to act on the will of the community for this package, but as far as I am concerned, I am -1.

MarcosMenegazzo:
@bighappyface Thank you for the feedback, i agree with you about backward compatibility. I can provide full backward compatibility, no problem.

But I do not agree about the I18n. I18n must be a standard for all libraries and web applications today.

bighappyface:
@MarcosMenegazzo I agree that i18n should be standard for libraries and web applications, and I believe that this library can support i18n _implicitly_.

Simply put, I do not think there is a need to return human-readable messages in validation errors precisely because of PRs such as this one. The validator can and does return error codes describing the error state that can be used in a consumer web application to provide feedback to the user in any manner they choose, in any language they choose.

The human-readable error messaging in this package, IMO, does not really add any value. In addition to the need for translations there are also situations where the semantics or wording of the message is unclear or not accurate. Between i18n and comprehension concerns the error messages cause more problems than they solve and augmenting the library like this makes everything more difficult. There are enough issues with core functionality as it is.

Again, I am not in support of this but am open if the community wants it.

MarcosMenegazzo:
@bighappyface Thanks again for your considerations.  May you please clarify what means "support i18n implicitly"?

MarcosMenegazzo:
Currently i do not have time to garner community support, so i'm closing this PR.

PR #287, opened 2016-07-07
Updating to use more correct hostname regex. This fixes #129

### LINKED ISSUE #129: format: hostname is too restrictive
Currently the regexp ^[_a-z]+\.([_a-z]+\.?)+$ is used to validate hostnames.
However hyphens should be allowed in the middle, also digits.

Perhaps see http://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address for better regular expressions usable for this.

There are multiple RFCs with slight differences in what is allowed. Unfortunately json-schema refers to RFC 1034, which does not actually define what is permissible in a host name at all, other than that it contains dots (and maybe not even that if it is relative and consists of only one label).

sinbonders:
Also, `localhost` is definitely a valid hostname...

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -173,6 +173,7 @@ class FormatConstraint extends Constraint
 
     protected function validateHostname($host)
     {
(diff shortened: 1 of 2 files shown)


bighappyface:
+1

PR #286, opened 2016-06-28
Support json_decode assoc

boekkooi-fresh:
| Q             | A
| ------------- | ---
| Branch?       | "master"
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | Since `Constraint::CHECK_MODE_TYPE_CAST` was not actively used I would say no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #267
| License       | MIT

Hi there,

I was running into the problem where `json_decode($input, true)` was not validation as I was hoping. 
To fix the problem I implemented a special `TypeCheckInterface` this will now be used to determine if something is a object, array and will do special logic to get properties etc.

I need to add some special cases where tests failed because of the array structure.

To use the `json_decode($input, true)` with the `Validator` a user must set the checkMode to `Constraint::CHECK_MODE_TYPE_CAST`.

I hope you like the PR and I'm looking forward to the comments.

Cheers,
Warnar

PR #288, opened 2016-07-08
some optimizations

narcoticfresh:
Hi guys..

this PR tries to minimize the calls done by the validation. Most is quite minor, but if you're dealing with call graphs like this (check the call numbers) - even small things matter..

call graph

I could squeeze out ~100ms out with those changes, whole request is now ~1.5s, so it saved maybe ~6-7%.. I'm still on the quest to squeeze more out, but only those were the low hanging fruits I've found..

It seems there is a certain threshold where validation is 'quite quick' - if the graph becomes quite big (like in our case); there is a continuous increase in request time. Makes sense of course.. 

There is something I commented out in this PR where I wasn't sure why it's even there, I'll comment it separately..

### DIFF
--- a/src/JsonSchema/Constraints/ObjectConstraint.php
+++ b/src/JsonSchema/Constraints/ObjectConstraint.php
@@ -105,11 +105,6 @@ class ObjectConstraint extends Constraint
                 $this->addError($path, "The presence of the property " . $i . " requires that " . $require . " also be present", 'requires');
             }
 
-            if (!$definition) {
-                // normal property verification
-                $this->checkUndefined($value, new \stdClass(), $path, $i);
-            }
-
             $property = $this->getProperty($element, $i, new UndefinedConstraint());
             if (is_object($property)) {
                 $this->validateMinMaxConstraint(!($property instanceof UndefinedConstraint) ? $property : $element, $definition, $path);
@@ -129,7 +124,11 @@ class ObjectConstraint extends Constraint
         foreach ($objectDefinition as $i => $value) {
             $property = $this->getProperty($element, $i, $this->getFactory()->createInstanceFor('undefined'));
             $definition = $this->getProperty($objectDefinition, $i);
-            $this->checkUndefined($property, $definition, $path, $i);
+
+            if (is_object($definition)) {
+                // Undefined constraint will check for is_object() and quit if is not - so why pass it?
+                $this->checkUndefined($property, $definition, $path, $i);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/JsonSchema/Constraints/ObjectConstraint.php
```diff
@@ -103,10 +103,10 @@ public function validateElement($element, $matches, $objectDefinition = null, $p
                 $this->addError($path, "The presence of the property " . $i . " requires that " . $require . " also be present", 'requires');
             }
 
-            if (!$definition) {
+            //if (!$definition) {
                 // normal property verification
-                $this->checkUndefined($value, new \stdClass(), $path, $i);
-            }
+                //$this->checkUndefined($value, new \stdClass(), $path, $i);
```
narcoticfresh: I really don't know why this was there.. what does the comment "`// normal property verification`" mean?

If we have no schema at all, is there any need to continue the recursion? If I check `UndefinedConstraint`, all it does is always checking stuff in `$schema` which was an empty `\stdClass()` all the way - anybody knows more?

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -98,45 +98,26 @@ public function check($value = null, $schema = null, $path = null, $i = null)
...
     protected function validateType($value, $type)
     {
-        //mostly the case for inline schema
         if (!$type) {
+            //mostly the case for inline schema
             return true;
-        }
-
-        if ('integer' === $type) {
+        } elseif ('integer' === $type) {
```
boekkooi-fresh: There is no need for a `elseif` because return is used in the previous `if`

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -52,29 +52,23 @@ public function check($value, $schema = null, $path = null, $i = null)
...
     {
-        // check array
         if (is_array($value)) {
+            // check array
             $this->checkArray($value, $schema, $path, $i);
-        }
-
-        // check object
-        if (is_object($value)) {
+        } elseif (is_object($value)) {
```
boekkooi-fresh: If PR #286 would get merged this `elseif` will need to be changed to `if` again

bighappyface:
@narcoticfresh please rebase this on master

narcoticfresh:
@bighappyface  rebased and squashed - i also removed some things as suggested by earlier comments..

bighappyface:
@narcoticfresh I merged a few PRs this morning and I want to be sure this wasn't affected before I merge. Would you please rebase once more? Once the tests run/pass I will merge this PR.

narcoticfresh:
@bighappyface rebased & squashed again.. please merge ;-)

bighappyface:
+1

PR #294, opened 2016-07-25
Change error reporting for invalid types with multiple valid types

### LINKED ISSUE #293: An array of valid value types is not respected in error message
If an allowed value type for a property is defined as an array, like:

```
"ExtensionFunctions": {
	"type": [
		"array",
		"string"
	],
	"description": "Function to call after setup has finished",
	"items": {
		"type": "string"
	}
},
```
the error message, if a value does not conform to this schema, is somtheing like:
`[ExtensionFunctions] Object value found, but a string is required`

while it should be something like:
`[ExtensionFunctions] Object value found, but an array, a string or an integer is required`

The error message is misleading, as it indicates, that only the type `string` is valid, while in fact other types are valid, too. The error mesage should indicate all valid types, instead of just one of them.

Original report in Wikimedia Phabricator: https://phabricator.wikimedia.org/T141281

FlorianSW:
Currently, if a property has multiple valid types, an error message
when a value doesn't conform to the schema, would mention only the
last valid type, without mentioning the other valid types.

This commit changes, how the TypeConstraint works internally. It
does a little bit of refactoring (removing also the "Refacter this"
ToDo) and now mentions any valid types on a failure.

Fixes #293

### DIFF
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -41,48 +41,99 @@ class TypeConstraint extends Constraint
     public function check($value = null, $schema = null, $path = null, $i = null)
     {
         $type = isset($schema->type) ? $schema->type : null;
-        $isValid = true;
+        $isValid = false;
+        $wording = array();
 
         if (is_array($type)) {
-            // @TODO refactor
-            $validatedOneType = false;
-            $errors = array();
-            foreach ($type as $tp) {
-                $validator = new static($this->checkMode);
-                $subSchema = new \stdClass();
-                $subSchema->type = $tp;
-                $validator->check($value, $subSchema, $path, null);
-                $error = $validator->getErrors();
-
-                if (!count($error)) {
-                    $validatedOneType = true;
-                    break;
-                }
-
-                $errors = $error;
-            }
-
-            if (!$validatedOneType) {
-                $this->addErrors($errors);
-
-                return;
-            }
+            $this->validateTypesArray($value, $type, $wording, $isValid, $path);
         } elseif (is_object($type)) {
             $this->checkUndefined($value, $type, $path);
+            return;
         } else {
             $isValid = $this->validateType($value, $type);
         }
 
         if ($isValid === false) {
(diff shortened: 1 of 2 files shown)


src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -41,48 +41,99 @@ class TypeConstraint extends Constraint
...
+     * @param boolean $isValid The current validation value
+     */
+    protected function validateTypesArray($value, array $type, &$validTypesWording, &$isValid,
+                                          $path) {
+        foreach ($type as $tp) {
+            // $tp can be an object, if it's a schema instead of a simple type, validate it
+            // with a new type constraint
+            if (is_object($tp)) {
+                if (!$isValid) {
+                    $validator = new static($this->checkMode);
```
FlorianSW: I thought we could get rid of this (at least for me) ugly looking workaround, but it seems it's the best solution we currently have :/

FlorianSW:
I'm not sure, but the failed test https://travis-ci.org/justinrainbow/json-schema/jobs/147291856 looks unrelated to this change?

FlorianSW:
@bighappyface Have you found some time to take a look at this commit? :-)

bighappyface:
@FlorianSW did we lose the test?

FlorianSW:
What test? I'm pretty sure we don't need extra test for this change as it's already covered by existing tests? :-)

bighappyface:
I feel like a test was added that would have encouraged me to remove the "needs tests" label. I don't recall it exactly but the comment history suggests a test was created.

Regardless, I feel because this PR changes behavior that a new test exhibiting the new behavior is desirable to ensure the outcome described in #293 has been achieved.

FlorianSW:
Ah, my bad, in my second commit I had in focus to unbreak existing tests (because of my refactoring), so I haven't noticed anymore, that I added some functionality. :D Let me add some tests! :)

FlorianSW:
@bighappyface Maybe you haven't seen it, but I've added some tests to an already existing test case :-)

bighappyface:
+1

PR #304, opened 2016-08-23
Return points, spelling and dead code

calcinai:
Replaces #302

### DIFF
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -40,7 +40,7 @@ class UriRetriever implements BaseUriRetrieverInterface
      *
      * @param UriRetrieverInterface $uriRetriever
      * @param string $uri
-     * @return bool
+     * @return bool|void
      */
     public function confirmMediaType($uriRetriever, $uri)
     {
@@ -48,11 +48,11 @@ class UriRetriever implements BaseUriRetrieverInterface
 
         if (is_null($contentType)) {
             // Well, we didn't get an invalid one
-            return true;
+            return;
         }
 
         if (in_array($contentType, array(Validator::SCHEMA_MEDIA_TYPE, 'application/json'))) {
-            return true;
+            return;
         }
 
         if (substr($uri, 0, 23) == 'http://json-schema.org/') {
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -40,7 +40,7 @@ class UriRetriever implements BaseUriRetrieverInterface
      *
      * @param UriRetrieverInterface $uriRetriever
      * @param string $uri
-     * @return bool|void
+     * @return bool
      */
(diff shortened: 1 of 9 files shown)


bighappyface:
@calcinai please check the PR build for test failures

calcinai:
Oh no! Do you think the tests should actually be updated in this case? From my understanding that function is designed to throw an exception or silently return?

calcinai:
Have updated the test, I can only assume that this is deserted behaviour - it would be an incredibly strange case if someone was actually looking for nulls in this case..

bighappyface:
@calcinai thanks for this work.

While the updating some return values to match doc blocks makes a lot of sense, I want to caution that there are some BC implications to changing return values. My assessment is that no external uses would change so I think we are good, but I would like a second opinion.

@FlorianSW @mirfilip @thewilkybarkid @narcoticfresh do you see any BC issues with this? Rolling 4.0.0 is not desireable to me at the present time.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -40,19 +40,19 @@ class UriRetriever implements BaseUriRetrieverInterface
...
+     * @return bool
      */
     public function confirmMediaType($uriRetriever, $uri)
     {
         $contentType = $uriRetriever->getContentType();
 
         if (is_null($contentType)) {
             // Well, we didn't get an invalid one
-            return;
+            return true;
```
FlorianSW: I'm not sure, how many users uses this method, but changing from null to a boolean value could be problematic, especially in code something like:

```
if ( confirmMediaType( $uriRetriever, $uri ) ) {
  // the code that runs when true
} else {
  // the former "null" value-code
}
```
It now would break, as null evaluates to false, and not to true. So in some cases, this could be a problem. However, the new values makes sense, too. I'm really not sure :(

calcinai: Given that (in master) the function could return `true` when it's a schema from json-schema.org, and `null` if it is explicitly valid, it just isn't seem all that logical.  The only usage internally just relies on it halting with an exception if it's invalid.

Is it even feasible that someone has written code to handle the different return types?  If you're using the library for validation, it's going to be called again anyway and you won't have any control of the result..

That said, maybe there is an argument for leaving it as it is, because the return value is basically irrelevant anyway!

bighappyface: @FlorianSW thank you so much for your feedback. This is exactly my concern.

As you pointed out, `null` is falsy. Perhaps returning `false` here is the solution as it would be compliant with the `bool` return type and also be compatible with previous control-flow usage, although that would not be true for `=== null` or `is_null` comparisons.

calcinai: Moving forward with this, should I just revert the section in question?

bighappyface: Maintaining BC is preferred.

bighappyface:
+1

PR #309, opened 2016-09-19
Update README.md

stamster:
Changed composer install method (it will target latest stable branch), since 2.0 is broken
added few comments to clarify usage example

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,18 +19,19 @@ See [json-schema](http://json-schema.org/) for more details.
 #### [`Composer`](https://github.com/composer/composer) (*will use the Composer ClassLoader*)
 
     $ wget http://getcomposer.org/composer.phar
-    $ php composer.phar require justinrainbow/json-schema:~2.0
+    $ php composer.phar require --no-update justinrainbow/json-schema
+    $ composer.phar update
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,18 +19,19 @@ See [json-schema](http://json-schema.org/) for more details.
...
 ```php
 <?php
 
-$data = json_decode(file_get_contents('data.json'));
+$data = json_decode(file_get_contents('data.json')); //NOTE: this needs to be an object, not an array!
 
 // Validate
 $validator = new JsonSchema\Validator;
-$validator->check($data, (object)['$ref' => 'file://' . realpath('schema.json')]);
+$validator->check($data, (object)['$ref' => 'file://' . realpath('schema.json')]); //either provide PHY schema (after json_decode) or just point out to the correct schema on the local filesystem
```
stamster: PHY - Physical. Might be misleading, sorry. The point was - either provide object after JSON decode, either provide a link to your schema on the filesystem.

bighappyface: Let's go with something more clear

bighappyface:
Closing pending clarity... probably addressed in #320

PR #314, opened 2016-09-28
Updated to check that $ref is a string

calcinai:
This is an issue when processing the OpenAPI/Swagger spec, where `$ref` is an object.

Example on line 315 of the OAPI spec:
```json
"$ref": {
      "type": "string"
},
```

From what I understand, the spec is a valid document, but beaks the current resolver logic.

### DIFF
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -98,7 +98,7 @@ class SchemaStorage
      */
     public function resolveRefSchema($refSchema)
     {
-        if (is_object($refSchema) && property_exists($refSchema, '$ref')) {
+        if (is_object($refSchema) && property_exists($refSchema, '$ref') && is_string($refSchema->{'$ref'})) {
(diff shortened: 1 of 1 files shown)


calcinai:
@bighappyface - what do you think about this one?

bighappyface:
@calcinai thanks for the contribution. To be clear, will this adjustment only allow refs that are strings in this library, or does the check pass through object refs to support the example you provided from OAPI?

calcinai:
From what I understood, `$ref` is always a string if it's actually a reference.  The issue here is that the current logic eagerly treats any object key `$ref` as a reference and tries to follow it, which in this case will fail.

The OAPI issue is that in the document it's actually defining what `$ref` can be.

Apologies if I didn't answer your question, can you elaborate if not?

calcinai:
Just to add - I deliberately made the smallest change possible to correct this to try and have the smallest impact.

bighappyface:
I think we are good here. +1

PR #316, opened 2016-10-09
fix(NumberConstraint): force positive value on first arg

### LINKED ISSUE #315: `multipleOf` is inconsistent, fails for negative numbers
The [specification](http://json-schema.org/latest/json-schema-validation.html#anchor14) doesn't say negative numbers should fail a `multipleOf` constraint, yet the current documentation fails `-49.89` with a `multipleOf` value of `0.01`.

Funnily enough, the implementation of this constraint has some edge cases where it lets negative numbers pass:

Calling the [`NumberConstraint::fmod` function](https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/NumberConstraint.php#L70) with the parameters:

- `-49.89` and `0.01`: returns `-0.01`, thus **fails**
- `-49.88` and `0.01`: returns `-0`, which **passes** due to the [non-strict comparison in the `NumberConstraint::check` function](https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/NumberConstraint.php#L63).

Calling `abs()` on the first argument of `NumberConstraint::fmod` should both fix this inconsistency and bring `multipleOf` in line with the spec.

bighappyface:
Resolves #315 - `multipleOf` is inconsistent, fails for negative numbers

### DIFF
--- a/src/JsonSchema/Constraints/NumberConstraint.php
+++ b/src/JsonSchema/Constraints/NumberConstraint.php
@@ -69,6 +69,7 @@ class NumberConstraint extends Constraint
 
     private function fmod($number1, $number2)
     {
+        $number1 = abs($number1);
         $modulus = fmod($number1, $number2);
         $precision = abs(0.0000000001);
         $diff = (float)($modulus - $number2);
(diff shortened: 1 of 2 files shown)


bighappyface:
@leongersen please review

leongersen:
`abs` has no side effects and doesn't negatively affect `multipleOf` behaviour, test shows a previously failing case. Looks good to me.

PR #325, opened 2016-10-24
handle negative integers

shmax:
Coercion of negative integers. 
Fixes https://github.com/justinrainbow/json-schema/issues/323

@bighappyface

### DIFF
--- a/src/JsonSchema/Constraints/ObjectConstraint.php
+++ b/src/JsonSchema/Constraints/ObjectConstraint.php
@@ -184,7 +184,7 @@ class ObjectConstraint extends Constraint
 
     protected function toInteger($value)
     {
-        if(ctype_digit ($value)) {
+        if(is_numeric($value) && (int)$value == $value) {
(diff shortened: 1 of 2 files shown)


src/JsonSchema/Constraints/ObjectConstraint.php
```diff
@@ -184,7 +184,7 @@ protected function toNumber($value)
 
     protected function toInteger($value)
     {
-        if(ctype_digit ($value)) {
+        if(is_numeric($value) && (string)(int)$value == $value) {
```
bighappyface: Is casting to string here necessary with loose equals?

```php
-1 == "-1" // true
(int)"-1" == "-1" // true
```

shmax: Not totally sure of the implications, to be honest. I suppose at some point I should cover all the edges in the tests, but for now I removed the extra cast and the existing tests still pass.

PR #326, opened 2016-10-27
update readme for 4.0.0

shmax:
A lot of people are getting confused by the fact that the readme doesn't reflect the current release version.  Until https://github.com/justinrainbow/json-schema/pull/320 gets merged, this will at least put the documentation in sync with the current state of the code.

@bighappyface @digitalkaoz @nwolff @stamster


jgao1:
We should also be using `$jsonValidator->check($jsonToValidateObject, (object)['$ref' => 'file://mySchema']);` instead of `$jsonValidator->check($jsonToValidateObject, $jsonSchemaObject);`? At least it's the first one that's working for me.

PR #329, opened 2016-10-30
Coerce items

shmax:
Adds support for coercion of `items` for schemas with a `type` of `array`:

``` php
$value = (object)[
    "foo"=>["7","9.1","true"]
];
$schema = (object)[
    "type"=>"object",
    "properties"=>(object)[
        "foo"=>(object)[
            "type"=>"array",
            "additionalItems"=>[
                "type"=>"boolean"
            ],
            "items"=>[
                (object)[
                    "type"=>"integer"
                ],
                (object)[
                    "type"=>"number"
                ]
            ]
        ]
    ]
];

echo gettype($value->foo[0]); // "string"
echo gettype($value->foo[1]); // "string"
echo gettype($value->foo[2]); // "string"

$validator->check($value, $schema);

echo gettype($value->foo[0]); // "integer"
echo gettype($value->foo[1]); // "double"
echo gettype($value->foo[2]); // "boolean"
```
- moved the coercion helpers out of `ObjectConstraint` and down into `Constraint`, because there are non-objects that need to coerce
- Got rid of `CollectionConstraint`. Because we need to pass arrays by reference to coerce them, and because `CollectionConstraint` was concerned exclusively with arrays of things, its `check` method was rendered useless. My choices seemed to be to factor it into the base `Constraint` class (which already had a `checkArray` method) or change all `check`implementations to take a reference; I chose the former because the latter would mean breaking backwards compatibility (only variables can be passed by reference, so any inline array literals being used out there in the wild would cause a run-time exception). Hope that's not a problem.

@bighappyface @digitalkaoz


shmax:
Would love some feedback, guys

shmax:
@bighappyface Is there any chance we can get this merged before https://github.com/justinrainbow/json-schema/pull/331 ? I'm not looking forward to resolving those conflicts...

bighappyface:
@shmax I can work with you on that. Is this PR rebased on the latest?

shmax:
@bighappyface Yep, good to go. No conflicts with the base branch.

shmax:
I'm working on a complete redesign of the coercion functionality that makes this PR unnecessary.

PR #330, opened 2016-11-07
[BUGFIX] Use unicode matching for patterns

jbaron-gingco:
JSON allows all unicode characters, thus the matching should work when using unicode characters.

PR #331, opened 2016-11-08
improve validation performance of long string, number and integer arrays

SereneAnt:
Validation is unexpectedly slow for large uniform arrays with simple json schema like below:
```
{
  "type": "array",
    "items": {
      "type": "string"
    }
}
```
For 100K string arrays validation lasts for about 2000 ms.
With such a workaround, the performance gain is 3X.
It can be improved twice more with lazy $path calculation on validation error.

Please note, the proposed fix is not the 'true' way - major refactoring with api changes is required,
it rather pinpoints the problem.

### DIFF
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -63,23 +63,51 @@ class CollectionConstraint extends Constraint
     {
         if (is_object($schema->items)) {
             // just one type definition for the whole array
-            foreach ($value as $k => $v) {
-                $initErrors = $this->getErrors();
 
-                // First check if its defined in "items"
-                $this->checkUndefined($v, $schema->items, $path, $k);
+            if (isset($schema->items->type)
+                && (
+                    $schema->items->type == 'string'
+                    || $schema->items->type == 'number'
+                    || $schema->items->type == 'integer'
+                )
+                && !isset($schema->additionalItems)
+            ) {
+                // performance optimization
+                $type = $schema->items->type;
+                $validator = $this->factory->createInstanceFor($type === 'integer' ? 'number' : $type);
 
-                // Recheck with "additionalItems" if the first test fails
-                if (count($initErrors) < count($this->getErrors()) && (isset($schema->additionalItems) && $schema->additionalItems !== false)) {
-                    $secondErrors = $this->getErrors();
-                    $this->checkUndefined($v, $schema->additionalItems, $path, $k);
+                foreach ($value as $k => $v) {
+                    $k_path = $this->incrementPath($path, $k);
+
+                    if (($type === 'string' && !is_string($v))
+                        || ($type === 'number' && !(is_numeric($v) && !is_string($v)))
+                        || ($type === 'integer' && !is_int($v))
+                    ){
+                        $this->addError($k_path, ucwords(gettype($v)) . " value found, but $type is required", 'type');
+                    } else {
+                        $validator->check($v, $schema, $k_path, $i);
+                    }
                 }
+                $this->addErrors($validator->getErrors());
+            } else {
+                foreach ($value as $k => $v) {
+                    $initErrors = $this->getErrors();
 
-                // Reset errors if needed
-                if (isset($secondErrors) && count($secondErrors) < count($this->getErrors())) {
-                    $this->errors = $secondErrors;
-                } elseif (isset($secondErrors) && count($secondErrors) === count($this->getErrors())) {
-                    $this->errors = $initErrors;
+                    // First check if its defined in "items"
+                    $this->checkUndefined($v, $schema->items, $path, $k);
+
+                    // Recheck with "additionalItems" if the first test fails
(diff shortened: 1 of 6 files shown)


shmax:
The maintainers of this repo are fairly dormant these days. If you feel you can refactor the project in such a way as to meaningfully improve performance I don't think anyone would object, but none of the maintainers are likely to do it for you. As far as changing the API goes, I don't think small changes are necessarily a problem as long as they are well justified.

bighappyface:
I see no BC issues here and the tests are all good.

I am +1. @shmax are you good with this?

shmax:
Well, I don't think he means this to be ready for merge--it's more of a demonstration of the core problem. Some of the concepts in this PR are sort of quick-and-dirty and not fully realized. For example, ```isSimpleSchema``` just checks to see if the type is "string", but what does that mean? Is a ```string``` type any simpler, than say, an ```integer``` type?

@SereneAnt do you intend to flesh this out into a full solution?

SereneAnt:
Sure, let me isolate bulk-optimized code into a separate class, sayTypedItemsCollectionConstraint and apply the same approach for other primitive types.
I'll try to not change api drastically.

bighappyface:
With API changes coming in this PR I will combine them with others from @shmax to roll another major release. In the meantime I am going to roll a patch release for the changes in #330

SereneAnt:
Ready for merge.
No API changes made to keep things solid.

src/JsonSchema/Constraints/CollectionConstraint.php
```diff
@@ -63,23 +63,29 @@ protected function validateItems($value, $schema = null, JsonPointer $path = nul
     {
         if (is_object($schema->items)) {
             // just one type definition for the whole array
-            foreach ($value as $k => $v) {
-                $initErrors = $this->getErrors();
+            if (SimpleItemsConstraint::isSupported($schema)) {
```
shmax: I find this a little inconsistent with the rest of the codebase... the constraint classes don't generally have functional dependencies on each other outside of methods specified in `ConstraintInterface` ( eg. `check()` ). Is a new constraint type really necessary? What do you think about just doing everything you need to do inside of `CollectionConstraint`?

src/JsonSchema/Constraints/CollectionConstraint.php
```diff
@@ -111,4 +130,17 @@ protected function validateItems($value, $schema = null, JsonPointer $path = nul
...
         }
     }
+
+    const OPTIMIZED_ITEM_TYPES = array('string', 'number', 'integer');
+
+    /**
+     * @param \stdClass $schema
+     * @return bool
+     */
+    private function isSupported($schema)
```
shmax: `isSupported` is a little vague. Can you think of a better function name that better describes what it does? Given that you only call it in one place, another option would be to just remove it and inline the code you have here.

src/JsonSchema/Constraints/CollectionConstraint.php
```diff
@@ -111,4 +130,17 @@ protected function validateItems($value, $schema = null, JsonPointer $path = nul
             }
         }
     }
+
+    const OPTIMIZED_ITEM_TYPES = array('string', 'number', 'integer');
```
shmax: You'll have to come up with some other way to encode these values, unfortunatley; older versions of PHP don't allow assigning arrays to constants.

SereneAnt: fixed

src/JsonSchema/Constraints/CollectionConstraint.php
SereneAnt: the method has been inlined.

src/JsonSchema/Constraints/CollectionConstraint.php
```diff
@@ -111,4 +139,19 @@ protected function validateItems($value, $schema = null, JsonPointer $path = nul
             }
         }
     }
+
+    /**
+     * @param \stdClass $schema
+     * @return bool
+     */
+    private function isOptimized($schema)
```
SereneAnt: oops)
done

shmax:
LGTM

tests/Constraints/LongArraysTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $schemaStorage = new SchemaStorage($this->getUriRetrieverMock(json_decode($schema)));
+        $schema = $schemaStorage->getSchema('http://www.my-domain.com/schema.json');
+
+        $validator = new Validator(new Factory($schemaStorage));
+        $validator->check(json_decode($input), $schema);
+        $this->assertTrue($validator->isValid(), print_r($validator->getErrors(), true));
+    }
+
+    private static function millis()
```
bighappyface: Unused?

SereneAnt: Reserved for future use. This unit test looks rather like performance test.
No idea how to organize performance testing in PHP properly.
If you insist - I'll kill that)

SereneAnt: I'd prefer to add trait named PerformanceTest with a conditional output of running time to console. Since we support PHP 5.3 (no traits), I have no idea of a proper way of doing that.

PR #335, opened 2016-11-25
Fix various details in README.md

rtucek:
Add newline after heading and code block (consistency).

Use triple backticks for bash code examples.

Change how library will be installed via composer.

Remove explicit version constraint in `composer require` example. Recent
versions of composer will default the most recent version and
set a non-major breaking constraint.
Version constraint was still pointing to the most recent version of 2.*
release.

Soft limit text to 80 chars (does node affect code examples).

Remove 'Basic usage' heading.

Fix heading 'Type coercion' (space and typo).

Add missing opening PHP tag (section 'Type coercion'.

Fix indentation in code example (consistent 4 spaces).

Use single backticks for inline code.
Add parenthesis to method in description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,18 +12,19 @@ See [json-schema](http://json-schema.org/) for more details.
 
 ### Library
 
-    $ git clone https://github.com/justinrainbow/json-schema.git
+```bash
+git clone https://github.com/justinrainbow/json-schema.git
+```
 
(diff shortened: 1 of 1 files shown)


rtucek:
Btw, it took me longer to document the changes in the commit message than changing in the source... 😄

PR #336, opened 2016-11-26
Better coercion

shmax:
This is a second attempt at providing coercion support. My first attempt suffered from a few faults:

1. It piggy-backed on `check()`--which did not take a reference as a first argument--so only objects passed in as variables were affected by coercion.
2. Again due to the limitation on references not always being available, I had to sort of sneak coercion in here and there where I could, and it wasn't always possible for use cases like `allOf`

This new approach has a number of advantages and improvements over the older one:

* added a new top-level ```coerce``` method to kick things off. It takes a reference, and everything after that point switches between `coerce` and `check` methods as needed. `check` is unaffected.
* you can coerce non-objects... eg
```php
$value = "false";
$schema = (object)[
  "type":"boolean"
];
$validator->coerce($value, $schema); // validates
echo get_type($value); // "boolean"
```
* coercion now happens more deeply in the code, at the `TypeConstraint` level. This means that as long as the right `coerce` flags are passed along and references are used where appropriate internally, coercion just sort of works properly without having to wedge additional code in all over the place.

* The `CHECK_MODE_COERCE` flag is no longer necessary, so you're no longer obliged to create a `Factory`, which means simpler client code in most cases.

Fixes https://github.com/justinrainbow/json-schema/issues/334, https://github.com/justinrainbow/json-schema/pull/329 and https://github.com/justinrainbow/json-schema/pull/322

API Note: while this does not break the `check()` API, you must now use `coerce()` to do coercion. The `CHECK_MODE_COERCE`  flag has been removed.

Touched a lot of files, apologies, but hopefully it's not too hard to follow.

@bighappyface @pcjacobse @digitalkaoz please review

### DIFF
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -72,6 +72,7 @@ class CollectionConstraint extends Constraint
      * @param \stdClass        $schema
      * @param JsonPointer|null $path
      * @param string           $i
+	 * @param boolean          $coerce
      */
     protected function validateItems(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
     {
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -19,10 +19,24 @@ use JsonSchema\Entity\JsonPointer;
  */
 class CollectionConstraint extends Constraint
 {
+
     /**
      * {@inheritDoc}
      */
     public function check($value, $schema = null, JsonPointer $path = null, $i = null)
+    {
+        $this->_check($value, $schema, $path, $i);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = null)
+    {
+        $this->_check($value, $schema, $path, $i, true);
+    }
+
+    protected function _check(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
     {
         // Verify minItems
         if (isset($schema->minItems) && count($value) < $schema->minItems) {
@@ -47,7 +61,7 @@ class CollectionConstraint extends Constraint
 
         // Verify items
         if (isset($schema->items)) {
-            $this->validateItems($value, $schema, $path, $i);
+            $this->validateItems($value, $schema, $path, $i, $coerce);
         }
     }
 
@@ -59,7 +73,7 @@ class CollectionConstraint extends Constraint
      * @param JsonPointer|null $path
      * @param string           $i
      */
-    protected function validateItems($value, $schema = null, JsonPointer $path = null, $i = null)
+    protected function validateItems(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
     {
         if (is_object($schema->items)) {
             // just one type definition for the whole array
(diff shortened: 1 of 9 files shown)


shmax:
Please review!

digitalkaoz:
I did...

Max Loeb <notifications@github.com> schrieb am Mi., 30. Nov. 2016, 18:41:

shmax:
@digitalkaoz Did you? I don't see any comments, stamps, or 'approved' events. Did I miss something?

digitalkaoz:
ah now, forgot to submit the review

src/JsonSchema/Constraints/CollectionConstraint.php
```diff
@@ -19,10 +19,24 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = null)
+	{
+		$this->_check($value, $schema, $path, $i, true);
+	}
+
+    protected function _check(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
```
shmax: Interesting idea, but unfortunately any kind of coercion that depends on `check` is a non-starter because I would have to change the first argument to a reference, which *would* break bc.

shmax:
Fixed the whitespace issues, thanks; one of these days I should quit being lazy and modify my IDE settings.

shmax:
@bighappyface How are your plans for the next release coming along?

bighappyface:
4.1.0 with this PR?

shmax:
Sounds good. Are we waiting on anything? I believe I'll need to rebase once you merge https://github.com/justinrainbow/json-schema/pull/331

shmax:
@bighappyface Rebased. 
@SereneAnt I de-simplified your code a bit; I've brought in a `TypeConstraint` validator to replace the stuff you were doing inline because I need coercion to work. I don't think it should slow things down much, if at all, because it pretty much does the same things you were doing, only within its own domain (so important functionality like coercion happens). If it seems like a problem to you, we can just add a check for `$coerce` to the conditions we consider before branching into your optimized code.

Please review!

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -125,10 +124,14 @@ protected function incrementPath(JsonPointer $path = null, $i)
      * @param JsonPointer|null $path
      * @param mixed            $i
      */
-    protected function checkArray($value, $schema = null, JsonPointer $path = null, $i = null)
+    protected function checkArray(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
```
shmax: Ah, good eye, thank you. I've added phpdocs, following the patterns and style already in place.

As for adding $coerce as a `Factory` flag, well, that was the approach I tried previously. It made sense in my previous approach because I was driving everything through `check`, and it just became a `mode` flag. The problem there was that I had no reference to work with in `check`, so I couldn't properly coerce in all situations. 

Now that I'm driving it through an explicit function call, `coerce`, it's no longer really a configuration value. If one were to, say, call `check` immediately after calling `coerce`, I would just have to set any stored fields back to false, so it would be another moving part with no real value.

I was actually interested in your reaction to what I did in `CollectionConstraint`. Any thoughts?

shmax:
@bighappyface please merge, or let me know what I have to do to move forward.

src/JsonSchema/Constraints/Constraint.php
SereneAnt: I have no write permissions, can't approve.

SereneAnt:
LGTM

pcjacobse:
Sorry for the late reply, but works like a charm ;)

shmax:
@bighappyface What's the plan?

shmax:
Thank you!

alcohol:
Nobody noticed the mixed spaces and tabs indentation?

shmax:
Nope, nobody did, and you're commenting on a PR that was merged almost two months ago. If you can find formatting issues in the current state of the code I'll be happy to clean 'em up.

alcohol:
Yeah, sorry, I was going through the changes between 4.x and 5.x as I wanted to see if Composer can easily upgrade/be compatible with the latest major version (we currently are compatible with 3.x and 4.x). Noticed some inconsistencies in the change sets. Hence my code-style PR submission :-)

bighappyface:
Thanks for the help ensuring things stay smooth with Composer, @alcohol.

Being a dependency for Composer is our crown jewel 👑

PR #339, opened 2016-12-03
Added SchemaStorageInterface

kynx:
This PR adds an interface for SchemaStorage. This enables users to supply alternate storage implementations (for instance, if they have persistent storage or wish to implement a schema cache).


shmax:
LGTM

PR #342, opened 2016-12-12
fix  #341 validate-json command

### LINKED ISSUE #341: Command seems broken
```
$ grep -ri RefResolver .
./bin/validate-json:$refResolver = new JsonSchema\RefResolver($retriever, $resolver);
./bin/validate-json:$schema = $refResolver->resolve($urlSchema);

```
But this class didn't exists in 4.x version, so:

`PHP Fatal error:  Uncaught Error: Class 'JsonSchema\RefResolver' not found `

remicollet:
Seems to work, please review as I'm probably not an expert of this library.

### DIFF
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -207,8 +207,8 @@ try {
     echo $e->getMessage() . "\n";
     exit(2);
 }
-$refResolver = new JsonSchema\RefResolver($retriever, $resolver);
-$schema = $refResolver->resolve($urlSchema);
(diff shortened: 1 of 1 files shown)


shmax:
We're not all experts, either. Perhaps you could provide a little background? Eg. what problem are you solving here, what was wrong, what does your fix change, etc...

remicollet:
@shmax see #341

shmax:
Ah, okay. LGTM

PR #345, opened 2017-01-11
Allow coercion to boolean from '0' and '1'

JanTvrdik:
Both `'1'` and `'0'` are very common string representations of boolean values.

### DIFF
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -225,11 +225,11 @@ class TypeConstraint extends Constraint
      */
     protected function toBoolean($value)
     {
-        if($value === "true"){
+        if ($value === 'true' || $value === '1') {
             return true;
         }
 
-        if($value === "false"){
+        if ($value === 'false' || $value === '0') {
             return false;
         }
 
--- a/tests/Constraints/CoerciveTest.php
+++ b/tests/Constraints/CoerciveTest.php
@@ -50,28 +50,32 @@ class CoerciveTest extends BasicTypesTest
 
         $this->assertTrue(gettype($value->number) == "string");
         $this->assertTrue(gettype($value->integer) == "string");
-        $this->assertTrue(gettype($value->boolean) == "string");
+        $this->assertTrue(gettype($value->boolean1) == "string");
+        $this->assertTrue(gettype($value->boolean2) == "string");
+        $this->assertTrue(gettype($value->boolean3) == "string");
+        $this->assertTrue(gettype($value->boolean4) == "string");
 
         $validator->coerce($value, $schema);
 
         $this->assertTrue(gettype($value->number) == "double");
         $this->assertTrue(gettype($value->integer) == "integer");
         $this->assertTrue(gettype($value->negativeInteger) == "integer");
-        $this->assertTrue(gettype($value->boolean) == "boolean");
+        $this->assertTrue(gettype($value->boolean1) == "boolean");
+        $this->assertTrue(gettype($value->boolean2) == "boolean");
+        $this->assertTrue(gettype($value->boolean3) == "boolean");
+        $this->assertTrue(gettype($value->boolean4) == "boolean");
 
         $this->assertTrue($value->number === 1.5);
         $this->assertTrue($value->integer === 1);
         $this->assertTrue($value->negativeInteger === -2);
-        $this->assertTrue($value->boolean === true);
+        $this->assertTrue($value->boolean1 === true);
+        $this->assertTrue($value->boolean2 === false);
+        $this->assertTrue($value->boolean3 === true);
+        $this->assertTrue($value->boolean4 === false);
 
         $this->assertTrue(gettype($value->multitype1) == "boolean");
         $this->assertTrue(gettype($value->multitype2) == "double");
         $this->assertTrue(gettype($value->multitype3) == "integer");
-
-        $this->assertTrue($value->number === 1.5);
-        $this->assertTrue($value->integer === 1);
-        $this->assertTrue($value->negativeInteger === -2);
-        $this->assertTrue($value->boolean === true);
 
         $this->assertTrue($validator->isValid(), print_r($validator->getErrors(), true));
     }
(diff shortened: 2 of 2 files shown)


shmax:
I'm not totally sure about this one, mainly because if you want to use "1" and "0" you can just define a type of "integer" and use the string coercion that already exists, right? I want to check a few other libraries and see what their policy is for this kind of thing... will have more information soon.

moderndeveloperllc:
That would seem to be contrary to RFC 7159 that defines [JSON grammar](https://tools.ietf.org/html/rfc7159#section-2). There is a caveat that [JSON parsers](https://tools.ietf.org/html/rfc7159#section-9) "MAY accept non-JSON forms or extensions", but that could be tricky. The XML Schema Part 2 spec [does define](https://www.w3.org/TR/xmlschema-2/#boolean) `true`, `false`, `0`, and `1` as legal literals for a boolean, but not sure how much that matters.

It would be great if @awwright could be summoned to give his input on 1/0/true/false for boolean in JSON Schema as opposed to only true/false.

shmax:
@moderndeveloperllc I don't think this really has anything to do with the JSON spec, or even the JSON schema spec. We're talking about a piece of functionality that is commonly layered onto validation libraries, specifically. This library already supports some coercion (it has to be invoked explicitly with the `coerce` method), and the question here is whether *string* 1 and *string* 0 should be coerced to boolean. 

I haven't had a chance to research this in depth yet, but one of the most robust and fully featured validation libraries I know of does not coerce string "1" and "0" to boolean (though it *does* coerce numeric 1 and 0 to boolean):

https://github.com/epoberezkin/ajv/blob/master/COERCION.md

JanTvrdik:
My use  case is to validate query parameters ($_GET) with JSON schema, therefore coercion from integers to booleans would not help me as query parameters are always string or array.

shmax:
Of course, that exact use case is why we added coercion in the first place. My point is that there is no need for you to coerce your string 1/0 to boolean; you can simply change your expected type to "integer", then check them in your client PHP code for truthiness the exact same way you would check a boolean for truthiness.

awwright:
@JanTvrdik I'm not sure what "coercion" means in this case. All a validator does is accept an instance and return "valid" or "invalid". 

What's the proposed change in behavior, and how does it vary from the current behavior?

shmax:
@awwright See https://github.com/justinrainbow/json-schema/pull/308, https://github.com/justinrainbow/json-schema/pull/336, and https://github.com/justinrainbow/json-schema/blob/master/README.md#type-coercion

shmax:
One more comment for @JanTvrdik:

`toBoolean` is only a protected method--not private--which means that you are free to simply subclass TypeConstraint and override `toBoolean` to get the functionality you desire. I can provide a code sample if you have trouble.

awwright:
Oooh yes.

Of course, JSON Schema only specifies validation.  But I've implemented "coercion" like this, i.e. take some data (JSON or otherwise) and cast it into JSON valid against a certain schema. I did this for taking a URI and turning it into a JSON-compatible object.

This just gets a little complex if you support multiple data types, for example `type: ["string", "boolean"]`. That's the only consideration I can think of.

shmax:
I was about to argue that "1" and "0" are different from the cases already handled in that they are not simply a reversal of the stringification performed by operations like `http_build_query`, but I just tried it and 'http_build_query' *does* string-ify true and false as 1 and 0. 

Sooo, I guess I'm not as sure about this, now. I would still like to know why the folks on the other project don't coerce "1" and "0"...

shmax:
So we have this elegant counter-example and explanation from the author of ajv: https://github.com/epoberezkin/ajv/issues/395#issuecomment-272743288

Essentially, because coercion of string 1 and 0 to boolean is not reversible, we run into cases where validation on type lists (and possibly other complex nested structures) can unexpectedly fail with coercion turned on, when it would pass with it turned off. 

So, @JanTvrdik I think your options are still the same:

1. stay with your string 1 and 0 but change your schema to expect an integer. Use `coerce`. Your client code can continue to use the same logical checks you would do as if they were boolean
2. Prepare your http requests such that true and false are stringified to "true" and "false" instead of "1" and "0", and coerce
3. Subclass `TypeConstraint` in your own code and do whatever you like with `toBoolean`, but be prepared for subtle issues like the one described by @epoberezkin

JanTvrdik:
> we have this elegant counter-example and explanation from the author of ajv

@shmax But his points are irrelevant int context of this library because this library does not suport multiple types with anyOf nor does it support reversing the coercion. You cannot reproduce his example with this library.

shmax:
> You cannot reproduce his example with this library.

I did try it, and what happened was exactly what he predicted. Did you get different results?

> this library does not suport multiple types with anyOf

What do you mean? Do you mean that coercion is not fully supported? If you can supply a failing unit test we can investigate it together.

> nor does it support reversing the coercion

Nobody is saying it should support reversing the coercion. When we say that the coercion of 1 and 0 is not reversible we mean that it is one way; once coerced its original form is lost, so that a further rule that might have otherwise validated its original form will fail (like in the schema sample we've been looking at). It's just too much of a transformation. Like @epoberzkin says, this can happen with the other coercion types as well, but it's much less likely.

I guess I don't understand why you're fighting so hard on this? There are many different ways for you to get the functionality you want, and none of them have to involve changing this repo.

bighappyface:
@JanTvrdik @shmax is this PR still in play?

shmax:
@bighappyface My opinion is no, especially considering we haven't heard a peep from @JanTvrdik in quite a while.

bighappyface:
Closing until further notice from @JanTvrdik

vicary:
For historical reasons we did way too much of assumption to substitute "0" and "1" where a true boolean is not possible, say `$_GET` and for mysql columns requries indexing.

I'd vote for it if this is where we can start "coercing" things back to standard.

For the case of https://github.com/epoberezkin/ajv/issues/395#issuecomment-272743288, I am not sure any coerce should happen for `anyOf`, `oneOf` and `allOf` when the supplied value is valid in that array.

PR #347, opened 2017-01-16
Fix regex for Rfc3339 validation

kekko1212:
This is fixing a `date-time` validation problem where an invalid datetime string is still considered valid if it starts or ends with invalid chars (see the tests for a more clear example).


bighappyface:
+1

PR #349, opened 2017-01-17
Add option to apply default values from the schema

erayd:
For cases when the object being validated does not define a property, but that property has a default value in the schema, set the property value to the default. This PR has been rebased on PR #351.

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -147,7 +147,7 @@ class UndefinedConstraint extends Constraint
                         }
                     }
                 }
-            } elseif (($value instanceof UndefinedConstraint || $value === null) && isset($schema->default)) {
+            } elseif (($value instanceof self || $value === null) && isset($schema->default)) {
                 // $value is a leaf, not a container - apply the default directly
                 $value = is_object($schema->default) ? clone $schema->default : $schema->default;
             }
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -9,6 +9,7 @@
 
 namespace JsonSchema\Constraints;
 
+use JsonSchema\Constraints\TypeCheck\LooseTypeCheck;
 use JsonSchema\Entity\JsonPointer;
 use JsonSchema\Uri\UriResolver;
 
@@ -57,7 +58,9 @@ class UndefinedConstraint extends Constraint
         }
 
         // check object
-        if ($this->getTypeCheck()->isObject($value)) {
+        if (LooseTypeCheck::isObject($value)) { // object processing should always be run on assoc arrays,
+                                                // so use LooseTypeCheck here even if CHECK_MODE_TYPE_CAST
+                                                // is not set (i.e. don't use $this->getTypeCheck() here).
             $this->checkObject(
                 $value,
                 isset($schema->properties) ? $this->factory->getSchemaStorage()->resolveRefSchema($schema->properties) : $schema,
@@ -104,6 +107,37 @@ class UndefinedConstraint extends Constraint
                 }
             } else {
                 $this->checkUndefined($value, $schema->extends, $path, $i);
+            }
+        }
+
+        // Apply default values from schema
+        if ($this->factory->getConfig(self::CHECK_MODE_APPLY_DEFAULTS)) {
+            if ($this->getTypeCheck()->isObject($value) && isset($schema->properties)) {
+                // $value is an object, so apply default properties if defined
+                foreach ($schema->properties as $i => $propertyDefinition) {
+                    if (!$this->getTypeCheck()->propertyExists($value, $i) && isset($propertyDefinition->default)) {
+                        $this->getTypeCheck()->propertySet($value, $i, $propertyDefinition->default);
+                    }
+                }
+            } elseif ($this->getTypeCheck()->isArray($value)) {
+                if (isset($schema->properties)) {
+                    // $value is an array, but default properties are defined, so treat as assoc
+                    foreach ($schema->properties as $i => $propertyDefinition) {
+                        if (!isset($value[$i]) && isset($propertyDefinition->default)) {
+                            $value[$i] = $propertyDefinition->default;
+                        }
+                    }
+                } elseif (isset($schema->items)) {
+                    // $value is an array, and default items are defined - treat as plain array
+                    foreach ($schema->items as $i => $itemDefinition) {
+                        if (!isset($value[$i]) && isset($itemDefinition->default)) {
+                            $value[$i] = $itemDefinition->default;
+                        }
+                    }
+                }
+            } elseif (($value instanceof UndefinedConstraint || $value === null) && isset($schema->default)) {
+                // $value is a leaf, not a container - apply the default directly
+                $value = $schema->default;
             }
         }
 
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -116,7 +116,11 @@ class UndefinedConstraint extends Constraint
                 // $value is an object, so apply default properties if defined
                 foreach ($schema->properties as $i => $propertyDefinition) {
                     if (!$this->getTypeCheck()->propertyExists($value, $i) && isset($propertyDefinition->default)) {
-                        $this->getTypeCheck()->propertySet($value, $i, $propertyDefinition->default);
+                        if (is_object($propertyDefinition->default)) {
+                            $this->getTypeCheck()->propertySet($value, $i, clone $propertyDefinition->default);
+                        } else {
+                            $this->getTypeCheck()->propertySet($value, $i, $propertyDefinition->default);
+                        }
                     }
                 }
             } elseif ($this->getTypeCheck()->isArray($value)) {
@@ -124,20 +128,28 @@ class UndefinedConstraint extends Constraint
                     // $value is an array, but default properties are defined, so treat as assoc
                     foreach ($schema->properties as $i => $propertyDefinition) {
                         if (!isset($value[$i]) && isset($propertyDefinition->default)) {
-                            $value[$i] = $propertyDefinition->default;
+                            if (is_object($propertyDefinition->default)) {
+                                $value[$i] = clone $propertyDefinition->default;
+                            } else {
+                                $value[$i] = $propertyDefinition->default;
+                            }
                         }
                     }
                 } elseif (isset($schema->items)) {
                     // $value is an array, and default items are defined - treat as plain array
                     foreach ($schema->items as $i => $itemDefinition) {
                         if (!isset($value[$i]) && isset($itemDefinition->default)) {
-                            $value[$i] = $itemDefinition->default;
+                            if (is_object($itemDefinition->default)) {
+                                $value[$i] = clone $itemDefinition->default;
+                            } else {
+                                $value[$i] = $itemDefinition->default;
+                            }
                         }
                     }
                 }
             } elseif (($value instanceof UndefinedConstraint || $value === null) && isset($schema->default)) {
                 // $value is a leaf, not a container - apply the default directly
-                $value = $schema->default;
+                $value = is_object($schema->default) ? clone $schema->default : $schema->default;
             }
         }
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,40 @@ $validator->coerce($request, $schema);
 // equivalent to $validator->validate($data, $schema, Constraint::CHECK_MODE_COERCE_TYPES);
 ```
 
+### Default values
+
+If your schema contains default values, you can have these automatically applied during validation:
+
+```php
+<?php
+
+use JsonSchema\Validator;
+use JsonSchema\Constraints\Constraint;
+
+$request = (object)[
+    'refundAmount'=>17
+];
+
+$validator = new Validator();
+
+$validator->validate(
+    $request,
+    (object)[
+        "type"=>"object",
+        "properties"=>(object)[
+            "processRefund"=>(object)[
+                "type"=>"boolean",
+                "default"=>true
+            ]
+        ]
+    ],
+    Constraint::CHECK_MODE_APPLY_DEFAULTS
+); //validates, and sets defaults for missing properties
+
+is_bool($request->processRefund); // true
+$request->processRefund; // true
+```
+
 ### With inline references
 
 ```php
@@ -152,9 +186,11 @@ third argument to `Validator::validate()`, or can be provided as the third argum
 | `Constraint::CHECK_MODE_NORMAL` | Validate in 'normal' mode - this is the default |
 | `Constraint::CHECK_MODE_TYPE_CAST` | Enable fuzzy type checking for associative arrays and objects |
 | `Constraint::CHECK_MODE_COERCE_TYPES` | Convert data types to match the schema where possible |
+| `Constraint::CHECK_MODE_APPLY_DEFAULTS` | Apply default values from the schema if not set |
 | `Constraint::CHECK_MODE_EXCEPTIONS` | Throw an exception immediately if validation fails |
 
-Please note that using `Constraint::CHECK_MODE_COERCE_TYPES` will modify your original data.
+Please note that using `Constraint::CHECK_MODE_COERCE_TYPES` or `Constraint::CHECK_MODE_APPLY_DEFAULTS`
+will modify your original data.
 
 ## Running the tests
 
--- a/src/JsonSchema/Constraints/Factory.php
+++ b/src/JsonSchema/Constraints/Factory.php
@@ -15,6 +15,7 @@ use JsonSchema\SchemaStorage;
 use JsonSchema\SchemaStorageInterface;
 use JsonSchema\Uri\UriRetriever;
 use JsonSchema\UriRetrieverInterface;
+use JsonSchema\Constraints\Constraint;
 
 /**
  * Factory for centralize constraint initialization.
--- a/src/JsonSchema/Constraints/Factory.php
+++ b/src/JsonSchema/Constraints/Factory.php
@@ -9,13 +9,13 @@
 
 namespace JsonSchema\Constraints;
 
+use JsonSchema\Constraints\Constraint;
 use JsonSchema\Exception\InvalidArgumentException;
 use JsonSchema\Exception\InvalidConfigException;
 use JsonSchema\SchemaStorage;
 use JsonSchema\SchemaStorageInterface;
 use JsonSchema\Uri\UriRetriever;
 use JsonSchema\UriRetrieverInterface;
-use JsonSchema\Constraints\Constraint;
 
 /**
  * Factory for centralize constraint initialization.
--- a/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
+++ b/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
@@ -36,7 +36,6 @@ class LooseTypeCheck implements TypeCheckInterface
         }
     }
 
-
     public static function propertyExists($value, $property)
     {
         if (is_object($value)) {
--- a/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
+++ b/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
@@ -27,6 +27,16 @@ class LooseTypeCheck implements TypeCheckInterface
         return $value[$property];
     }
 
+    public static function propertySet(&$value, $property, $data)
+    {
+        if (is_object($value)) {
+            $value->{$property} = $data;
+        } else {
+            $value[$property] = $data;
+        }
+    }
+
+
     public static function propertyExists($value, $property)
     {
         if (is_object($value)) {
--- a/src/JsonSchema/Constraints/TypeCheck/StrictTypeCheck.php
+++ b/src/JsonSchema/Constraints/TypeCheck/StrictTypeCheck.php
@@ -19,6 +19,11 @@ class StrictTypeCheck implements TypeCheckInterface
         return $value->{$property};
     }
 
+    public static function propertySet(&$value, $property, $data)
+    {
+        $value->{$property} = $data;
+    }
+
     public static function propertyExists($value, $property)
     {
         return property_exists($value, $property);
--- a/src/JsonSchema/Constraints/TypeCheck/TypeCheckInterface.php
+++ b/src/JsonSchema/Constraints/TypeCheck/TypeCheckInterface.php
@@ -10,6 +10,8 @@ interface TypeCheckInterface
 
     public static function propertyGet($value, $property);
 
+    public static function propertySet(&$value, $property, $data);
+
     public static function propertyExists($value, $property);
 
     public static function propertyCount($value);
--- a/tests/Constraints/DefaultPropertiesTest.php
+++ b/tests/Constraints/DefaultPropertiesTest.php
@@ -34,11 +34,6 @@ class DefaultPropertiesTest extends VeryBaseTestCase
                 '{"properties":{"propertyTwo":{"default":"valueTwo"}}}',
                 '{"propertyOne":"valueOne","propertyTwo":"valueTwo"}'
             ),
-            array(// fulfil required property with a default value
-                '{"propertyOne":"valueOne"}',
-                '{"properties":{"propertyTwo":{"required":true,"default":"valueTwo"}}}',
-                '{"propertyOne":"valueOne","propertyTwo":"valueTwo"}'
-            ),
             array(// default value for sub-property
                 '{"propertyOne":{}}',
                 '{"properties":{"propertyOne":{"properties":{"propertyTwo":{"default":"valueTwo"}}}}}',
@@ -68,11 +63,6 @@ class DefaultPropertiesTest extends VeryBaseTestCase
                 '{"propertyOne":"alreadySetValueOne"}',
                 '{"properties":{"propertyOne":{"default":"valueOne"}}}',
                 '{"propertyOne":"alreadySetValueOne"}'
-            ),
-            array(//default value is required
-                '{"propertyOne":"valueOne"}',
-                '{"properties":{"propertyTwo":{"default":"valueTwo","required":true}}}',
-                '{"propertyOne":"valueOne","propertyTwo":"valueTwo"}'
             ),
             array(//default item value for an array
                 '["valueOne"]',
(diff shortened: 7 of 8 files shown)


src/JsonSchema/Validator.php
```diff
@@ -45,10 +45,10 @@ public function check($value, $schema = null, JsonPointer $path = null, $i = nul
      *
      * {@inheritDoc}
      */
-    public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = null)
+    public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = null, $default = false)
```
shmax: Rather than trail another argument on the end of `check` and `coerce`, you might consider adding a config option to drive this. Something like:

```PHP
abstract class Constraint implements ConstraintInterface
{
//...
    const CHECK_MODE_NORMAL = 0x00000001;
    const CHECK_MODE_TYPE_CAST = 0x00000002;
    const CHECK_MODE_USE_DEFAULTS = 0x00000004;
```

That way you would only really have to change the code in one place and not have to chase down all the different paths like you're currently doing (and wonder if you missed any).

src/JsonSchema/Constraints/ObjectConstraint.php
```diff
@@ -170,6 +175,26 @@ protected function &getProperty(&$element, $property, $fallback = null)
     }
 
     /**
+     * set and return a property on an object or array
+     *
+     * @param mixed $element Element to manipulate
+     * @param string $property  Property to set
+     * @param mixed $value Value to set
+     */
+    protected function &setProperty(&$element, $property, $value)
```
shmax: Why is it necessary to return a reference? Why is it necessary to return anything at all? `$value` is passed in, so presumably whoever is calling this method already has what it needs.

src/JsonSchema/Validator.php
erayd: OK - will change to use config, add unit tests, and update the readme :-).

Re additional use-cases, was there something specific you are referring to that I've missed? My understanding from reading the spec is that the default can be applied directly, and *should* be schema-compliant, but isn't mandated to be so.

src/JsonSchema/Constraints/ObjectConstraint.php
erayd: To avoid an extra call to getProperty, and for consistency with that method. The reference is to the newly set property, not to $value (unless $element isn't an object or an array, in which case it behaves the same as $fallback from getProperty).

Would you rather I implemented it without the reference return? The way I've done it feels tidier, to me, but quite happy to change it if you don't like that approach.

src/JsonSchema/Validator.php
shmax: I was going to wait until we had some unit tests to look at, but you'll want to make sure that setting a default on `items` works, for example.

src/JsonSchema/Constraints/ObjectConstraint.php
shmax: I guess there's no law against it, but having a setter return something just feels like an unnecessary mixing of concerns to me. Throwing in the reference is just extra squirm factor. You're only calling this method in one place, and though it's late and my head is fuzzy I really can't see why it's necessary or how its inclusion makes the code any simpler.

```
if ($default && isset($definition->default) && $property === $undefinedConstraint) {
    $property = $definition->default;
    $this->setProperty($element, $i, $property);
}
```

shmax: Also--and I may be able to come up with something more helpful tomorrow--I remember writing a block of similar object-vs-array set code when I was doing type coercion, but I eventually managed to find a point in the code where the assignment could be done a little more naturally in the inner workings, and it wasn't necessary to shoehorn it in like this. I'm off to bed, but I'll see if I can remember what I'm talking about tomorrow.

erayd:
OK, how's that?

What are your thoughts on adding helper functions in the Validator class (e.g. checkDefault and coerceDefault)? Currently, the only way that I can see to run this with the config option is to pass a Factory object to the Validator constructor.

Haven't done the unit tests yet - those are next on my list.

shmax:
> Currently, the only way that I can see to run this with the config option is to pass a Factory object to the Validator constructor

That is a bit awkward. I have a PR where I move check mode out of Factory to make this kind of thing easier to do, but I abandoned it when my coercion code evolved into not needing a flag. If you're interested we can revive it... 

https://github.com/justinrainbow/json-schema/pull/320

shmax:
Try this schema and value:

```PHP
$value = json_decode('[ 1 ]');
$schema = json_decode('{
    "type": "array",
    "items": [
        { "type": "number" },
	    { "type": "string", "default": "foo" }
    ]
}');
```

shmax:
When I use this schema and value...

```
$value = json_decode('{ "foo": 1 }');
$schema = json_decode('{
    "type": "object",
    "properties": {
        "foo": { "type": "number" },
	"bar": { "type": "string", "default": "baz" }
    },
    "required": [ "foo", "bar" ]
}');
```

The default value is applied, but validation fails ("The property bar is required")

shmax:
Also, if you try the above sample but use `coerce` instead of `check`, it all blows up.

shmax:
> OK, how's that?

I don't know. I'm looking at the code more closely, now, and I see that the reference was needed because it gets forwarded to `checkUndefined` for further validation, so I guess my original review was a little shaky. Sorry about that. But I'm wondering if you're filling-in your defaults too early in the process. I have to head off to work, but I'll poke around some more after work...

erayd:
Looks like there's some kind of issue with documents that are passed as an array too - discovered that when I was writing unit tests last night. I'm trying to track down what's going on there at the moment.

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -66,7 +66,8 @@ public function validateTypes(&$value, $schema = null, JsonPointer $path, $i = n
         }
 
         // check object
```
erayd: For some reason, most assoc tests pass, but two of mine fail when strict type checking is used. From what I can see, it looks like properties which contain children simply aren't being parsed, which doesn't seem right...

Currently trying to find the root cause of this - not sure whether the problem is in my code or in master, but starting with the assumption that the problem is mine.

erayd:
Have updated the PR again. Currently working on adding more tests, finding the root cause of an issue that fails a couple of the tests in assoc mode, and making default options for array items work.

shmax:
Sounds good. I'll let you keep plugging away. Let me know when it's really ready for another look.

erayd:
Will do. I'll keep the PR updated so you can see where I'm up to, and will let you know when I'm ready for you to take another look :-).

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -66,7 +66,8 @@ public function validateTypes(&$value, $schema = null, JsonPointer $path, $i = n
...
-        if ($this->getTypeCheck()->isObject($value)) {
+        if (TypeCheck\LooseTypeCheck::isObject($value)) { // Fixes failing assoc tests for default values - currently investigating
+        //if ($this->getTypeCheck()->isObject($value)) {  // to find the root cause of this, noting all other assoc tests pass.
```
erayd: Strict type-checking here causes two of my tests to fail, but only in assoc mode (both tests relate to setting default values on properties in a child object). It's not immediately obvious what the purpose of a strict check here is - are you able to shed some light on why this is necessary, and why it's not OK to always run $this->checkObject() (via $this->validateTypes) on an associative array?

Changing to a loose check on this line (i.e. allowing associative arrays to be considered objects too) resolves the issue, and allows all tests to succeed, but I don't want to do this if it's going to break something, either now or in the future. If you could provide some insight into the reason behind this restriction, it would be extremely helpful.

shmax: The loose checking stuff is from before my time, but I can dig deeper after work.

erayd: Thanks :-). I don't suppose anyone involved with the addition of that code is still around? Looks like it was done midway through last year.

shmax: Sorry for the delay.

I've stepped through your code, and the problem is that you are deferring to `testValidCases` from `testValidCasesUsingAssoc`, and you're passing along an array instead of an object.  If you look in BaseTestCase::testValidCasesUsingAssoc, you'll see that it prepares  `Validator`  by passing in `$checkMode`, and the default value of `$checkMode` is `CHECK_MODE_TYPE_CAST`, which will have no problem with associative arrays.

I see that you extended `VeryBaseTestCase` instead of `BaseTestCase`; presumably you did that because you wanted a different signature for `testValidCasesUsingAssoc` and `testValidCases`. That's okay, but make sure that you flesh out `testValidCasesUsingAssoc` with the proper boilerplate.

src/JsonSchema/Validator.php
```diff
@@ -52,4 +52,17 @@ public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = n
...
     }
+
+    /**
+     * Does everything that coerce does, but will also set values to their default, if the value is not
+     * set and a default is available in the schema. Note that the first argumen is passwd by
+     * reference, so you must pass in a variable.
+     *
+     * {@inheritDoc}
+     */
+    public function coerceDefault(&$value, $schema = null, JsonPointer $path = null, $i = null)
```
shmax: Really not a fan of this kind of helper, for a couple of reasons:

1. You're jamming together two concepts, coercion and default'ing. What happens if I want to coerce, but not default (call `coerce`, okay)? What happens if I want to default, but not coerce (well, we don't have that one yet). What's going to happen when a third thing comes along? Do we just keep fragmenting this into more and more helpers, one for each possible combination? 
2. It has the feel of a fire-and-forget method, but it secretly twiddles a flag and leaves it there.

Can we please just leave this as a config option, for now? I know it's awkward, but again, once we survive this PR I promise to resurrect my other PR which simplifies the API.

src/JsonSchema/Constraints/UndefinedConstraint.php
erayd: Thanks :-).

Unfortunately, turning on `CHECK_MODE_TYPE_CAST` results in pretty much all the assoc tests breaking, and doesn't fix the issue. Am I perhaps misunderstanding what you're suggesting here?

```php
    /**
     * @dataProvider getValidTests
     */
    public function testValidCasesUsingAssoc($input, $schema, $expectOutput = null)
    {
        if (is_string($input)) {
            $inputDecoded = json_decode($input, true);
        } else {
            $inputDecoded = $input;
        }

        $validator = new Validator(new Factory(null, null, Constraint::CHECK_MODE_TYPE_CAST));
        $validator->coerceDefault($inputDecoded, json_decode($schema));

        $this->assertTrue($validator->isValid(), print_r($validator->getErrors(), true));

        if ($expectOutput !== null) {
            $this->assertEquals($expectOutput, json_encode($inputDecoded));
        }
    }
```
Using the above for assoc tests, it fails for datasets 1, 2, 3, 4, 5, 6, 7, 8 & 10, regardless of whether or not I'm using loose type checking in UndefinedConstraint::validateTypes(). Using what I originally had, 4 & 5 are the only tests that fail. Using what I originally had plus loose type checking, everything passes.

src/JsonSchema/Validator.php
erayd: Sure. It's a little awkward, but it sounds like you have a plan to resolve that in your other PR.

However... if it's a config option, then how would you like me to handle checking without coercion? The various check() methods currently don't require a reference, which means that defaulting breaks if $value is an array (because PHP doesn't pass arrays by reference unless the reference is explicit).

This is the reason I wrapped it in coerceDefault() in the first place - to prevent people trying to use it with check(), and then wondering why it didn't work. In an ideal world, I could simply alter the many check() methods to take a reference also... but is that something you're OK with?

src/JsonSchema/Constraints/UndefinedConstraint.php
shmax: Yup. Now you're hitting the next problem, which is that in `LooseTypeCheck::propertySet` you are doing a write to $value, which is passed by value, so the write doesn't stick. You need to either pass by reference, or pass in an object. It was working in your old code before because when you derferred to `testValidCases`, that was converting to object. 

Hope that all makes sense. I'm off to bed, but I can help more tomorrow morning if you leave some comments. G'nite!

erayd: Oops - that is definitely a mistake on my part, sorry for wasting your time with that one! Not sure how I missed it, but thanks for finding it.

All sorted now, and no need for the loose type checking any more :-).

src/JsonSchema/Validator.php
shmax: Yeah, that's a good point. Tell you what--as long as we're going to change the API, let's go all the way and just get rid of `coerce`, and convert that (back) to a config flag, as well. Then we'll modify `check` to just take a reference everywhere, and we'll be all set. I didn't do that on my second coercion attempt because I didn't want to break compatibility, in other words, this kind of thing would break:

```PHP
$validator->check(['foo'=>'bar'], $schema);
```
But now that I think about it, that's a pretty outlandish use case; I mean, I literally can't think of any rational reason to be doing it, outside of writing sample code for readmes and whatnot.

What do you think? That will should simplify things for us both. If you're game you can make the change in this PR, or I can do it when I move `$checkmode` back up into `Constraint` (which you're also welcome to do if you're feeling ambitious).

shmax:
Okay, now I'm REALLY off to bed. Just one parting note: you don't have to keep squashing. When you do that I can't view the latest changes, and the narrative is lost. You can squash at the very end.

src/JsonSchema/Validator.php
erayd: I like that idea a lot, and yes, quite happy to make the change. Do you want it in this PR, or in its own PR (and then I can rebase this after that one is merged)?

I don't see that use-case you mention above as a problem - anyone who needs to do something like that will just assign it to a variable and then check the variable - but it does raise the question of whether it's OK to break this functionality 'in the wild'.

I do think that Validator::coerce() should stay, if only as a flag-setting alias, to prevent breaking things for anyone who is currently using it.

erayd:
OK - goodnight, and sleep well :-).

Noted re the squashing, I'll leave the commit history there 'til the end.

src/JsonSchema/Validator.php
erayd: What are your thoughts on PR #351?

In hindsight, I think the ability to check-by-value needs to stay, as people will be doing things like `check(json_decode($someJSON), $schema)`, and it's important that we don't break that.

shmax: I'm really digging it! Will have more feedback after work.

erayd:
Have rebased this on PR #351.

erayd:
Now rebased on master following merge of PR #351.

erayd:
This will need rebasing again after PR #354 is merged, plus another update to README.md (adding `Constraint::CHECK_MODE_APPLY_DEFAULTS` to the table).

erayd:
OK, have rebased this on master.

@bighappyface / @shmax - anything else you think I need to do to this before it's mergeable?

shmax:
> anything else you think I need to do to this before it's mergeable?

Ah, so we finally worked our way back to this PR, eh? Good stuff. I'll take a fresh look at it tomorrow...

bighappyface:
I like the ability to push defaults into candidate data; however, modifying the candidate data as part of validation really isn't validation, is it? Coercion already sets the precedent of modifying the candidate data when the flag is set, so I can't push back too hard, but it would be nice if I could fill in defaults for data separately without it being a flagged version of validating that data.

Does that make sense? Separation of concerns? Is `validate` doing too much?

shmax:
> if I could fill in defaults for data separately without it being a flagged version of validating that data.

That would mean a second pass, wouldn't it?

shmax:
Just a few thoughts before I head off to work... I don't really see a major problem with this idea. The alternative would be to make a two passes on the data, which I see as a non-starter--the traversal overhead is already quite large. The precedent for doing coercion and default setting in the same pass as validation has been set in various other projects (eg. https://github.com/epoberezkin/ajv#assigning-defaults). Would you have been more comfortable back when the main entry point was still called `check`?

erayd:
@bighappyface @shmax I had the same thoughts regarding having the set-defaults thing as part of the validation logic, but I cannot in all honesty think of anywhere else to put it that makes sense.

 * Putting it anywhere else would necessitate two parsing runs, which isn't good for performance.
 * Validating the defaults is a good idea, which this PR does. If it's a separate step, then validating the defaults becomes more complex.

Regarding whether having a function called `validate()` modify the input data is a good idea - the way I think about it is that it's not just checking for validation, but also (optionally) enforcing validity. Both type coercion and defaults fall within the scope of enforced validity, and neither are default behavior - you have to explicitly turn them on before anything is modified, so it's not like the user won't be aware what's happening.

bighappyface:
It still feels funny to me. Transforming data is not validating data, and transforming data to make it valid is definitely not validation; however, again, the functionality is behind a flag, and things have already gone this way.

erayd:
@bighappyface I'm certainly open to changing things if you have a suggestion for where you'd like it - I just can't think of anywhere else to put it that makes sense!

bighappyface:
I think we are good here. I am really only waiting for #355 before I merge this in.

mathroc:
@bighappyface about modifying input data, how about this ?

```php
$validator = new Validator();
$validator->validate(
  $data, 
  $schema, 
  Constraint::CHECK_MODE_APPLY_DEFAULTS
);
if (!$this->validator->isValid()) {
  throw ...;
}
$data = $validator->validatedData();
```

the validator would have to make a copy of the input data when working with read/write flags and store it for `validatedData()`

could also be another flag (to avoid BC break) : `$validator->validate($data, $schema, Constraint::CHECK_MODE_APPLY_DEFAULTS | Constraint::COPY_INPUT);`

bighappyface:
@mathroc I am comfortable with your idea. @erayd @shmax thoughts?

bighappyface:
@erayd please rebase

erayd:
@bighappyface Have rebased.

@mathroc @bighappyface @shmax I'm also OK with the idea of optionally copying the data. Rather than have a whole separate method for it though, I've just made `validate()` return either the original or the copy, depending on whether `CHECK_MODE_COPY_INPUT` is set. Thoughts on this implementation of the concept?

shmax:
@bighappyface @mathroc Not a fan of the idea; it complicates the API, and weakens it at the same time. The implication is that we don't trust ourselves to not make any unintended edits to the incoming data, but that's what we have unit tests for. If a consumer of this library REALLY wants to make sure that he doesn't lose his original input for some reason AND wants to coerce or apply defaults, he can certainly do so with no help from us:

```PHP
$validator = new Validator();
$originalData = clone $data;
$validator->validate(
  $data, 
  $schema, 
  Constraint::CHECK_MODE_APPLY_DEFAULTS
);
if (!$this->validator->isValid()) {
  throw ...;
}

```
Now the onus is on him to expend the extra memory or not, and our API remains simple and intuitive for what I expect are the majority of people that aren't interested in this use case.

erayd:
Good points @shmax, I've changed my mind and now agree with you. Have removed the CHECK_MODE_COPY_INPUT commits again.

bighappyface:
Good stuff on 6c231a61f376721d145cabf6accdf426bd3e7213

bighappyface:
@all bump to 5.1.0 with this?

shmax:
> @all bump to 5.1.0 with this?

Sounds good to me, unless anyone is interested in revisiting the idea of moving $checkMode up out of Factory?

erayd:
@shmax Must say I'm a fan of leaving it where it is (thought we settled that back in #351)... but if you do still want it to move at some point, can that please happen in another PR after this one is merged?

shmax:
@erayd Oh, it can certainly go in a different PR, but if we're going to do it then we should wait to bump the release. Then again, it would break compatibility, so perhaps it would warrant major version bump. But just to review, the idea is only that there is no need to bury `$checkMode` deep in `Factory` when it could just as easily live in `Constraint`. There would be no major functional difference, but it would make configuring an internal $checkMode easier for people who otherwise have no interest in overriding any of the machinery in `Factory`. Thoughts?

erayd:
`Factory` is instantiated, `Constraint` is not. Putting it in `Constraint` would require making it a static variable, which implies global state. Global state is usually best avoided IMO unless there's a very good reason for it, as it can cause weird bugs.

erayd:
@shmax `Factory` is instantiated, `Constraint` is not. Putting it in `Constraint` would require making it a static variable, which implies global state. Global state is usually best avoided IMO unless there's a very good reason for it, as it can cause weird bugs.

shmax:
`Constraint` is the base class for most of the classes in the project. It is definitely an instance (unless something happened that I missed), and is where `$checkMode` used to live. Have a look here:

https://github.com/justinrainbow/json-schema/commit/31b29546f49524cb0123fb5619107a9306444b60#diff-e7c19f3eac4c5373538af0e003cbb7c0L52

@digitalkaoz got a little carried away in [this PR](https://github.com/justinrainbow/json-schema/pull/310), and $checkMode wound up getting swept along in his changes to the other components ($schemaStorage and such), even though it didn't suffer from the same performance overhead as the other. 

I had a PR to make this change at one point... let me see if I can dig it up... ah, here we go:
https://github.com/justinrainbow/json-schema/pull/320

As you can see I closed it myself once the coerce flag was removed, but now that it's back I figure we can take another look at the idea.

I'm perfectly willing to wait for 6.0.

erayd:
@shmax `Constraint` is an abstract class, even in the diff you linked to. There is no possible way it could be instantiated.

What you're looking at is the default constructor, which if used to set `$checkMode` as per the red side of that commit (e.g. via calling `parent::__construct()` in a subclass), would result in an individual $checkMode` for *every* instance of a `Constraint` subclass.

shmax:
> @shmax Constraint is an abstract class

Yep, and abstract classes can have members, same as anything else. Just dig back in the history for `Constraint` and you'll see that it used to host all the stuff that is now in `Factory`.

shmax:
Here's a snapshot of the file from just before @digitalkaoz gutted it: https://github.com/justinrainbow/json-schema/blob/a918d3b5d9fb3eceb44826343bfe735bcb5c9292/src/JsonSchema/Constraints/Constraint.php

You can see that it has a variety of members, and is still abstract.

shmax:
Oh, you might be getting confused because `$factory` has moved from `Constraint` down to `BaseConstraint`, which might be where `$checkMode` would wind up.

shmax:
> is an abstract class, even in the diff you linked to. There is no possible way it could be instantiated.

Sure there is! Simply subclass it, implement the interface, and instantiate.

erayd:
@shmax Definitely not confused. I'm sorry, but you're really, really wrong on this one. See [here](http://php.net/manual/en/language.oop5.abstract.php):

> Classes defined as abstract may not be instantiated...

Abstract classes may _define_ members, which are inherited by their subclasses, but those members don't exist on the abstract class itself - they exist on subclass instances only. 

If you instantiate a subclass, the inherited member has its own value for that subclass instance; it's not shared by all subclass instances.

erayd:
Here's a quick example:
```PHP
<?php

abstract class ParentClass {
    public $myField = 0;
}

class ChildClassOne extends ParentClass {
    public function __construct()
    {
        $this->myField = 1;
    }
}

class ChildClassTwo extends ParentClass {
    public function getMyField()
    {
        return $this->myField;
    }
}

$one = new ChildClassOne();
$two = new ChildClassTwo();

assert($two->getMyField() === 0); //true
```

shmax:
I never argued that there wouldn't be multiple copies of the value. There would be multiple copies, and in fact there are even now, as `Factory` feeds a copy of itself to whatever it instantiates. So even after my change there would be the same number of `$checkModes` floating around, they would just be moved up outside of `Factory`.

erayd:
Gotcha - misunderstood your original argument. However... `Factory` does *not* feed a copy of itself to whatever it instantiates; it's an object - it passes `$this` by reference. Unless you're deliberately making more for some reason, there's only ever one instance of `Factory`.

It sounds like you are arguing in favour of going from one instance of `$checkMode` in a central location, to many instances passed all over the stack the way `$coerce` used to be? I don't see how that is in any way an improvement. If I've misunderstood, then can you correct me please?

shmax:
Ah, no, not arguing in favor of passing it all over the stack. It would just go back to being a member on `Constraint`. But you have a good point about `Factory` and references. We could probably still simplify the interface (such that checkmode is passed in as a top level argument when creating `Validator` then passed to `Factory`), but we can revisit it later. As I said, let's worry about it in 6.0.

erayd:
OK. I have no objection in principle to simplifying things.

Shall we leave this discussion for now, get this PR merged, and revisit making changes to `$checkMode` in a later PR?

shmax:
Sure, sounds good. By the way, did you ever try setting a default value for `items`?

erayd:
Yep - see array test cases in `DefaultPropertiesTest::getValidTests()`.

erayd:
Oh, hang on - do you mean defaulting all of items, rather than setting individual default items?

shmax:
Mm, the latter. Didn't consider the former. Not too worried about it, at least for now.

erayd:
If I'm understanding your question correctly, the test case labelled _'default item value for an empty array'_ addresses setting all of items, and the one labelled _'default item value for an array'_ addresses setting individual items.

If there's a case I've missed, please let me know, and I'll add tests for it.

shmax:
Nope, looks great. :shipit:

erayd:
Sweet :grinning:

erayd:
Thanks for asking a question that made me look at the tests again though - spotted a couple of duplicates in there :-).

bighappyface:
5.1.0 it is

bighappyface:
https://github.com/justinrainbow/json-schema/releases/tag/5.1.0

erayd:
Yay! Thanks @bighappyface :-).

mathroc:
> @bighappyface @mathroc Not a fan of the idea; it complicates the API, and weakens it at the same time. The implication is that we don't trust ourselves to not make any unintended edits to the incoming data, but that's what we have unit tests for. If a consumer of this library REALLY wants to make sure that he doesn't lose his original input for some reason AND wants to coerce or apply defaults, he can certainly do so with no help from us

@shmax absolutely cloning the data before handing it to the validator is easy enough

thx all for this PR!

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -107,6 +110,49 @@ protected function validateCommonProperties(&$value, $schema = null, JsonPointer
...
+                    if (!$this->getTypeCheck()->propertyExists($value, $i) && isset($propertyDefinition->default)) {
+                        if (is_object($propertyDefinition->default)) {
+                            $this->getTypeCheck()->propertySet($value, $i, clone $propertyDefinition->default);
+                        } else {
+                            $this->getTypeCheck()->propertySet($value, $i, $propertyDefinition->default);
+                        }
+                    }
+                }
+            } elseif ($this->getTypeCheck()->isArray($value)) {
+                if (isset($schema->properties)) {
```
mathroc: I think this could be refactored with https://github.com/justinrainbow/json-schema/pull/349/files#diff-9aeb900077027f2ca2e0a03af4ea6a2fR117 by using the `LooseTypeCheck` instead of `$this->getTypeCheck()`

PR #350, opened 2017-01-18
Fixed regex for Content-Type header parsing

hrasekj:
Server can return Content-Type header with charset property. That throws exception when JsonSchema\Uri\UriRetriever::confirmMediaType is called.


bighappyface:
@hrasekj thanks for this contribution. Would you please rebase this branch on master and look at supplementing this update with unit testing?

bighappyface:
Closing due to inactivity.

PR #351, opened 2017-01-19
Add new config option, refactor coerce feature to use this, and simplify the public API

erayd:
## Why
 * The current API isn't particularly good at exposing new functionality to users...
 * ...and I want to add new functionality (PR #349 and PR #354) and expose it in a sane way.
 * Passing `$coerce` all over the stack is risky - it's easy to lose track of it when making other changes.
 * The current design mandates exposing internal method arguments (`$path` and `$i`) to users.

## Changes
 * Split the Constraint class to make Validator independent from it
 * Add Validator::validate() as the main entry point
 * Turn Validator::coerce() and Validator::check() into aliases
 * Add Factory::setConfig(), getConfig(), addConfig() & removeConfig()
 * Make type-coercion a checkMode option, don't pass $coerce everywhere
 * Add some extra tests

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -20,81 +20,15 @@ use JsonSchema\Entity\JsonPointer;
  * @author Robert Schönthal <seroscho@googlemail.com>
  * @author Bruno Prieto Reis <bruno.p.reis@gmail.com>
  */
-abstract class Constraint implements ConstraintInterface
+abstract class Constraint extends BaseConstraint implements ConstraintInterface
 {
-    protected $errors = array();
     protected $inlineSchemaProperty = '$schema';
 
-    const CHECK_MODE_NORMAL = 		0x00000001;
-    const CHECK_MODE_TYPE_CAST = 	0x00000002;
-
-    /**
-     * @var Factory
-     */
-    protected $factory;
-
-    /**
-     * @param Factory $factory
-     */
-    public function __construct(Factory $factory = null)
-    {
-        $this->factory = $factory ? : new Factory();
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function addError(JsonPointer $path = null, $message, $constraint='', array $more=null)
-    {
-        $error = array(
-            'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
-            'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
-            'message' => $message,
-            'constraint' => $constraint,
-        );
-
-        if (is_array($more) && count($more) > 0)
-        {
-            $error += $more;
-        }
-
-        $this->errors[] = $error;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function addErrors(array $errors)
-    {
-        if ($errors) {
-            $this->errors = array_merge($this->errors, $errors);
-        }
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function getErrors()
-    {
-        return $this->errors;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function isValid()
-    {
-        return !$this->getErrors();
-    }
-
-    /**
-     * Clears any reported errors.  Should be used between
-     * multiple validation checks.
-     */
-    public function reset()
-    {
-        $this->errors = array();
-    }
+    const CHECK_MODE_NONE =             0x00000000;
+    const CHECK_MODE_NORMAL =           0x00000001;
+    const CHECK_MODE_TYPE_CAST =        0x00000002;
+    const CHECK_MODE_COERCE_TYPES =     0x00000004;
+    const CHECK_MODE_APPLY_DEFAULTS =   0x00000008;
 
     /**
      * Bubble down the path
@@ -123,16 +57,11 @@ abstract class Constraint implements ConstraintInterface
      * @param mixed            $schema
      * @param JsonPointer|null $path
      * @param mixed            $i
-	 * @param boolean          $coerce
      */
-    protected function checkArray(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
+    protected function checkArray(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
         $validator = $this->factory->createInstanceFor('collection');
-        if($coerce) {
-            $validator->coerce($value, $schema, $path, $i);
-        } else {
-            $validator->check($value, $schema, $path, $i);
-        }
+        $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
     }
@@ -145,16 +74,11 @@ abstract class Constraint implements ConstraintInterface
      * @param JsonPointer|null $path
      * @param mixed            $i
      * @param mixed            $patternProperties
-	 * @param boolean          $coerce
      */
-    protected function checkObject(&$value, $schema = null, JsonPointer $path = null, $i = null, $patternProperties = null, $coerce = false)
+    protected function checkObject(&$value, $schema = null, JsonPointer $path = null, $i = null, $patternProperties = null)
     {
         $validator = $this->factory->createInstanceFor('object');
-        if($coerce){
-            $validator->coerce($value, $schema, $path, $i, $patternProperties);
-        } else {
-            $validator->check($value, $schema, $path, $i, $patternProperties);
-        }
+        $validator->check($value, $schema, $path, $i, $patternProperties);
 
         $this->addErrors($validator->getErrors());
     }
@@ -166,16 +90,11 @@ abstract class Constraint implements ConstraintInterface
      * @param mixed            $schema
      * @param JsonPointer|null $path
      * @param mixed            $i
-	 * @param boolean          $coerce
      */
-    protected function checkType(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
+    protected function checkType(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
         $validator = $this->factory->createInstanceFor('type');
-        if($coerce) {
-            $validator->coerce($value, $schema, $path, $i);
-        } else {
-            $validator->check($value, $schema, $path, $i);
-        }
+        $validator->check($value, $schema, $path, $i);
 
         $this->addErrors($validator->getErrors());
     }
@@ -187,17 +106,12 @@ abstract class Constraint implements ConstraintInterface
      * @param mixed            $schema
      * @param JsonPointer|null $path
      * @param mixed            $i
-	 * @param boolean          $coerce
      */
-    protected function checkUndefined(&$value, $schema = null, JsonPointer $path = null, $i = null, $coerce = false)
+    protected function checkUndefined(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
         $validator = $this->factory->createInstanceFor('undefined');
 
-        if($coerce){
-            $validator->coerce($value, $this->factory->getSchemaStorage()->resolveRefSchema($schema), $path, $i);
-        } else {
-            $validator->check($value, $this->factory->getSchemaStorage()->resolveRefSchema($schema), $path, $i);
-        }
+        $validator->check($value, $this->factory->getSchemaStorage()->resolveRefSchema($schema), $path, $i);
 
         $this->addErrors($validator->getErrors());
     }
--- a/src/JsonSchema/Constraints/ConstraintInterface.php
+++ b/src/JsonSchema/Constraints/ConstraintInterface.php
@@ -40,7 +40,7 @@ interface ConstraintInterface
      * @param string           $constraint the constraint/rule that is broken, e.g.: 'minLength'
      * @param array            $more more array elements to add to the error
      */
-    public function addError(JsonPointer $path = null, $message, $constraint='', array $more=null);
+    public function addError(JsonPointer $path = null, $message, $constraint='', array $more = null);
 
     /**
      * checks if the validator has not raised errors
@@ -59,5 +59,5 @@ interface ConstraintInterface
      * @param mixed            $i
      * @throws \JsonSchema\Exception\ExceptionInterface
      */
-    public function check($value, $schema = null, JsonPointer $path = null, $i = null);
+    public function check(&$value, $schema = null, JsonPointer $path = null, $i = null);
 }
--- a/src/JsonSchema/Constraints/Factory.php
+++ b/src/JsonSchema/Constraints/Factory.php
@@ -10,6 +10,7 @@
 namespace JsonSchema\Constraints;
 
 use JsonSchema\Exception\InvalidArgumentException;
+use JsonSchema\Exception\InvalidConfigException;
 use JsonSchema\SchemaStorage;
 use JsonSchema\SchemaStorageInterface;
 use JsonSchema\Uri\UriRetriever;
@@ -33,7 +34,7 @@ class Factory
     /**
      * @var int
      */
-    private $checkMode;
+    private $checkMode = Constraint::CHECK_MODE_NORMAL;
 
     /**
      * @var TypeCheck\TypeCheckInterface[]
@@ -54,8 +55,7 @@ class Factory
         'enum' => 'JsonSchema\Constraints\EnumConstraint',
         'format' => 'JsonSchema\Constraints\FormatConstraint',
         'schema' => 'JsonSchema\Constraints\SchemaConstraint',
-        'validator' => 'JsonSchema\Validator',
-        'coercer' => 'JsonSchema\Coerce'
+        'validator' => 'JsonSchema\Validator'
     );
 
     /**
@@ -73,9 +73,56 @@ class Factory
         UriRetrieverInterface $uriRetriever = null,
         $checkMode = Constraint::CHECK_MODE_NORMAL
     ) {
+        // set provided config options
+        $this->setConfig($checkMode);
+
         $this->uriRetriever = $uriRetriever ?: new UriRetriever;
         $this->schemaStorage = $schemaStorage ?: new SchemaStorage($this->uriRetriever);
+    }
+
+    /**
+     * Set config values
+     *
+     * @param int $checkMode Set checkMode options - does not preserve existing flags
+     */
+    public function setConfig($checkMode = Constraint::CHECK_MODE_NORMAL)
+    {
         $this->checkMode = $checkMode;
+    }
+
+    /**
+     * Enable checkMode flags
+     *
+     * @param int $options
+     */
+    public function addConfig($options)
+    {
+        $this->checkMode |= $options;
+    }
+
+    /**
+     * Disable checkMode flags
+     *
+     * @param int $options
+     */
+    public function removeConfig($options)
+    {
+        $this->checkMode &= ~$options;
+    }
+
+    /**
+     * Get checkMode option
+     *
+     * @param int $options Options to get, if null then return entire bitmask
+     *
+     * @return int
+     */
+    public function getConfig($options = null)
+    {
+        if ($options === null) {
+            return $this->checkMode;
+        }
+        return ($this->checkMode & $options);
     }
 
     /**
@@ -139,13 +186,5 @@ class Factory
         }
 
         return clone $this->instanceCache[$constraintName];
-    }
-
-    /**
-     * @return int
-     */
-    public function getCheckMode()
-    {
-        return $this->checkMode;
     }
 }
--- a/src/JsonSchema/Validator.php
+++ b/src/JsonSchema/Validator.php
@@ -10,7 +10,8 @@
 namespace JsonSchema;
 
 use JsonSchema\Constraints\Constraint;
-use JsonSchema\Entity\JsonPointer;
+use JsonSchema\Constraints\BaseConstraint;
+use JsonSchema\Exception\InvalidConfigException;
 
 /**
  * A JsonSchema Constraint
@@ -19,37 +20,47 @@ use JsonSchema\Entity\JsonPointer;
  * @author Bruno Prieto Reis <bruno.p.reis@gmail.com>
  * @see    README.md
  */
-class Validator extends Constraint
+class Validator extends BaseConstraint
 {
     const SCHEMA_MEDIA_TYPE = 'application/schema+json';
 
     /**
      * Validates the given data against the schema and returns an object containing the results
      * Both the php object and the schema are supposed to be a result of a json_decode call.
-     * The validation works as defined by the schema proposal in http://json-schema.org
+     * The validation works as defined by the schema proposal in http://json-schema.org.
+     *
+     * Note that the first argument is passwd by reference, so you must pass in a variable.
      *
      * {@inheritDoc}
      */
-    public function check($value, $schema = null, JsonPointer $path = null, $i = null)
+    public function validate(&$value, $schema = null, $checkMode = null)
     {
+        $initialCheckMode = $this->factory->getConfig();
+        if ($checkMode !== null) {
+            $this->factory->setConfig($checkMode);
+        }
+
         $validator = $this->factory->createInstanceFor('schema');
         $validator->check($value, $schema);
+
+        $this->factory->setConfig($initialCheckMode);
 
         $this->addErrors(array_unique($validator->getErrors(), SORT_REGULAR));
     }
 
     /**
-     * Does everything that check does, but will also coerce string values in the input to their expected
-     * types defined in the schema whenever possible. Note that the first argument is passed by reference,
-     * so you must pass in a variable.
-     *
-     * {@inheritDoc}
+     * Alias to validate(), to maintain backwards-compatibility with the previous API
      */
-    public function coerce(&$value, $schema = null, JsonPointer $path = null, $i = null)
+    public function check($value, $schema)
     {
-        $validator = $this->factory->createInstanceFor('schema');
-        $validator->coerce($value, $schema);
+        return $this->validate($value, $schema);
+    }
 
-        $this->addErrors(array_unique($validator->getErrors(), SORT_REGULAR));
+    /**
+     * Alias to validate(), to maintain backwards-compatibility with the previous API
+     */
+    public function coerce(&$value, $schema)
+    {
+        return $this->validate($value, $schema, Constraint::CHECK_MODE_COERCE_TYPES);
     }
 }
--- a/tests/Constraints/FormatTest.php
+++ b/tests/Constraints/FormatTest.php
@@ -23,7 +23,8 @@ class FormatTest extends BaseTestCase
         $validator = new FormatConstraint();
         $schema = new \stdClass;
 
-        $validator->check('10', $schema);
+        $checkValue = 10;
+        $validator->check($checkValue, $schema);
         $this->assertEmpty($validator->getErrors());
     }
 
@@ -34,15 +35,18 @@ class FormatTest extends BaseTestCase
         $schema->format = 'regex';
 
         $validator->reset();
-        $validator->check('\d+', $schema);
+        $checkValue = '\d+';
+        $validator->check($checkValue, $schema);
         $this->assertEmpty($validator->getErrors());
 
         $validator->reset();
-        $validator->check('^(abc]', $schema);
+        $checkValue = '^(abc]';
+        $validator->check($checkValue, $schema);
         $this->assertCount(1, $validator->getErrors());
 
         $validator->reset();
-        $validator->check('^猡猡獛$', $schema);
+        $checkValue = '^猡猡獛$';
+        $validator->check($checkValue, $schema);
         $this->assertEmpty($validator->getErrors());
     }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,4 @@ install:
   - travis_retry composer install --no-interaction --prefer-dist
 
 script:
-  - if [[ "$WITH_COVERAGE" == "true" ]]; then composer coverage; else composer test; fi
+  - if [[ "$WITH_COVERAGE" == "true" ]]; then ./vendor/bin/phpunit --coverage-text; else composer test; fi
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -1,0 +1,92 @@
+<?php
+
+/*
+ * This file is part of the JsonSchema package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace JsonSchema\Constraints;
+
+use JsonSchema\Entity\JsonPointer;
+
+/**
+ * A more basic constraint definition - used for the public
+ * interface to avoid exposing library internals.
+ */
+class BaseConstraint
+{
+    /**
+     * @var array Errors
+     */
+    protected $errors = array();
+
+    /**
+     * @var Factory
+     */
+    protected $factory;
+
+    /**
+     * @param Factory $factory
+     */
+    public function __construct(Factory $factory = null)
+    {
+        $this->factory = $factory ? : new Factory();
+    }
+
+    /**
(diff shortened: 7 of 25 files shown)


erayd:
I'm aware that I need to write some tests that cover the new bits - I'm working on that now!

bighappyface:
Quick tip: tests and new functionality come out better when the tests come first.

erayd:
You make an excellent point!

src/JsonSchema/Validator.php
```diff
@@ -43,13 +41,45 @@ public function check($value, $schema = null, JsonPointer $path = null, $i = nul
...
+    }
+
+    /**
+     * Alias of validate which does not allow any changes to the original document
+     * 
+     * {@inheritDoc}
+     */
+    public function validateImmutable($value, $schema = null, $checkMode = self::CHECK_MODE_NORMAL)
+    {
+        $checkMode &= $this->getCheckModeImmutableMask();
```
erayd: Should this throw an exception if `$checkMode` contains something that wants to modify `$value`, or just silently switch off the offending options?

Currently, it just switches them off, as this behavior is the closest to how the old `check()` method works in assoc mode. But I must admit, I would really rather have an exception here.

src/JsonSchema/Constraints/ConstraintInterface.php
```diff
@@ -59,5 +59,10 @@ public function isValid();
      * @param mixed            $i
      * @throws \JsonSchema\Exception\ExceptionInterface
      */
+    public function _check(&$value, $schema = null, JsonPointer $path = null, $i = null);
```
shmax: Isn't it a little weird for a public method to have the _ prefix?

src/JsonSchema/Validator.php
```diff
@@ -43,13 +41,50 @@ public function check($value, $schema = null, JsonPointer $path = null, $i = nul
...
+        // able to set it via the factory before calling this method.
+        $checkMode = $this->factory->getCheckMode() | self::CHECK_MODE_COERCE;
 
-        $this->addErrors(array_unique($validator->getErrors(), SORT_REGULAR));
+        return $this->validate($value, $schema, $checkMode);
+    }
+
+    /**
+     * Validates the given data against the schema and returns an object containing the results
+     * Both the php object and the schema are supposed to be a result of a json_decode call.
```
shmax: Why would you make a point of mentioning json_decode? There are other ways to come up with an object...

```PHP
$schema = (object)['foo'=>'bar'];

$schema = new stdClass();
$schema->foo = 'bar';
```

src/JsonSchema/Validator.php
shmax: Why is this necessary? Isn't it enough that the default $checkmode is CHECK_MODE_NORMAL? I would think that as long as all of your helpers put things back the way they found them that this kind of thing is not needed...

src/JsonSchema/Validator.php
```diff
@@ -43,13 +41,50 @@ public function check($value, $schema = null, JsonPointer $path = null, $i = nul
...
+     * The validation works as defined by the schema proposal in http://json-schema.org.
+     *
+     * @param mixed $value Decoded JSON document to validate
+     * @param mixed $schema Decoded JSON schema to validate against
+     * @param int $checkMode Set various options to change the validation behavior
+     *
+     * {@inheritDoc}
+     */
+    public function validate(&$value, $schema = null, $checkMode = self::CHECK_MODE_NORMAL)
```
shmax: I like the idea that you have sort of an advanced usage method where you can explicitly state the kinds of bells and whistles you want. I would argue that this should be the *main* entry point, and that everything else (`coerce`, `check`, `forceDefaults`, etc) could be considered helpers, shortcuts, or possibly even eventually deprecated and removed entirely.

One suggestion you might consider is to convert all instances of `_check` (which is confusingly public and mirroroing `check`) with `validate`, and just pass $checkmode along everywhere.

src/JsonSchema/Constraints/ConstraintInterface.php
erayd: It's not part of the public API (hence why it's prefixed with _, as is convention for internal stuff). It needs to be designated a public method because it gets called from all over the place within the library, but not just from class descendants.

src/JsonSchema/Validator.php
erayd: I didn't - this is verbatim from master (courtesy of @digitalkaoz 5 years ago). I saw no reason to change it, as it explains the intended use-case quite well.

src/JsonSchema/Constraints/ConstraintInterface.php
shmax: Well, right, but this isn't a situation where you have to rely on the underscore convention to suggest visibility (like we often do in javascript), because we have actual visibility keywords. I think most people would agree it's sort of poor form to make a method public, but suggest that client code shouldn't call it with the underscore convention. Either it's private and you can't call it, or it's public and you can. Take a stand!

src/JsonSchema/Validator.php
shmax: I find it misleading. Not everybody is decoding json (I'm not--all of my schemas come from php files).

src/JsonSchema/Validator.php
erayd: The `validate()` method is *supposed* to be the main entry point for everything except pass-by-value calls, which use `validateImmutable()` - so not sure what you're disagreeing with here? `check()` and `coerce()` are intended to be deprecated; the only reason they're still around is to avoid breaking things for people who are currently using them.

I would prefer to avoid conflating `validate()` with `_check()` - they're intended to be different things. Having the public API be the same thing as all the internal stuff seems rather messy IMO - that's what happened with `coerce()`, and half the point of this PR is to remove any need for the 'pass-everything-down-the-stack' approach. Unless you're suggesting we make every part of this library thread-safe, then going back to passing stuff around seems like a backwards step.

src/JsonSchema/Constraints/ConstraintInterface.php
erayd: This method *cannot* be made private (or protected), because the entire library will break as a result. The various internal callers need a public method to talk to; that's just the way the library is architected. 

I agree that having it actually *be* a public method is undesirable, but unless you have a suggestion for how to mitigate that requirement, then I think it's just something we need to live with, and all we can do is make it look as undesirable as possible for an end-user to rely on. Re-architecting the entire library to change a single visibility keyword seems like a big waste of time - it seems reasonable in this case to just be pragmatic and leave it alone, albeit well-documented as an internal method.

src/JsonSchema/Validator.php
erayd: Can certainly change it... what would you like to see instead?

src/JsonSchema/Validator.php
erayd: It's there because there needs to be some kind of entry point for pass-by-value, and not enforcing immutability results in inconsistent behavior between object mode and assoc mode, because objects are *never* passed by value. Having the mask means that no matter what the user decides to feed it for $checkMode, the result will be the same regardless of whether $value is an object or an array.

src/JsonSchema/Validator.php
shmax: I'm not disagreeing, really, just trying to help massage this into something a little more structurally solid. Well, I guess I do disagree with `validateImmutable`--adding a method whose sole purpose is to ignore certain flags you pass in seems a little bizarro-world to me--but otherwise I like the general tack you're taking. 

Really, I was trying to give you an elegant way to avoid the corner you've painted yourself into with `_check`. Before you started all this we had two single public methods at the top level. I suggested reducing them to one (yes, that would conceivably break backwards compatibility for what is probably a fairly small set of people, but we're not shy about doing that in this project if the rationale is strong, and I think this qualifies). But now, instead of just `check`, we now have `check`. `validateImmutable`, `coerce`, `_check`, and `validate`. What happened!?

src/JsonSchema/Constraints/ConstraintInterface.php
erayd: I guess another option would be moving the contents of `Validator::check()` into `validate()`, and leaving `_check()` as an empty method (it can't be removed without also removing it from the interface, which is an important sanity check for those writing constraints). I don't much mind either way - would having `Validator::_check()` be a noop be preferable for you?

src/JsonSchema/Validator.php
shmax: Don't know. Might be enough to just remove the comment about json_decode, and document the mixed type for $value?

src/JsonSchema/Constraints/ConstraintInterface.php
shmax: My original suggestion from last night was to bite the bullet and modify `check` (which is already public) to take a reference. Yes, it will break bc for some people, but I think we have enough rationale at this point to justify making the change. If you want to do it in steps, we could begin by replacing `coerce` with `validate` (which takes a reference, but otherwise makes no assumptions about what you want to do, and bringing the grand total of public methods to 3), and deprecating `check`, then eventually removing `check`.

src/JsonSchema/Validator.php
erayd: Hmm.... maybe we see this a little differently? I don't feel like this is a corner.

We went from two public methods (`check` and `coerce`) to two public methods (`validate` and `validateImmutable`), and in doing so significantly improved the interface from an end-user perspective, plus allowing for easy addition of new functionality in the future without needing to change the public API again.

If you don't care about pass-by-value at all, then `validateImmutable` can be removed; that's the entire reason it's there.

If you don't care about breaking backwards-compatibility, then `check` and `coerce` can certainly be dropped... but IMO keeping them is safer - they're just aliases now anyway. Perhaps having them trigger a deprecation notice when called would be a sensible middle-ground, and then remove them in a future release?

`_check` is *not* a public method, whatever its visibility keyword says. The only reason it exists at all is because `ConstraintInterface` requires it. Would making `_check` be a noop be an improvement? Because the content of that method can easily be moved into `validate`.

src/JsonSchema/Validator.php
shmax: Still not following. What exactly is the use case? How is calling this different from calling validate with no flags?

src/JsonSchema/Constraints/ConstraintInterface.php
erayd: I think we may have different ideas on what constitutes acceptable grounds for breaking a public API.

My position on this kind of thing is that breaking a five-year-old public API immediately and without warning when a couple of aliases can allow for graceful deprecation over the next few releases is unconscionable, regardless of how trivial the fix on their end. Users deserve more respect than that.

That said, this is not my library... so if the consensus is that breaking the API is in fact the desired outcome, then I'll go along with that. But I'm not happy about it.

I wrote this PR because I thought that your suggestion was to deprecate `coerce`, unify behavior with `check`, and expose `checkMode` in the public API - and this PR seemed like the best way to go about it - but have I misunderstood your original intention?

src/JsonSchema/Validator.php
erayd: That works - I'll do that.

src/JsonSchema/Validator.php
erayd: The use-case is being able to pass $value as not-a-reference. If we don't care about that, then `validateImmutable` can go; there's no other reason to having it. Thoughts?

The immutable mask thing is only to enforce consistency between assoc and object modes, and if this method goes then it's a non-issue.

src/JsonSchema/Validator.php
shmax: I count two public methods on ConstraintInterface, `check` and `_check`, and three more on Validator (which is the public face of the app, and formerly was a vanilla override of Constraint), `coerce`, `validate` and `validateImmutable`. You're arguing that you are making things simpler, but we've gone from 2 publics on the main app instance to 5, and I barely understand it and I'm trying my honest best to keep up.

You have my endorsement to just change `check`. That'll take us down to one public method that can take all comers. We can gather some opinions from other people, but I feel like that will be much, much easier to sell than all this other stuff. 

Gonna take a dinner break. Overall I'm liking the energy. Hang in there.

src/JsonSchema/Constraints/ConstraintInterface.php
shmax: > My position on this kind of thing is that breaking a five-year-old public API immediately and without warning when a couple of aliases can allow for graceful deprecation over the next few releases is unconscionable, regardless of how trivial the fix on their end. Users deserve more respect than that.

You're new to this project. The main API has gone through a lot of changes. It is not 5 years old. The API as you know it now is about three months old. We had a breaking change just a few months ago that was more dramatic than this (they changed the API such that instead of feeding storage and uriretriever instances and so on to Validator, you pass them to Factory). To be honest that one went in a little faster than I would have liked--the checkmode got swept along as well for no real reason, which led us to our current situation. But we bumped the major version, and had no complaints.

shmax: > I wrote this PR because I thought that your suggestion was to deprecate coerce, unify behavior with check, and expose checkMode in the public API - and this PR seemed like the best way to go about it - but have I misunderstood your original intention?

Nope, you nailed it. Here's my comment from the other thread:

>>Tell you what--as long as we're going to change the API, let's go all the way and just get rid of coerce, and convert that (back) to a config flag, as well. Then we'll modify check to just take a reference everywhere, and we'll be all set.

erayd: OK - I'm certainly new to this project, and if you guys are OK with breaking the API, then that's your call, and I'll go along with that.

So... I will make it so that this PR has one top-level method named `check`, which will do what `validate` does now. All the other ones will go. Does that sound OK?

`_check` will need to stay, unless I do some significant refactoring to change the way those calls work. Keeping it there seems like the most sensible option at this point - however if you feel strongly about that, in prepared to do that work.

src/JsonSchema/Validator.php
shmax: Well, you already know my opinion. :)

src/JsonSchema/Constraints/ConstraintInterface.php
shmax: > however if you feel strongly about that,

I do feel strongly about it, but I'm also available to help code. Feel free to put me to work.

src/JsonSchema/Validator.php
erayd: OK then - I'll get rid of `validateImmutable` :-).

src/JsonSchema/Constraints/ConstraintInterface.php
erayd: Nah, I'm happy to do it, seeing as my wanting to add the defaults PR is the source of this whole headache - it seems reasonable that if I want that feature, it should be my time that gets used making it happen. Code review is always appreciated though :-).

shmax: 🥇 By the way, just so there's no misunderstanding, I am not a project maintainer--just the most vocal contributor these days.

erayd: Gotcha. Do you know whether the maintainer is likely to agree with you on this? Is there any chance of us getting their opinion on the situation?

Vocal means you care, and if you're the only contributor with a strong opinion, and the maintainer is OK with that, then I'm happy to defer to your preference somewhat.

shmax: @bighappyface Is the only active contributor with write privileges. He's usually pretty mellow, but no harm in getting his opinion now.

erayd: Have not heard anything from @bighappyface, so I'm just going to go ahead and implement this as discussed above (one top-level method Validator::check(), all others to go, refactor so that _check is not public).

erayd:
@shmax - have redone this PR more along the lines we discussed - there's now just one public method defined in `Validator`. Thoughts?

I've opted to use a config array rather than just $checkMode, in order to future-proof things in case somebody wants to add a feature that can't be configured via a partial bitmask. That way the risk of needing to change the public API again in order to support new features is minimised.

shmax:
I think it's a huge improvement in a lot of ways; I like that all those `$coerce` parameters can go away, and I really like that you got us back down to a single public method, and I *love* that you abstracted `BaseConstraint` out so that the top level `check` doesn't have all that weird internal $i stuff on it (I've been wanting to do something similar). I can't say I'm a fan of the config property bag idea, however, for a few reasons:

* You see property bags all the time in javascript, but it's not as common in PHP. I don't know of any native functions that do it, but there are definitely a few that use the bitmask style. 
* One nice thing about using the bitmasks and consts defined on classes is that you can discover them via intellisense in most modern IDEs.
* this will *really* break backwards compatibility. Your point about having a little more flexibility for future changes is fair and well-taken, but my instinct is that we can cross that bridge if and when we come to it, and that this much of a change might be hard to sell without something a little more concrete to support it than "we might need it someday". But if you're dead set on it I'm willing to defer to other reviewers.

What do you think?

erayd:
Can you clarify what your issue is with having that `$config` array? It would be relatively easy for me to refactor that bit of the PR to work differently, but I guess I don't really understand where you're seeing it as being a problem, and I don't want to just guess. Would you prefer that the config have its own class, rather than being an array?

I do feel fairly strongly that a bitmask alone is insufficient - but a bitmask is definitely still useful, which is why I turned `checkMode` into a config property, rather than completely getting rid of it.

Re backwards compatibility - yes, this definitely breaks compatibility. I thought that you decided that did not matter - but I'm quite happy to keep some backwards compatibility if you let me know which bits of the API you feel need to be not-broken so that I can ensure they maintain compatibility. `checkMode` does still exist as a config property, and it would be very easy to expose that in the places you feel that it should remain part of the public API.

erayd:
@shmax Further to my comment above, one of the reasons I made coerce & apply-defaults as independent booleans rather than being part of the `checkMode` bitmask was to make them easy to set independently. This isn't a fire-and-forget thing the way most PHP bitmask configs are (because `Factory` can be reused across multiple validations), so it makes sense to allow the user to adjust individual settings without requiring them to do a bunch of extra bitwise operations just to avoid clobbering unrelated settings.

shmax:
Sure, I'll take another stab at explaining my position: I liked it better before when $checkMode was just an int, and turning coercion on was a matter of applying CHECK_MODE_COERCE (that might have been before your time--I eventually removed it when I added the `coerce` method, but now we're in a position to put it back). Since we don't currently have any features or options that can't easily be handled with the bitmask, my vote is to stick with it. 

I didn't say that breaking compatibility doesn't matter--I said we're not shy about doing it when the situation warrants it. The config bag thing is not a must-have--it's more like a personal coding style that you favor, and it doesn't buy us anything in the short term (and we're a little hazy on the long term). To be fair, changing `check` to take a reference also breaks compatibility for some people (I started poking through projects at random on github--more people than I would have expected do use `json_decode` inline with `check`, as you warned), but it makes two major features that you and I want (apply defaults and coerce) much more doable, and adapting to the change is trivial (not that yours is that much of a big deal, but people would have to look at the documentation).

Really, I think everything you did is great, but if it were me I would get rid of the property bag, put $checkMode back the way it was (as a top-level argument to `Factory`), and re-add `CHECK_MODE_COERCE`.

erayd:
@shmax Also further to your point re future-proofing - it makes sense to me that if we are breaking the API *anyway* (which we are, because `Validator::coerce()` is going away and `Validator::check()` now requires a reference, we might as well do it properly, once, rather than making the API too limited now and potentially breaking it again later - because every time the API breaks and requires users to refactor stuff, that will annoy them.

erayd:
@shmax Give me a moment, and I'll see if I can make a few changes that will bring things more in line with what I think you're after.

shmax:
Righto, sounds good.

shmax:
Just a quick response to your comment about `Validator::coerce` going away--that's true, but I don't think too many people have started using that. I did a quick search on github and only turned up 25 results, and of those the first 5 or 6 I checked were people that added their vendor directory to their repo for some reason, and the match was coming from `CHECK_MODE_COERCE` (which was around for longer).

shmax:
Hmm, I just realized something. Now that you've divorced Validator::check from Constraint::check, we can pass $checkMode along as a third argument. That wouldn't break compatibility because existing code will only be using two arguments (with the third and fourth currently being $path and $i, which don't really have meaning from the top entry point), AND you then have true "fire and forget" functionality, with more of a one-liner feel (instead of having to keep a Factory instance around so that you can twiddle flags on it)...

shmax:
Oh, damn, I see that you already are passing $checkMode as a third argument. My apologies--now I understand some of your bafflement. I'd better get some sleep before I make things any worse, but I'll be ready to resume the discussion tomorrow, if that's okay with you.

shmax:
Heading to bed, but I think it's looking great--we're almost there. More soon.

erayd:
@shmax OK, so I've done the following - thoughts?

  1. Renamed the new API to `Validator::validate()`;
  2. Added a `Validator::check()`->`Validator::validate()` alias to maintain backwards-compatibility for people who use pass-by-value;
  3. Restored `$checkMode` as the third argument to `Factory::__construct()`;
  4. Added `$checkMode` as the third argument to `Validator::validate()` (`$config` moves to fourth);
  5. Defined bitmask flags `CHECK_MODE_COERCE_TYPES` and `CHECK_MODE_APPLY_DEFAULTS`;
  6. If `$checkMode` is set on `Validator::validate()`, then those flags will override the config options.

These changes preserve much more backwards-compatibility, still keep a simple API, allow fire-and-forget via `$checkMode` flags, and still keeps the config array around so that factory reuse and bitmask-incompatible features remain easy. Does this sufficiently address the concerns you had earlier?

I see your heading-to-bed comment - sleep well :-).

erayd:
Do you want an alias for `Validator::coerce()` as well? Or prefer to take this out regardless, and not worry about backwards-compatibility for those users?

It doesn't need to be more than a one-liner:
```PHP
public function coerce(&$value, $schema)
{
    return $this->validate($value, $schema, null, array('coerce-types' => true));
}
```

shmax:
Had a busy morning and now I need to get to work, but I'll get back to you tonight...

shmax:
So it's looking good, but I don't quite understand why `checkMode` is expected as both/either the 3rd argument to `validate` AND/OR as a property on $config. It seems to me that if we've managed to agree on $checkMode as the 3rd argument, then there's not much need to support it on $config. And that leaves $config empty, which means it doesn't really need to be there, right? If you ever do come up with a complex config option that can't be handled with $checkMode, then you can re-introduce $config at that time as the 4th argument without breaking compatibility.

And my vote is to drop `coerce`. This is an easy band-aid to rip off, as few people are likely to be using it. If you do decide to leave it in, okay, but maybe we could at least put a @deprecated tag in its PHPDoc...

erayd:
I mainly put $checkMode as the third argument because it seemed like something that was important to you - I wouldn't go so far as to say I agree with it, but it seems to me that you care about it a lot, so I'm trying to come up with a compromise. It sounds like you're not entirely happy with the latest suggestion, so I'll tweak it again - stand by...

erayd:
@shmax How about now?

The biggest point in favour of having a config array *now*, IMO, is the ability to change options individually without requiring the user to do a bunch of bitwise operations just to avoid clobbering the settings they're *not* changing. I guess I'm not really understanding why you have such a strong resistance to it - does the latest lot of changes implement it in a away you find acceptable, or am I missing something fundamental about your objection to it?

No idea what you were talking about with `CHECK_MODE_SET_DEFAULTS` / `set-default`; those never existed in any of my code. Can you clarify?

shmax:
Hmm, interesting. I'll have to take a closer look at this tomorrow; but here's a few reasons why I object to your property bags idea in this specific case, and a few why I don't like them in general in PHP: 

* you can't typehint the properties. Best you can do is typehint the object itself. You've been implying that the config object will come in handy if we start dreaming up more complex config properties, but I would just argue that you're making a bad pattern worse because the more complex a property is, the more useful a typehint might be. 
* if we stick with a bitmask and flag constants, it's very difficult to misuse. Most IDEs will help you discover the available consts with intellisense, and even if you don't make use of that feature any mistyped flags will certainly get caught at runtime. The same is not true of strings like "coerce-types". If you mistype it as "coerce-type", it will simply fail silently.
* It's not a very PHP-y kind of thing, at least not as an argument to a method. You see it all the time in javascript, and I'll confess I do it in my own code sometimes (usually when constructing widgets with a lot of config properties), but I almost never see it out in the wild. I'm not arguing that people won't be able to understand it or something, just that it's not a popular pattern, and it makes me suspect that there are probably other reasons not to do it that I haven't even thought of.

But anyway, let me take another look at it tomorrow with fresh eyes. G'nite.

shmax:
> the ability to change options individually without requiring the user to do a bunch of bitwise operations just to avoid clobbering the settings they're not changing

One last comment; I thought we had dodged this problem by passing checkMode in as an argument on every call to `validate`? Or is it still storing the value of checkMode internally until the next call?

erayd:
Thanks for explaining :-). Your point about typehinting is a really good one, and not something I'd considered. So is the thing with typos - at least with constants, you get an error if you typo it.

Noting those two points, I think you've convinced me to ditch the config array. I still care about being able to set individual options, but that's doable with checkmode alone if I tweak it a bit. Stand by for more changes...

> One last comment; I thought we had dodged this problem by passing checkMode in as an argument on every call to validate? Or is it still storing the value of checkMode internally until the next call?
It's still stored internally on `Factory` - `Validator::validate()` just saves and resets it to allow fire-and-forget using different settings.

erayd:
OK, what about this?
  1. `$config` is completely gone;
  2. `$checkMode` is the sole means of configuring features;
  3. `$checkMode` is still stored internally in `Factory` to allow reuse;
  4. `Validator::validate()` saves and restores the existing `$checkMode` to allow fire-and-forget;
  5. `Factory::setConfig()` can (avoids the user needing to manually preserve settings):
        * set the entire `$checkMode` bitmask;
        * add options to the current `$checkMode`;
        * remove options from the current `$checkMode`.
  6. `Factory::getConfig()` can:
        * get the entire `$checkMode` bitmask;
        * get a subset of the `$checkMode` bitmask.

shmax:
Still with you... had a busy day, but I'll come back to this tomorrow. Thanks for hanging in there...

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -20,81 +20,15 @@
...
-    /**
-     * Clears any reported errors.  Should be used between
-     * multiple validation checks.
-     */
-    public function reset()
-    {
-        $this->errors = array();
-    }
+    const CHECK_MODE_COERCE_TYPES =   	0x00000004;
+    const CHECK_MODE_APPLY_DEFAULTS =  	0x00000008;
```
shmax: can you line the numeric literals up so that they're in a neat column?

src/JsonSchema/Constraints/Factory.php
```diff
@@ -73,9 +73,43 @@ public function __construct(
...
-        $this->checkMode = $checkMode;
+    }
+
+    /**
+     * Set config values
+     *
+     * @param int $checkMode Set checkMode options
+     * @param bool $enable Whether to enable or disable the provided options. If null, overwrites checkMode.
+     */
+    public function setConfig($checkMode = Constraint::CHECK_MODE_NORMAL, $enable = null)
```
shmax: I find this a little confusing. If I pass in $checkMode but $not enable, then I overwrite the whole thing. Otherwise, you either add or remove the flag, dependent on $enable. Do you really think people are going to be busily setting and unsetting specific flags over the course of a program? Wouldn't it be simpler to just drop the $enable arg and just make this a simple set like you have on line 92? If someone want to unset a flag, they would do so by just setting the exact config they want and blowing away the old config...

shmax: And if you really want the ability to add and disable specific flags, maybe you could add `addConfig` and `removeConfig` helpers

tests/Constraints/FormatTest.php
```diff
@@ -23,7 +23,8 @@ public function testNullThing()
         $validator = new FormatConstraint();
         $schema = new \stdClass;
 
-        $validator->check('10', $schema);
+        $checkValue = 10;
+        $validator->check($checkValue, $schema);
```
shmax: Just curious; is it still necessary to pass in a variable now that top-level check no longer takes a reference?

src/JsonSchema/Validator.php
```diff
@@ -19,37 +20,47 @@
...
+     *
+     * Note that the first argument is passwd by reference, so you must pass in a variable.
      *
      * {@inheritDoc}
      */
-    public function check($value, $schema = null, JsonPointer $path = null, $i = null)
+    public function validate(&$value, $schema = null, $checkMode = null)
     {
+        $initialCheckMode = $this->factory->getConfig();
+        if ($checkMode !== null) {
```
shmax: so if I understand correctly, `validate` always bulldozes whatever is already  set in the internal checkMode, and, respects the incoming $checkMode, then restores it afterward. If that's true, then shouldn't $checkMode default to CHECK_MODE_NORMAL, and always call `$this->factory->setConfig` whether $checkMode (the arg) is set or not?

src/JsonSchema/Constraints/Constraint.php
erayd: Sorted - looks like there were tabs in there that shouldn't have been, and I have vim configured to display tabs using a 4-char tabstop, so I didn't see them. Well spotted!

src/JsonSchema/Constraints/Factory.php
erayd: `$enable` is now gone, and `setConfig()` will always overwrite. I do really want to keep the ability to enable / disable specific flags, so have added `addConfig()` and `removeConfig()` per your suggestion.

tests/Constraints/FormatTest.php
erayd: Yes, because the individual validators *do* require a reference. The references are needed for anything that modifies the input value, including type coercion and setting default values.

The line you've highlighted here tests `FormatConstraint` directly, not the top-level `Validator` class.

src/JsonSchema/Validator.php
erayd: No. `validate()` only clobbers `checkMode` if the user actually passes it. If `$checkMode` is left as null, it uses whatever was already set on the factory. This is why the default value on this method is `null`.

The actual default *is* `CHECK_MODE_NORMAL` - it's set as the third argument to `Factory::__construct()`.

src/JsonSchema/Constraints/Factory.php
erayd: > Do you really think people are going to be busily setting and unsetting specific flags over the course of a program?

Not so much that - I had two specific use-cases in mind:
  1. If a flag is default-enabled internally, and the user doesn't know that, but wants to set a flag of their own - `addConfig`() is clearly safer in that situation.
  2. If a user wishes to disable a single default-enabled flag without messing up whatever else might be in `$checkMode`.

At the moment the default is just `CHECK_MODE_NORMAL`, but that may not always be the case.

tests/Constraints/FormatTest.php
shmax: Ah, ok. Sounds good.

src/JsonSchema/Validator.php
shmax: Okay. I think I can live with that.

shmax:
Are you ready for it..? 🥁 

LGTM!

Beautiful work. Sorry for being such a pain.

erayd:
Hooray! Thanks for all the time you've spent reviewing this :-).

You weren't a pain - you were thinking through the ramifications of the changes from a different perspective to mine, and it was useful - I think the end result is much better because of all your input.

Does anyone else need to review before merge / do we need to flag anyone to get this merged?

erayd:
Squashed in 2159893, because the commit history for this PR was full of in-progress junk that's no longer applicable.

shmax:
Well, at this point @bighappyface face needs to look it over. He might want to get a few more eyes on it... 

@digitalkaoz please review

shmax:
🎉 Thanks @bighappyface and congrats @erayd

erayd:
Woohoo! :-D

PR #356, opened 2017-02-21
Add composer scripts for php-cs-fixer

erayd:
## What
 * Add php-cs-fixer to require-dev
 * Add `style-check` and `style-fix` scripts to composer

## Why
 * To make style checks & fixes easy (currently, php-cs-fixer must be manually fetched and run)
 * To encourage users to run these checks locally first rather than relying on travis integration to flag errors


bighappyface:
Thoughts on TravisCI failure?

erayd:
I've been trying to figure out how to resolve it - think I have a solution now. Have set sed to remove the require-dev dependency on php-cs-fixer for hhvm and nightly builds, because php-cs-fixer explicitly conflicts with them, and we're not trying to run it on those platforms anyway.

erayd:
Yep, that worked - Travis is now happy :-).

bighappyface:
Works for me. @alcohol @shmax any concerns?

erayd:
Rebased on master following merge of PR #349.

alcohol:
None at all, seems like an elegant enough solution for CI, definitely. :-)

Issue #353, opened 2017-01-20
Validation of JSON-Schema


o5:
Am I missing something or there is not validation of schema itself? I don't know if it is possible, but it would be nice to throw something like json-schema parse error.

For json-schema:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
      "id": {
        "type": "number"
      },
      "type": "object"
    }
}
```

I've expected similar error message as I got from [jsonschemavalidator.net](http://www.jsonschemavalidator.net): 

```
Unexpected token encountered when reading value for 'type'. Expected StartObject, got String. Path 'properties.type', line 7, position 22.
```

shmax:
I don't think the library has that built-in, but if you want to set that up in your own code you might have a look at the files in `/tests/fixtures` for some ideas; they appear to be validation schemas for the different draft specs.

mathroc:
@o5 you can validate it yourself using the validator and the [json-schema schema](http://json-schema.org/draft-04/schema)

```php
$validator->validate(json_decode($schema), (object)['$ref' => 'http://json-schema.org/draft-04/schema']);
```

o5:
Hello @mathroc,

I'm sorry for my late response. I'm trying to use validate feature:

```php
$schema = json_decode('{
	"$schema": "http://json-schema.org/draft-04/schema#",
	"properties": {
		"id": {
			"type": "number"
		},
		"type": "object"
	}
}');

$validator = new Validator();
$validator->validate(
    $schema,
    (object) ['$ref' => 'http://json-schema.org/draft-04/schema#'],
    Constraint::CHECK_MODE_VALIDATE_SCHEMA
);
```

I'm expecting some error like *Error parsing schema*, but none given. Am I missing something?

erayd:
>I'm expecting some error like Error parsing schema, but none given. Am I missing something?

You're asking it to validate the schema, and the schema is valid (it's the official draft-04 meta-schema). That isn't an error condition.

`$schema` is not a valid schema, but you are presenting it as the document to be validated against the official spec, not as the schema. If I'm correctly understanding what you're doing there, the result should be a validation error.

o5:
@erayd:

Sorry, the snippet should be this :)

```php
$schema = json_decode('{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
        "id": {
            "type": "number"
        },
        "type": "object"
    }
}');

$value = (object) [
    'id' => 10,
];

[...]

Issue #239, opened 2016-03-15
Custom error message 


csimplestring:
Hi, it is great to support customized error message in the 1.6.0 release. But after reading the source code  for a while, I notice a minor issue during adding error:

`
$this->addError($path, 'The item ' . $i . '[' . $k . '] is not defined and the definition does not allow additional items', 'additionalItems', array('additionalItems' => $schema->additionalItems,));
`

The problem is that the 'not-allowed' field '$i' and '$k' is in the error message, external users have to parse it from error message. Once the error message is changed, external users have to change parsing accordingly. 

Therefore to fully support customized error message, the error message should not contain any error field path, which shall be moved to '$path' instead.

jonathan-s:
I haven't checked how you implement custom error messages, but the approach described in this issue seems like a rather nice one: https://github.com/geraintluff/tv4/issues/115#issuecomment-41093371

abreksa4:
+1

DannyvdSluijs:
The error message you mentioned above is still in the code base but has shifted a bit. For reference you can find them here:
https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/ConstraintError.php#L59
https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/CollectionConstraint.php#L101-L110

@csimplestring Since your issue report the repo has progressed and since #364 the property became available. Using below  example code you can see the property is now part of the errors.
In an attempt to cleanup this repo we are trying to filter the issues and see which ones might be closed. Is it safe to assume this is a rather old issue, which sadly was left unanswered, but now can be closed? Feel free to close it yourself with some comments if helpful.

**Schema:**
```json
{
  "type": "object",
  "additionalProperties": false
}
```
**Code:**
```php
<?php

use JsonSchema\Constraints\Constraint;
use JsonSchema\Uri\UriRetriever;
use JsonSchema\Validator;

require_once '../../vendor/autoload.php';

$retriever = new UriRetriever();
$schema = $retriever->retrieve('internal://' . realpath('./schema.json'));
$obj = json_decode('{"foo": "bar"}', false);
$validator = new Validator();
$result = $validator->validate($obj, $schema, Constraint::CHECK_MODE_APPLY_DEFAULTS);

var_dump($validator->getErrors()[0]["constraint"]["params"]["property"]);
```
**Output:**
```sh
string(3) "foo"
```

PR #357, opened 2017-02-22
Add option to validate the schema

erayd:
Validates the schema against whatever spec is referenced in `$schema` (see issue #353).

This PR is based on #362; that one needs merging first.

### DIFF
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -14,6 +14,7 @@ use JsonSchema\Constraints\TypeCheck\LooseTypeCheck;
 use JsonSchema\Entity\JsonPointer;
 use JsonSchema\Exception\InvalidArgumentException;
 use JsonSchema\Exception\ValidationException;
+use JsonSchema\Validator;
 
 /**
  * A more basic constraint definition - used for the public
@@ -25,6 +26,11 @@ class BaseConstraint
      * @var array Errors
      */
     protected $errors = array();
+
+    /**
+     * @var int All error types which have occurred
+     */
+    protected $errorMask = Validator::ERROR_NONE;
 
     /**
      * @var Factory
@@ -56,7 +62,8 @@ class BaseConstraint
             'constraint' => array(
                 'name' => $name,
                 'params' => $more
-            )
+            ),
+            'context' => $this->factory->getErrorContext(),
         );
 
         if ($this->factory->getConfig(Constraint::CHECK_MODE_EXCEPTIONS)) {
@@ -64,18 +71,42 @@ class BaseConstraint
         }
 
         $this->errors[] = $error;
+        $this->errorMask |= $error['context'];
     }
 
     public function addErrors(array $errors)
     {
         if ($errors) {
             $this->errors = array_merge($this->errors, $errors);
+            $errorMask = &$this->errorMask;
+            array_walk($errors, function ($error) use (&$errorMask) {
+                if (isset($error['context'])) {
+                    $errorMask |= $error['context'];
+                }
+            });
         }
     }
 
-    public function getErrors()
+    public function getErrors($errorContext = Validator::ERROR_ALL)
     {
-        return $this->errors;
+        if ($errorContext === Validator::ERROR_ALL) {
+            return $this->errors;
+        }
+
+        return array_filter($this->errors, function ($error) use ($errorContext) {
+            if ($errorContext & $error['context']) {
+                return true;
+            }
+        });
+    }
+
+    public function numErrors($errorContext = Validator::ERROR_ALL)
+    {
+        if ($errorContext === Validator::ERROR_ALL) {
+            return count($this->errors);
+        }
+
+        return count($this->getErrors($errorContext));
     }
 
     public function isValid()
@@ -90,6 +121,17 @@ class BaseConstraint
     public function reset()
     {
         $this->errors = array();
+        $this->errorMask = Validator::ERROR_NONE;
+    }
+
+    /**
+     * Get the error mask
+     *
+     * @return int
+     */
+    public function getErrorMask()
+    {
+        return $this->errorMask;
     }
 
     /**
--- a/src/JsonSchema/Constraints/Factory.php
+++ b/src/JsonSchema/Constraints/Factory.php
@@ -16,6 +16,7 @@ use JsonSchema\SchemaStorage;
 use JsonSchema\SchemaStorageInterface;
 use JsonSchema\Uri\UriRetriever;
 use JsonSchema\UriRetrieverInterface;
+use JsonSchema\Validator;
 
 /**
  * Factory for centralize constraint initialization.
@@ -41,6 +42,11 @@ class Factory
      * @var TypeCheck\TypeCheckInterface[]
      */
     private $typeCheck = array();
+
+    /**
+     * @var int Validation context
+     */
+    protected $errorContext = Validator::ERROR_DOCUMENT_VALIDATION;
 
     /**
      * @var array
@@ -192,5 +198,25 @@ class Factory
         }
 
         return clone $this->instanceCache[$constraintName];
+    }
+
+    /**
+     * Get the error context
+     *
+     * @return string
+     */
+    public function getErrorContext()
+    {
+        return $this->errorContext;
+    }
+
+    /**
+     * Set the error context
+     *
+     * @param string $validationContext
+     */
+    public function setErrorContext($errorContext)
+    {
+        $this->errorContext = $errorContext;
     }
 }
--- a/src/JsonSchema/Constraints/SchemaConstraint.php
+++ b/src/JsonSchema/Constraints/SchemaConstraint.php
@@ -9,8 +9,13 @@
 
 namespace JsonSchema\Constraints;
 
+use JsonSchema\ConstraintError;
 use JsonSchema\Entity\JsonPointer;
 use JsonSchema\Exception\InvalidArgumentException;
+use JsonSchema\Exception\InvalidSchemaException;
+use JsonSchema\Exception\RuntimeException;
+use JsonSchema\SchemaStorage;
+use JsonSchema\Validator;
 
 /**
  * The SchemaConstraint Constraints, validates an element against a given schema
@@ -20,6 +25,8 @@ use JsonSchema\Exception\InvalidArgumentException;
  */
 class SchemaConstraint extends Constraint
 {
+    const DEFAULT_SCHEMA_SPEC = 'http://json-schema.org/draft-04/schema#';
+
     /**
      * {@inheritdoc}
      */
@@ -27,16 +34,62 @@ class SchemaConstraint extends Constraint
     {
         if ($schema !== null) {
             // passed schema
-            $this->checkUndefined($element, $schema, $path, $i);
+            $validationSchema = $schema;
         } elseif ($this->getTypeCheck()->propertyExists($element, $this->inlineSchemaProperty)) {
             $inlineSchema = $this->getTypeCheck()->propertyGet($element, $this->inlineSchemaProperty);
             if (is_array($inlineSchema)) {
                 $inlineSchema = json_decode(json_encode($inlineSchema));
             }
             // inline schema
-            $this->checkUndefined($element, $inlineSchema, $path, $i);
+            $validationSchema = $inlineSchema;
         } else {
             throw new InvalidArgumentException('no schema found to verify against');
         }
+
+        // validate schema against whatever is defined in $validationSchema->$schema. If no
+        // schema is defined, assume self::DEFAULT_SCHEMA_SPEC (currently draft-04).
+        if ($this->factory->getConfig(self::CHECK_MODE_VALIDATE_SCHEMA)) {
+            if (!$this->getTypeCheck()->isObject($validationSchema)) {
+                throw new RuntimeException('Cannot validate the schema of a non-object');
+            }
+            if ($this->getTypeCheck()->propertyExists($validationSchema, '$schema')) {
+                $schemaSpec = $this->getTypeCheck()->propertyGet($validationSchema, '$schema');
+            } else {
+                $schemaSpec = self::DEFAULT_SCHEMA_SPEC;
+            }
+
+            // get the spec schema
+            $schemaStorage = $this->factory->getSchemaStorage();
+            if (!$this->getTypeCheck()->isObject($schemaSpec)) {
+                $schemaSpec = $schemaStorage->getSchema($schemaSpec);
+            }
+
+            // save error count, config & subtract CHECK_MODE_VALIDATE_SCHEMA
+            $initialErrorCount = $this->numErrors();
+            $initialConfig = $this->factory->getConfig();
+            $initialContext = $this->factory->getErrorContext();
+            $this->factory->removeConfig(self::CHECK_MODE_VALIDATE_SCHEMA | self::CHECK_MODE_APPLY_DEFAULTS);
+            $this->factory->addConfig(self::CHECK_MODE_TYPE_CAST);
+            $this->factory->setErrorContext(Validator::ERROR_SCHEMA_VALIDATION);
+
+            // validate schema
+            try {
+                $this->check($validationSchema, $schemaSpec);
+            } catch (\Exception $e) {
+                if ($this->factory->getConfig(self::CHECK_MODE_EXCEPTIONS)) {
+                    throw new InvalidSchemaException('Schema did not pass validation', 0, $e);
+                }
+            }
+            if ($this->numErrors() > $initialErrorCount) {
+                $this->addError(ConstraintError::INVALID_SCHEMA(), $path);
+            }
+
+            // restore the initial config
+            $this->factory->setConfig($initialConfig);
+            $this->factory->setErrorContext($initialContext);
+        }
+
+        // validate element against $validationSchema
+        $this->checkUndefined($element, $validationSchema, $path, $i);
     }
 }
--- a/src/JsonSchema/Validator.php
+++ b/src/JsonSchema/Validator.php
@@ -26,6 +26,11 @@ class Validator extends BaseConstraint
 {
     const SCHEMA_MEDIA_TYPE = 'application/schema+json';
 
+    const ERROR_NONE                    = 0x00000000;
+    const ERROR_ALL                     = 0xFFFFFFFF;
+    const ERROR_DOCUMENT_VALIDATION     = 0x00000001;
+    const ERROR_SCHEMA_VALIDATION       = 0x00000002;
+
     /**
      * Validates the given data against the schema and returns an object containing the results
      * Both the php object and the schema are supposed to be a result of a json_decode call.
@@ -60,6 +65,8 @@ class Validator extends BaseConstraint
         $this->factory->setConfig($initialCheckMode);
 
         $this->addErrors(array_unique($validator->getErrors(), SORT_REGULAR));
+
+        return $validator->getErrorMask();
     }
 
     /**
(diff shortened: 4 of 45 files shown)


erayd:
Is using `$schema` alone to determine the spec sufficient, or would a user-override via `$checkMode` to specifically set draft-03 / draft-04 etc. also be useful?

shmax:
👍 for the idea. I wish I had had this back when I was getting started; would have saved loads of development time and confusion.

 I would check `$schema` to start, then default to draft 4.

erayd:
@shmax Done - it now defaults to draft-04, rather than the previous behavior of throwing a `RuntimeException` if `$schema` was undefined.

shmax:
I haven't tried out the code yet, but what does the error output look like? Is it clear that any errors on the schema itself are related to the schema, and not the input data?

erayd:
@shmax It currently brings across all schema validation errors using the standard format, and then adds another error to the stack stating that schema validation has failed. Do you think that this is sufficient, or would some kind of additional disambiguation on the errors that result from schema validation be worth having?

shmax:
I'm not sure, I guess. I'll grab your code tonight and try it out and see how it feels...

erayd:
@shmax Sounds good. I have no particular preference either way, but adding some disambiguation context seems like a good idea, so I've done this (prefixes the error messages with `Schema validation: `. The old behavior was to simply use the error message verbatim.

shmax:
That might be good enough, but I'm wondering if there should be an even stronger separation. Does the validation proceed if schema errors are encountered, or does it stop dead?

erayd:
It currently proceeds - I figured that as the current behavior seems to be to proceed despite errors, I should follow suit with the schema validation stuff as well (plus `CHECK_MODE_EXCEPTIONS` is a thing, so folks who want it to bail at the first hurdle can easily obtain that behavior).

erayd:
What kind of stronger separation did you have in mind?

shmax:
I don't know, I don't have the code in front of me and I can't even really remember offhand what normal error output looks like, but I'd like to see what approach [AJV](https://github.com/epoberezkin/ajv) takes before I send you off on any wild goose chases.

erayd:
Have just fixed a URL validation bug (non-compliance with RFC-3986 relative references) that showed up when I applied schema validation to the draft unit tests. Is that fix (b415918) OK as part of this PR, or should I split it out into a separate PR of its own?

shmax:
If it's unrelated it's probably better to branch it off to a new PR, so we can give each issue our full attention without getting all mixed-up.

erayd:
@shmax What do you consider unrelated? I'm not sure whether it should be considered related or not. It's not part of the feature, but some of the draft unit tests will fail when using schema validation until the bug is fixed.

Splitting them will mean rebasing this on the bugfix PR until such a time as it's merged. Quite happy to do this if you're wanting them separate.

shmax:
I have no strong opinion, it just sounded like something that would branch its own discussion. Makes no real difference to me.

erayd:
I'll split it then; if you think people will want to have a big discussion about it, then it makes sense to keep them separate :-).

erayd:
OK, split done - this PR is now based on PR #358.

shmax:
Sorry for the delay. I think I'm up to speed, now. `validate` returns an array of errors; I was hoping there would be an object with an `errors` property, but no such luck. I also read up on ajv, and analyzed out how it handles this kind of thing. The short version is that it doesn't mix schema validation errors with regular errors (which I kind of suspected). I've prepared a jsfiddle that you can experiment with that demonstrates the general idea:

https://jsfiddle.net/hahmvsfj/2/

I don't want to force any ideas down your throat or anything just yet, but I thought it might give you some ideas. Let me know what you think.

erayd:
...would you be keen for me to overhaul the way errors work across the whole library? I don't particularly like the way they are done currently, and would *love* to tackle the issue, provided people are open to the idea - I don't want to spend ages working on something if nobody wants it.

shmax:
Well, I for one am a big fan of your tinkerings and would love to see what you come up with. Have anything in particular in mind right off the bat?

shmax:
Oh, and if you think this is going to be a really big project and you want to get some buy-in ahead of time, you might want to start with an issue and see if you can get some thumbs (up) on it.

erayd:
I don't have any really solid ideas yet (although I have a few inklings), but I do know that I dislike the current model and want to change it. I might have a bit of a play, then open a PR with 'these are my ideas' code so others can comment, before I sink large amounts of time into it. Then once we have agreement on an approach, I can polish it up. I think the defaults thing / checkmode refactor benefited greatly from input other than mine, particularly yours, and ended up much better as a result.

erayd:
@mathroc - do you have an opinion on whether or not this schema-validation PR should be a thing? Saw your comment on #353.

bighappyface:
Thoughts:

* Spec validation is valuable
* Spec validation and constraint validation make sense as separate operations
* Overhauling errors has happened before and probably didn't go far enough; however, it would be best to handle in a dedicated PR. See #239 for more on this topic.

With another review I will merge #358 so you can rebase this work.

bighappyface:
#358 is merged. Please rebase.

erayd:
@bighappyface Rebased.

src/JsonSchema/Constraints/SchemaConstraint.php
```diff
@@ -20,23 +23,70 @@
...
+                'http://json-schema.org/draft-03/schema' => 'json-schema-draft-03.json',
+                'http://json-schema.org/draft-03/schema#' => 'json-schema-draft-03.json',
+                'http://json-schema.org/draft-04/schema' => 'json-schema-draft-04.json',
+                'http://json-schema.org/draft-04/schema#' => 'json-schema-draft-04.json'
+            );
+            $schemaStorage = $this->factory->getSchemaStorage();
+            foreach ($preload as $schemaID => $schemaFile) {
+                $schemaStorage->addSchema(
+                    $schemaID,
+                    json_decode(file_get_contents(__DIR__ . "/../../../schema-validation/$schemaFile"))
```
bighappyface: Should we add a constant somewhere that gives us the package root for needs like this?

erayd: Yes - I think that is an excellent idea. Do you have a preference for which class it's attached to?

bighappyface: Can a class constant be a dynamic value? The last time I checked it has to be a scalar and not the result of an expression.

shmax: You can do simple scalar expressions ["involving numeric and string literals and/or constants"](http://php.net/manual/en/language.oop5.constants.php), but only as of 5.6.

erayd: Ooh, good point - forgot that restriction. Would a namespaced constant not attached to a class be acceptable in this case?

shmax: I don't totally understand where you guys are going with this, but you could do something like:

```PHP
class SchemaConstraint {
   const VALIDATION_ROOT = "/../../../schema-validation/%s";
///
}

json_decode(file_get_contents(__DIR__ . sprintf(SchemaConstraint::VALIDATION_ROOT, $schemaFile")))
```

erayd: No. The problem is the inability to use scalar expressions < 5.6, which means we have to use a different mechanism. We don't know what the working directory is, which means we need to use a path relative to `__DIR__` (your example is missing this). Using `__DIR__` implies string concatenation, which implies an expression.

What we're trying to achieve is some kind of global mechanism that provides the root filesystem path of the json-schema package, for further use by whatever else might need to know that, rather than manually constructing the path wherever it's used.

src/JsonSchema/Constraints/Factory.php
```diff
@@ -82,6 +87,23 @@ public function __construct(
     }
 
     /**
+     * Get package files
+     *
+     * @param string File path, relative to the root of the package
+     */
+    public function getPackageFile($path)
```
erayd: @bighappyface @shmax Is this an acceptable solution to the getting-package-files thing, or would you rather keep things simpler?

bighappyface:
Works for me :smille:

src/JsonSchema/Constraints/Factory.php
shmax: Looks like it would work, but isn't this repeating some of the work done in `SchemaStorage`?

erayd: Can you clarify what you think it's duplicating? I don't see anything in`SchemaStorage` that would allow me to achieve the same result - as far as I can see, the only similarity is that they both cache stuff.

shmax: Well, the class brings in some helpers for resolving paths. It also resolves any `$refs` encountered, and yes, stores the schemas it finds. Not sure how you would integrate one for your purpose here, but it I figure it merits a look.

erayd: Hmm, the `UrlResolver` has possibilities. Stand by...

mathroc:
@erayd thx, I didn't noticed the associated PR. I don't mind having this or not in the lib: I'm already validating the schemas as part of my tests (the same way I told in the comment in the issue) so I don't want to validate them again when the code is running on production.

however I have a few questions about the implementation (code review comming ☺)

erayd:
@mathroc Yeah, the PR won't validate them by default - you'll have to explicitly turn it on if you want that behavior.

Implementation is fluid at present - feedback is welcomed, and I'm quite happy to change stuff.

src/JsonSchema/Constraints/SchemaConstraint.php
```diff
@@ -20,23 +23,70 @@
...
+
+            // preload standard schema specs
+            $preload = array(
+                'http://json-schema.org/draft-03/schema' => 'json-schema-draft-03.json',
+                'http://json-schema.org/draft-03/schema#' => 'json-schema-draft-03.json',
+                'http://json-schema.org/draft-04/schema' => 'json-schema-draft-04.json',
+                'http://json-schema.org/draft-04/schema#' => 'json-schema-draft-04.json'
+            );
+            $schemaStorage = $this->factory->getSchemaStorage();
+            foreach ($preload as $schemaID => $schemaFile) {
```
mathroc: couldn't that be moved inside the SchemaStorage class ? it would benefit everyone validating schemas (and protect us if something like https://github.com/json-schema-org/json-schema-org.github.io/issues/48 happen again)
for the record I now have a local copy of the http://json-schema.org/draft-04/schema# and would like to remove that if it's handled at the library level

erayd: Yes, absolutely.

src/JsonSchema/Constraints/SchemaConstraint.php
```diff
@@ -20,23 +23,70 @@
...
+                $schemaStorage->addSchema(
+                    $schemaID,
+                    json_decode(file_get_contents(__DIR__ . "/../../../schema-validation/$schemaFile"))
+                );
+            }
+
+            // validate schema
+            $checkMode = $this->factory->getConfig(self::CHECK_MODE_EXCEPTIONS) | self::CHECK_MODE_TYPE_CAST;
```
mathroc: why not using the same `$checkMode` only without `CHECK_MODE_VALIDATE_SCHEMA` ?

erayd: To avoid complications while testing; I haven't thought through all the ramifications yet and what that would mean for possible future checkmode options that might come along. The current state of `$checkMode` here is temporary until I have considered it a bit more carefully.

src/JsonSchema/Constraints/SchemaConstraint.php
```diff
@@ -20,23 +23,70 @@
...
+            $validator = new Validator(new Factory($schemaStorage, null, $checkMode));
```
mathroc: what is the use of a new `Validator` ? is it to force a brand new `UriRetriever` ?

src/JsonSchema/Constraints/SchemaConstraint.php
```diff
@@ -20,23 +23,70 @@
...
+            $validator->validate($validationSchema, $schemaStorage->getSchema($schemaSpec));
+            if (!$validator->isValid()) {
+                $validationErrors = array_map(function ($error) {
+                    $error['message'] = 'Schema validation: ' . $error['message'];
```
mathroc: just wondering, if `CHECK_MODE_EXCEPTIONS` is `true` should the exception be wrapped with one with the same prefix ?

src/JsonSchema/Constraints/SchemaConstraint.php
erayd: It's to keep schema validation in a separate context at this stage. Open to using the same instances once I've looked at it more closely and can confirm it won't cause any problems.

src/JsonSchema/Constraints/SchemaConstraint.php
erayd: I'm intending a new exception specifically for schema validation failures.

src/JsonSchema/Constraints/SchemaConstraint.php
shmax: Ooh, what would happen if you used the same instance and validated the schema with schema validation turned on? 

https://www.mathsisfun.com/numbers/images/divide-by-zero.jpg

erayd:
Rebased.

src/JsonSchema/Constraints/SchemaConstraint.php
erayd: @shmax This is why setting `$checkMode` is important; it needs to be `~CHECK_MODE_VALIDATE_SCHEMA` before the actual schema validation takes place.

shmax: Yeah, I know. I was just having fun. I think that's how Captain Kirk used to defeat enemy computers.

erayd: > Set `$checkMode` to `CHECK_MODE_STUN`!

shmax: hee hee

src/JsonSchema/Constraints/SchemaConstraint.php
erayd: @mathroc @shmax @bighappyface Thoughts on #362?

shmax:
I was just wondering why you decided to use the same instance when validating the schema. Memory reasons? Also, it looks like you unset some of the checkmode options, but do you ever restore them?

erayd:
Not just memory - also in case people have dont custom subclassing or whatever to modify behavior... this way, everything is consistent.

Yes, checkmode is restored - see the call to `setCheckMode()` lower down.

erayd:
What are your thoughts on this PR now?

shmax:
Looking good, but I would still feel better about it if you could find a way to physically separate the schema errors from the regular errors. You go so far as to make a distinction with your exception type, which is good, but people who prefer not to use exceptions--as far as I can tell--don't have any way to programmatically act on schema errors.

erayd:
@shmax No problem - what about now? Errors now have a new 'context' field, which defaults to 'document'. If the error occurs during schema validation, then this field will be 'schema'. This will allow anyone post-processing errors to easily determine where each one comes from.

`Factory::getContext()` and `Factory::setContext()` can be used to get or change the current context.

shmax:
Hmm. It's a start, and I can live with it, but it still feels a little wishy-washy to me, because if someone wants to know if there were any schema errors he has to walk over all the errors and peek into each one. That would work, but it feels a little passive, you know?

Just to brainstorm a few ideas (and some of these will probably be more appropriate for 6.0):

1. What about adding a separate `getSchemaErrors` method? A user that has deliberately turned on the schema validation feature would know to check this before checking `getErrors`.

2. I notice that `validate` doesn't return anything. What if it returned an error code? `0` for clean, `1` for schema errors, `2` for regular errors, `3` for both, etc. Then calling `getErrors` becomes more like an optional forensic task. A user could check the return value, then call one of `getErrors` or `getSchemaErrors` accordingly. That would achieve rough parity with something like PHP's `json_decode`  (where you can call `json_last_error` after the fact), so there's some precedent for that kind of thing.

Well, maybe that's enough for now. I'll see if I can think of any more in the morning.

erayd:
@shmax Aah, I think I misunderstood you then - I thought you were wanting to address a case when a user was _already_ walking the errors and wanted to filter in / out schema ones.

 1) Would adding a context argument to getErrors be sufficient here (and it only returns errors in the given context, if a context is provided, otherwise return them all)? Or would you prefer that the schema errors actually be catalogued separately? I am happy with either - if you let me know your preference, I'll implement it.
 2) I **really** like that idea. Having validate return an error bitmask with 0 as 'all OK' allows for a whole lot of useful scenarios. Will implement it.

shmax:
Yeah, I think passing an argument to `getErrors` is better, as you say. It could even be another bitmask type of thing (eg. `getErrors($type = Validator::ERRORS_ALL)`). As for the context thing, hmm. I guess it wouldn't be necessary to expose that to the user once he has the ability to fish one or the other (or both) out of the validator with `getErrors`, so the question becomes how do you want to do it internally? No strong opinion is dawning on me, so I'll leave it up to you to mull while you're doing the other stuff.

erayd:
@shmax How's that?

I want to go over this PR again at some point and look for messy code too. I've made lots of changes, and I need to make sure there's no remnants of previous approaches to stuff hanging around. Please let me know if you spot anything you think needs to go.

shmax:
Looking sharp, but I'll have to get back to you tomorrow evening after work. Nice work.

erayd:
Thanks :-). No rush.

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -54,18 +61,42 @@ public function addError(JsonPointer $path = null, $message, $constraint = '', a
...
+            array_walk($errors, function ($error) use (&$errorMask) {
+                if (isset($error['context'])) {
+                    $errorMask |= $error['context'];
+                }
+            });
         }
     }
 
-    public function getErrors()
+    public function getErrors($errorContext = null)
```
shmax: Just a quick comment before I head to work; can we have a more meaningful default value? It would make it a little more clear to the reader how this method works. Something like:

```
public function getErrors($errorContext = Validator::ERROR_ALL);
```

erayd: Good idea - done :-).

src/JsonSchema/Validator.php
```diff
@@ -26,6 +26,7 @@ class Validator extends BaseConstraint
     const SCHEMA_MEDIA_TYPE = 'application/schema+json';
 
     const ERROR_NONE                    = 0x00000000;
+    const ERROR_ALL                     = ~self::ERROR_NONE;
```
shmax: hah, clever

erayd: Better than using a single bit IMO, because it still works if you use it as a mask :-).

erayd: That way you can do fancy things like fetching errors for `(Validator::ERROR_ALL & ~Validator:ERROR_I_NEVER_WANT_TO_SEE)`. We only have two error contexts right now, but if that changes later it'll come in handy.

erayd: Forgot that *&@^$ no-expressions-in-constant-definitions thing though. Have changed it to 0xFFFFFFFF, which achieves exactly the same thing.

shmax: Yeah, being not clever, I would have just typed in `0xffffff` from the start.

shmax:
LGTM very nice

erayd:
@shmax Thanks :-). Time to revisit #362 now we've sat on it for a while? This PR (and one other) still depends on it.

erayd:
Junk commits squashed, ready to merge after #362 (URI translation & local spec bundling).

erayd:
Rebased.

erayd:
Rebased.

erayd:
Rebased.

PR #358, opened 2017-02-22
[BUGFIX] RFC-3986 compliant validation of URI relative references

erayd:
Adds some hackery to properly validate RFC-3986 relative references, as PHP's `FILTER_VALIDATE_URL` is based on an older RFC and can't handle them.


shmax:
Not sure that I totally understand this one, but I see no obvious black holes, so 👍

PR #360, opened 2017-02-23
[BUGFIX] Fix relative-path case where first segment contains a colon

erayd:
Resolves minor issue with previous bugfix #358:
 * The first segment of a relative-path reference may not contain a colon
 * The previous bugfix allowed colons in the first segment if it started with a '.'


shmax:
LGTM

PR #361, opened 2017-02-23
[BUGFIX] Don't clobber $i

erayd:
While using `$i` here follows convention, assigning to it here can occasionally clobber an existing value. It doesn't cause any testsuite errors, but is still undesirable behavior, and should not be used this way.

This PR renames the variable used within the set-defaults section to avoid having `foreach()` overwrite it.


shmax:
It looks fine, but it's not immediately clear to me which existing value you're referring to. Something earlier in the file?

erayd:
Yep. There is already a variable named `$i` elsewhere in `validateCommonProperties()`, so I shouldn't be naming my completely unrelated variable `$i`, as doing so means that whatever somebody else put in `$i` before I came along gets replaced when `foreach()` runs.

shmax:
Ah, yes, it shadows the $i parameter. LGTM

shmax:
(come to think of it, it's pretty hard to think of a worse parameter name to be sprinkling through the code base than `$i`)

erayd:
Indeed. It's nice and concise though... but not particularly descriptive, and conflicts with the conventional programming use of `$i` as a temporary loop variable.

PR #364, opened 2017-02-26
centralize errors

shmax:
Some improvements to the error system in an effort to make things a little easier for people interested in adding localization, such as https://github.com/justinrainbow/json-schema/issues/363:

* Added a new class: `ConstraintError`. This extends an `enum` class which I've brought into the project to ensure that nobody types constraint types or messages directly into the code (by typehinting `addError`). I've used it before in a number of projects and have found it quite useful. 

* Converted all error strings into a format that can be passed to `sprintf`.

* I had to invent a few new format error types to avoid ambiguity: `phoneFormat`, `ipFormat`, `styleFormat`, `timeFormat`, `urlFormat`. 

@bighappyface @erayd @harej

### DIFF
--- a/src/JsonSchema/ConstraintError.php
+++ b/src/JsonSchema/ConstraintError.php
@@ -95,6 +95,10 @@ class ConstraintError extends \MabeEnum\Enum
             self::UNIQUE_ITEMS => 'There are no duplicates allowed in the array'
         );
 
-        return isset($messages[$name]) ? $messages[$name] : "Unknown Error: $name";
+        if (!isset($messages[$name])) {
+            throw new InvalidArgumentException('Missing error message for ' . $name);
+        }
+
+        return $messages[$name];
     }
 }
--- a/src/JsonSchema/ConstraintError.php
+++ b/src/JsonSchema/ConstraintError.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace JsonSchema;
+
+class ConstraintError extends \MabeEnum\Enum
+{
+    const ADDITIONAL_ITEMS = 'additionalItems';
+    const ADDITIONAL_PROPERTIES = 'additionalProp';
+    const ALL_OF = 'allOf';
+    const ANY_OF = 'anyOf';
+    const DEPENDENCIES = 'dependencies';
+    const DISALLOW = 'disallow';
+    const DIVISIBLE_BY = 'divisibleBy';
+    const ENUM = 'enum';
+    const EXCLUSIVE_MINIMUM = 'exclusiveMinimum';
+    const EXCLUSIVE_MAXIMUM = 'exclusiveMaximum';
+    const FORMAT_COLOR = 'colorFormat';
+    const FORMAT_DATE = 'dateFormat';
+    const FORMAT_DATE_TIME = 'dateTimeFormat';
+    const FORMAT_DATE_UTC = 'dateUtcFormat';
+    const FORMAT_EMAIL = 'emailFormat';
+    const FORMAT_HOSTNAME = 'styleHostName';
+    const FORMAT_IP = 'ipFormat';
+    const FORMAT_PHONE = 'phoneFormat';
+    const FORMAT_REGEX= 'regexFormat';
+    const FORMAT_STYLE = 'styleFormat';
+    const FORMAT_TIME = 'timeFormat';
+    const FORMAT_URL = 'urlFormat';
+    const LENGTH_MAX = 'maxLength';
+    const LENGTH_MIN = 'minLength';
+    const MAXIMUM = 'maximum';
+    const MIN_ITEMS = 'minItems';
+    const MINIMUM = 'minimum';
+    const MISSING_MAXIMUM = 'missingMaximum';
+    const MISSING_MINIMUM = 'missingMinimum';
+    const MAX_ITEMS = 'maxItems';
+    const MULTIPLE_OF = 'multipleOf';
+    const NOT = 'not';
+    const ONE_OF = 'oneOf';
+    const REQUIRED = 'required';
+    const REQUIRED_D3 = 'selfRequired';
+    const REQUIRES = 'requires';
+    const PATTERN = 'pattern';
+    const PREGEX_INVALID = 'pregrex';
+    const PROPERTIES_MIN = 'minProperties';
+    const PROPERTIES_MAX = 'maxProperties';
+    const TYPE = 'type';
+    const UNIQUE_ITEMS = 'uniqueItems';
+
+    public function getMessage()
+    {
+        $name = $this->getValue();
+        static $messages = array(
+            self::ADDITIONAL_ITEMS => 'The item %s[%s] is not defined and the definition does not allow additional items',
+            self::ADDITIONAL_PROPERTIES => 'The property %s is not defined and the definition does not allow additional properties',
+            self::ALL_OF => 'Failed to match all schemas',
+            self::ANY_OF => 'Failed to match at least one schema',
+            self::DEPENDENCIES => '%s depends on %s, which is missing',
+            self::DISALLOW => 'Disallowed value was matched',
+            self::DIVISIBLE_BY => 'Is not divisible by %d',
+            self::ENUM => 'Does not have a value in the enumeration %s',
+            self::EXCLUSIVE_MINIMUM => 'Must have a minimum value greater than %d',
+            self::EXCLUSIVE_MAXIMUM => 'Must have a maximum value less than %d',
+            self::FORMAT_COLOR => 'Invalid color',
+            self::FORMAT_DATE => 'Invalid date %s, expected format YYYY-MM-DD',
+            self::FORMAT_DATE_TIME => 'Invalid date-time %s, expected format YYYY-MM-DDThh:mm:ssZ or YYYY-MM-DDThh:mm:ss+hh:mm',
+            self::FORMAT_DATE_UTC => 'Invalid time %s, expected integer of milliseconds since Epoch',
+            self::FORMAT_EMAIL => 'Invalid email',
+            self::FORMAT_HOSTNAME => 'Invalid hostname',
+            self::FORMAT_IP => 'Invalid IP address',
+            self::FORMAT_PHONE => 'Invalid phone number',
+            self::FORMAT_REGEX=> 'Invalid regex format %s',
+            self::FORMAT_STYLE => 'Invalid style',
+            self::FORMAT_TIME => 'Invalid time %s, expected format hh:mm:ss',
+            self::FORMAT_URL => 'Invalid URL format',
+            self::LENGTH_MAX => 'Must be at most %d characters long',
+            self::LENGTH_MIN => 'Must be at least %d characters long',
+            self::MAX_ITEMS => 'There must be a maximum of %d items in the array',
+            self::MAXIMUM => 'Must have a maximum value less than or equal to %d',
+            self::MIN_ITEMS => 'There must be a minimum of %d items in the array',
+            self::MINIMUM => 'Must have a minimum value greater than or equal to %d',
+            self::MISSING_MAXIMUM => 'Use of exclusiveMaximum requires presence of maximum',
+            self::MISSING_MINIMUM => 'Use of exclusiveMinimum requires presence of minimum',
+            self::MULTIPLE_OF => 'Must be a multiple of %d',
+            self::NOT => 'Matched a schema which it should not',
+            self::ONE_OF => 'Failed to match exactly one schema',
+            self::REQUIRED => 'The property %s is required',
+            self::REQUIRED_D3 => 'Is missing and it is required',
+            self::REQUIRES => 'The presence of the property %s requires that %s also be present',
+            self::PATTERN => 'Does not match the regex pattern %s',
+            self::PREGEX_INVALID => 'The pattern %s is invalid',
+            self::PROPERTIES_MIN => 'Must contain a minimum of %d properties',
+            self::PROPERTIES_MAX => 'Must contain no more than %d properties',
+            self::TYPE => '%s value found, but %s is required',
+            self::UNIQUE_ITEMS => 'There are no duplicates allowed in the array'
+        );
+
+        return isset($messages[$name]) ? $messages[$name] : "Unknown Error: $name";
+    }
+}
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -37,7 +37,7 @@ class BaseConstraint
         $this->factory = $factory ?: new Factory();
     }
 
-    public function addError(JsonPointer $path = null, ConstraintError $constraint = null, array $more = array())
+    public function addError(ConstraintError $constraint, JsonPointer $path = null, array $more = array())
     {
         $message = $constraint ? $constraint->getMessage() : '';
         $name = $constraint ? $constraint->getValue() : '';
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -49,15 +49,14 @@ class BaseConstraint
 
                 return json_encode($val);
             }, array_values($more)))),
-            'constraint' => $constraint ? $constraint->getValue() : '',
+            'constraint' => array(
+                'name' => $constraint ? $constraint->getValue() : '',
+                'params' => $more
+            )
         );
 
         if ($this->factory->getConfig(Constraint::CHECK_MODE_EXCEPTIONS)) {
             throw new ValidationException(sprintf('Error validating %s: %s', $error['pointer'], $error['message']));
-        }
-
-        if (is_array($more) && count($more) > 0) {
-            $error += $more;
         }
 
         $this->errors[] = $error;
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -39,10 +39,12 @@ class BaseConstraint
 
     public function addError(JsonPointer $path = null, ConstraintError $constraint = null, array $more = array())
     {
+        $message = $constraint ? $constraint->getMessage() : '';
+        $name = $constraint ? $constraint->getValue() : '';
         $error = array(
             'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
             'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
-            'message' => ucfirst(vsprintf($constraint->getMessage(), array_map(function ($val) {
+            'message' => ucfirst(vsprintf($message, array_map(function ($val) {
                 if (is_scalar($val)) {
                     return $val;
                 }
@@ -50,7 +52,7 @@ class BaseConstraint
                 return json_encode($val);
             }, array_values($more)))),
             'constraint' => array(
-                'name' => $constraint ? $constraint->getValue() : '',
+                'name' => $name,
                 'params' => $more
             )
         );
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -9,6 +9,7 @@
 
 namespace JsonSchema\Constraints;
 
+use JsonSchema\ConstraintError;
 use JsonSchema\Entity\JsonPointer;
 use JsonSchema\Exception\ValidationException;
 
@@ -36,13 +37,19 @@ class BaseConstraint
         $this->factory = $factory ?: new Factory();
     }
 
-    public function addError(JsonPointer $path = null, $message, $constraint = '', array $more = null)
+    public function addError(JsonPointer $path = null, ConstraintError $constraint = null, array $more = array())
     {
         $error = array(
             'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
             'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
-            'message' => $message,
-            'constraint' => $constraint,
+            'message' => ucfirst(vsprintf($constraint->getMessage(), array_map(function ($val) {
+                if (is_scalar($val)) {
+                    return $val;
+                }
+
+                return json_encode($val);
+            }, array_values($more)))),
+            'constraint' => $constraint ? $constraint->getValue() : '',
         );
 
         if ($this->factory->getConfig(Constraint::CHECK_MODE_EXCEPTIONS)) {
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,8 @@
         }
     }],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
+        "marc-mabe/php-enum":"2.3.1"
     },
     "require-dev": {
         "json-schema/JSON-Schema-Test-Suite": "1.2.0",
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -9,6 +9,7 @@
 
 namespace JsonSchema\Constraints;
 
+use JsonSchema\ConstraintError;
 use JsonSchema\Entity\JsonPointer;
 
 /**
@@ -26,12 +27,12 @@ class CollectionConstraint extends Constraint
     {
         // Verify minItems
         if (isset($schema->minItems) && count($value) < $schema->minItems) {
-            $this->addError($path, 'There must be a minimum of ' . $schema->minItems . ' items in the array', 'minItems', array('minItems' => $schema->minItems));
+            $this->addError($path, ConstraintError::MIN_ITEMS(), array('minItems' => $schema->minItems));
         }
 
         // Verify maxItems
         if (isset($schema->maxItems) && count($value) > $schema->maxItems) {
-            $this->addError($path, 'There must be a maximum of ' . $schema->maxItems . ' items in the array', 'maxItems', array('maxItems' => $schema->maxItems));
+            $this->addError($path, ConstraintError::MAX_ITEMS(), array('maxItems' => $schema->maxItems));
         }
 
         // Verify uniqueItems
@@ -43,7 +44,7 @@ class CollectionConstraint extends Constraint
                 }, $value);
             }
             if (count(array_unique($unique)) != count($value)) {
-                $this->addError($path, 'There are no duplicates allowed in the array', 'uniqueItems');
+                $this->addError($path, ConstraintError::UNIQUE_ITEMS());
             }
         }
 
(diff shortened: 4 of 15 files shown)


erayd:
@shmax Thanks for tackling this :-). I'm currently sick with some kind of nasty bug and can't think straight, but will try to review this properly in the next couple of days.

In the meantime, quick thoughts after glancing at it:
 * Why are you depending on an unstable branch of php-enum? Is there a reason you can't use a stable release?
 * Need to run `composer style-fix`; the travis builds are currently failing with code style errors.
 * If you're changing the API anyway, then could `$path` be bundled in with `ConstraintError`? If we're making a new class for this, might as well make sure it has all the information.
 * Same point as above re additional format-specific error information.
* I like the syntax for selecting an error to use - assume `php-enum` is using `__callStatic()`? Good way of doing it.
 * Does `php-enum` allow extending the class and only defining a subset of new errors, and falling back to the parent definitions for missing stuff? That's really important - need it to avoid breaking people's translations whenever a new error is added. Your current approach requires the user to manage falling back to `parent::getMessage()` manually in some way, which seems like asking for trouble later when somebody doesn't think to do it.

shmax:
I'll squash in a bit... I couldn't get the style checker to work locally, so I've been letting travis figure it out. Consider it a wip for now. I'm out and about at the moment, but will resume in the morning.

erayd:
Gotcha. What's the style-checker doing? It should run on everything but nightly / hhvm; if it's not working on your system then something rather strange is going on.

erayd:
You can also get an executable phar [here](http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar), which means you can run it without needing to go through composer - if you're having issues with the depended one, this may be a suitable temporary workaround.

shmax:
It's something to do with line endings. I'm on Windows, and I've got git configured to sort them out when pushing and pulling, but apparently the style checker isn't as savvy--it just aborts right away, and doesn't get as far as doing any style analysis.

shmax:
> Why are you depending on an unstable branch of php-enum? Is there a reason you can't use a stable release?
No reason. Updated to latest release.

> If you're changing the API anyway, then could $path be bundled in with ConstraintError? If we're making a new class for this, might as well make sure it has all the information.
 Not sure what you're suggesting. `$path` is dynamic.

> Same point as above re additional format-specific error information.
Not following.

> That's really important - need it to avoid breaking people's translations whenever a new error is added. 

I hadn't really gotten as far as deciding how to advise people wanting to localize. We could provide extra hooks, but really they could just do as @bighappyface suggests and post-process errors as they're caught. Definitely don't expect people to subclass `ConstraintError`, at least not in its current state. The only real difference is that now the meaning of `$constraint` is no longer ambiguous, and we have a clear, easy to read mapping of error strings to error codes in once place for reference.

As far as breaking their strings goes, well, I guess I don't follow that, either. I wouldn't expect adding new strings to break anything--they just wouldn't be localized until the client does something about them. We'll have to consider existing `$constraint` values to be part of the external API going forward, and be careful to update the release version accordingly if they ever do change.

shmax:
I did realize that with the old error layout it would be difficult (though not impossible) for localizers to extract the arguments for their translation calls; everything was just mixed into the same error object. I've restructured `constraint` to be an array with keys for `name` and `params`, which should sidestep the difficulty. It will break bc for anyone that was doing anything programmatic with the fields in the error arrays, unfortunately.

shmax:
@bighappyface @erayd So what do you think?

erayd:
@shmax I'm still sick, so I can't promise the world's best feedback, but I will take a decent look at this today and let you know.

shmax:
No rush. What about you, @bighappyface? You seemed to have some opinions about this kind of thing...

bighappyface:
While I am not opposed to the approach per se I am concerned by how loose we have become with this package and breaking changes.

Simply put, changing the major version of this package every third PR that comes in is frustrating to me and unfair to consumers. We have to do a better job at trying to make things better without redesigning them every time.

That being said, I ask that we start road-mapping releases and embrace the idea of breaking the API later while maintaining BC now. If we are to break the API then we should set goals for what needs to be changed and bundle them together as part of a plan and not a side-effect of any given PR. This will help us vet our changes and be far more prudent in determining what needs to happen, and how to get there, with due dilligence.

In this case that means not altering the method signature for `addError` nor it's return value array structure for the `constraint` element. Those changes can come as the second phase to this work, in another major release, where this first phase centralizes errors, establishes constants, and basically ends there.

bighappyface:
To help on this effort I have linked this PR and the original issue to a new 5.2.0 project:

https://github.com/justinrainbow/json-schema/projects/1

shmax:
I can see the wisdom in not altering the layout of the `error` object, but what's the harm in changing the `addError` signature? Do you consider that part of the public API?

erayd:
@bighappyface I agree very strongly with everything you just said in those last two posts.

Not sure how much of the discussion around the v5 refactor you read (it started in the defaults PR and then moved to its own thing), but I was very much in favour of not breaking things, and decided to lose that fight, because I'm new here and the impression I was getting was that maintaining backwards compatibility was at best a minor concern, and you guys didn't care nearly as much about it as I wanted to - the understanding I took from those conversations was that, as far as this library is concerned, minimising the number of public methods on `Validator` was preferable to maintaining backwards compatibility. I don't regret the code that resulted (although I'm very pleased that the `check` / `coerce` aliases on `Validator` ended up back in the mix), or the valuable contributions others here (particularly @shmax) made to help bring it to that final state, but if maintaining backwards compatibility in the API has now become a priority, or my original understanding was incorrect, I am very much in favor of that.

To that end, would it be possible to also have as a project goal properly documenting the public API? I'm happy to assist with the parts I'm familiar with, but I don't know enough about the intent behind the various parts to do much of it justice. Currently, `Validator` seems to be the only 'official' public interface, but then vast swathes of the rest of the API seem to be considered unofficially part of what users are intended to interface with - there is no clear line here.

erayd:
@shmax That's why I reckon the public API needs documenting (see above). It's not officially documented as part of the public API, but it seems to be treated like something an end user would interact with. Certainly `getErrors` is pretty obviously public by virtue of its intended use.

erayd:
@shmax Noting @bighappyface's request, would you still like me to do a full code-review of this PR today, or would you prefer to make changes first?

shmax:
Is there really that much to review? Here's the takeaway:

1. Moved error strings to `ConstraintError`, and did an accounting of all the possible values for `$constraint` (to eliminate ambiguity).
2. Changed the `addError` signature to take a `ConstraintError` (to enforce compliance with 1.)
3. Changed the format of the error object so that the constraint arguments can be easily extracted by someone wanting to pass them to their own localization mechanic.

That's pretty much it. @bighappyface disagrees with both 2. and 3., at least without some kind of long-term release plan, which is fine. I guess I'll just wait for further instruction from him.

erayd:
@shmax When I do code reviews, I like to unstable the full impact of every single changed line, and the full implications of any logic changed that result, whether intended or otherwise. This can take a while.

If all you're after is a review of the concept, then I'm very much in favour of tidying up the way errors are handled - I had a PR in the works myself following our discussion in the schema-validation PR, (although not yet at the point I'd be happy to submit it), however you beat me to it - you seem keen to tackle this task, and you doing it means I can keep focused on the schema validation (and that annoying recursion bug), so I'm all in favour of this one being yours :-).

Between my original review comment and what @bighappyface has said, there probably isn't too much more for me to say here unless you'd like a full code review, which I'm happy to do if you'd find it valuable, and was my original intention. If a code review *isn't* what you want, but you have specific questions or want feedback on something, please ask.

bighappyface:
@shmax sorry to see you close and delete. I don't think you needed to go that far.

May I suggest a variation?

```php
# before
$this->addError($path, 'There must be a minimum of ' . $schema->minItems . ' items in the array', 'minItems', array('minItems' => $schema->minItems));

# after
$this->addError($path, ConstraintError::MIN_ITEMS(), ConstraintError::MIN_ITEMS, array('minItems' => $schema->minItems));
```

This change would maintain the core of your contribution while also keeping the current method signature. I think we gain a great deal by centralizing and we can refactor `addError` in 6.0.0.

What do you think?

---

Regarding 5.0.0, you guys did a ton of work adding some necessary functionality that this package sorely needed to stay relevant. I felt your work merited a version, and your continued engagement and support merits much more. Now that a bit of a team is forming I think organizing will help us tremendously, and I hope to see this package continue to improve. Careful planning will focus work and help motivate efforts.

shmax:
```
$this->addError($path, ConstraintError::MIN_ITEMS(), ConstraintError::MIN_ITEMS, array('minItems' => $schema->minItems));
```

That's still a different signature. The second argument to `addError` as it stands now is currently a string, whereas in your sample the second argument is an instance of `ConstraintError` (which extends `Enum`). I guess I don't see the rationale in considering something like `addError` to be an inviolable part of the public API--by that logic we should never touch, add, or remove any public methods on any of the constraint classes, and I think that might be taking things a bit too far. 

I can go ahead and rearrange things a bit so that the original signature IS maintained (at least until 6.0) as you suggest, but we wouldn't have achieved the goal of helping the folks who want to localize without making the change to `error` output (we'd be closer, but they would be left in the position of having to write a lot of glue code to interpret each of the possible values for `error.constraint`).

Thanks for the encouragement. When do you see us doing a 6.0 release?

erayd:
@bighappyface - Works for me :-).

@shmax - I agree that your work is valuable, and closing this PR isn't necessary - with a few tweaks it can be made to work while still maintaining backwards-compatibility. it doesn't change the signature because the current approach doesn't define a type, so we can accept any argument we like there. It doesn't have to be a string.

I do think we need to seriously consider how people will extend the error messages to their own subset though - we can certainly come up with a viable approach for that without changing the signatures until v6. I don't think we need to actually create translations, but I do think we need to have a sensible approach that we can point to and tell users who want to translate to use, and it needs to be a method that allows people to only have a subset of the messages translated. IMO that translation should *not* be making the user interpret the output; it makes much more sense to bake some kind of way to apply translations into the message resolver you're busy writing here. That could be as simple as passing it an array of translations, and just merging it into the active messages array.

Regarding which parts of the error API are considered public - this is why I think we really need some better documentation regarding what is public API, and what is not. There's no clear line right now, and it's extremely frustrating trying to work on this without knowing what's safe to step on without breaking things for users.

shmax:
> it doesn't change the signature because the current approach doesn't define a type, so we can accept any argument we like there. It doesn't have to be a string.

Okay, fine, it's true that the signature doesn't technically change, but if we're going on the assumption that people are currently making their own calls to `addError`--which I admit is possible--and passing in strings in good faith (or overriding the method and doing something of their own with the incoming string before passing it down to the base), then their code will still break.

As far as the specifics on localization go, let's not make any assumptions. There are many different schemes, and not all of them involve typing away at the code (in fact I've never personally seen it done that way). More typically the translated strings come from a database, or generated resources that are maintained in a completely separate process. The changes I made in this PR did provide one crucial piece to help make them all possible, however, which was to separate the arguments that you pass along with the string from the rest of the information. As things stand now (in master), you would have to write a big crazy switch statement.

Anyway, I'll mull it over a bit more and see if I can come up with some kind of transitional compromise.

erayd:
>Okay, fine, it's true that the signature doesn't technically change, but if we're going on the assumption that people are currently making their own calls to addError--which I admit is possible--and passing in strings in good faith (or overriding the method and doing something of their own with the incoming string before passing it down to the base), then their code will still break.
It doesn't need to break if you add a simply typecheck to the method. If it's an instance of your constraint class, use the new behavior. Otherwise, stick with the old behavior (as far as the API is concerned), and handle it however seems appropriate.

I'm certainly curious to see what you come up with - overall I'm liking the direction your PR is heading in :-).

bighappyface:
I know that the previous work done on errors was work that those folks needed. Simply put, we can't really know what folks are using, so I think we have to assume whatever is public is in use, and plan with this assumption in mind for future versions.

I put a tentative note in the 6.0.0 project for refactoring the method but we can layout whatever milestones we wish to satisfy another major version. I simply want a plan and concensus we can execute on.

shmax:
Sounds good, though I was unable to find the note you mention (I'm not familiar with github's project feature).

erayd:
@bighappyface Makes sense. Are you willing to nominate a major version milestone - not too fussed which one, but would like *something* to work towards - as a point where the public API must be clearly defined, and anything that is *not* explicitly defined as part of the public API is allowed to break in point releases.

IMO having an issue for people to have some input on regarding what should be considered public API from that point would make a lot of sense, as we currently don't know where that line ought to be drawn.

erayd:
@shmax It's in Projects>>6.0.0>>To Do.

shmax:
> It doesn't need to break if you add a simple typecheck to the method

It would break if they were overriding the method in their subclass (because it would get called by our code, which passes in a `ConstraintError`) and doing something with what they believe to be a string. But if you guys are willing to waive that possibility, then it's okay with me.

shmax:
Ah, okay, I didn't see the 6.0 project. Got it, thanks.

erayd:
Hmm, that's a really good point. What if we just left all the old error infrastructure there for now, and built a completely parallel new one side-by-side that the library can use internally. Would need some way of merging the output at the end, so you can get at it from either interface, but it would solve a lot of the compatibility problems. Then we can just mark the old one deprecated, but avoid breaking anything, and it can go away at some future date.

shmax:
What do you think about the idea of creating a 6.0 branch now, and rebasing stuff like this against it? We can tinker away and even merge and we just leave master as-is until we're satisfied with 6.0.

erayd:
I like that idea.

erayd:
There's also nothing stopping us backporting in a compatible way from that branch if we can figure out a way to do it safely :-).

Downside is increased merge complexity by maintaining two heads, but I can live with that.

shmax:
True, it can get a little fiddly, but since there's only a small handful of us busily beavering away and we tend to favor exotic new features anyway (that would go mostly in 6.0) I don't think it would be anything we couldn't handle.

shmax:
Also, is there a general place where we can brainstorm 6.0 features?

erayd:
Are you talking IM (in which case maybe gitter), or something like a wiki?

shmax:
I was thinking more like an issue or a wiki; someplace where other people can chime in too, and we have a record of it.

erayd:
Will [this](https://github.com/justinrainbow/json-schema/wiki/Roadmap-Planning) do?

shmax:
Not sure? So we each just edit the same content? Would there be a discussion thread where we can go back and forth?

erayd:
That's what a wiki is. Any discussion-thread feature that *isn't* us editing the same content isn't a wiki, although many wiki apps do come with a discussion feature bolted on... if you're looking for a place to discuss with history, then maybe the issues here makes the most sense. It also has the added advantage of being very discoverable. The second we move away from github, most of the potential audience / participants vanish.

shmax:
Issue sounds good to me, at least for hashing ideas out; maybe we could shuttle the good ideas over to the wiki once they're a little more fully cooked.

erayd:
Works for me. Or we could shuttle them to Projects - I haven't used that feature before, but based on what @bighappyface has done, it looks like you can link issues from within the project interface.

shmax:
> What do you think about the idea of creating a 6.0 branch now, and rebasing stuff like this against it? We can tinker away and even merge and we just leave master as-is until we're satisfied with 6.0.

@bighappyface What do you think of this idea? If you like it, then I'll just rebase this branch against 6.0 and more or less leave it as-is. Otherwise, I'll modify this into a transitional form.

bighappyface:
I have no problem with a feature branch :)

https://github.com/justinrainbow/json-schema/tree/6.0.0-dev

erayd:
Thanks @bighappyface :-D

shmax:
@bighappyface = The Best

shmax:
Okay, open for business, again. I've changed the base to 6.0.0-dev, so go ahead and review.

src/JsonSchema/ConstraintError.php
```diff
@@ -0,0 +1,100 @@
...
+    const REQUIRED_D3 = 'selfRequired';
+    const REQUIRES = 'requires';
+    const PATTERN = 'pattern';
+    const PREGEX_INVALID = 'pregrex';
+    const PROPERTIES_MIN = 'minProperties';
+    const PROPERTIES_MAX = 'maxProperties';
+    const TYPE = 'type';
+    const UNIQUE_ITEMS = 'uniqueItems';
+
+    public function getMessage()
```
erayd: May we please also have a `setMessage` method that can add / overwrite messages, and store the messages on the class rather than inside `getMessage`? This would be useful for local translations and people's custom errors that never make it as far as upstream.

Would be good if this could accept either a single error message, or an array of them to allow setting multiple errors at once. e.g.:
```PHP
public static function setMessage($id, $message)
{
    if (is_array($id)) {
        self::$messages = array_merge(self::$messages, $id);
    } else {
        self::$messages[$id] = $message;
    }
}
```

src/JsonSchema/ConstraintError.php
```diff
@@ -0,0 +1,100 @@
...
+            self::REQUIRES => 'The presence of the property %s requires that %s also be present',
+            self::PATTERN => 'Does not match the regex pattern %s',
+            self::PREGEX_INVALID => 'The pattern %s is invalid',
+            self::PROPERTIES_MIN => 'Must contain a minimum of %d properties',
+            self::PROPERTIES_MAX => 'Must contain no more than %d properties',
+            self::TYPE => '%s value found, but %s is required',
+            self::UNIQUE_ITEMS => 'There are no duplicates allowed in the array'
+        );
+
+        return isset($messages[$name]) ? $messages[$name] : "Unknown Error: $name";
```
erayd: The unknown error should probably be in the error table with everything else - it seems neater, and there's nothing special about it to warrant being defined separately.

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -36,23 +37,28 @@ public function __construct(Factory $factory = null)
...
 
-    public function addError(JsonPointer $path = null, $message, $constraint = '', array $more = null)
+    public function addError(JsonPointer $path = null, ConstraintError $constraint = null, array $more = array())
     {
         $error = array(
             'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
             'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
-            'message' => $message,
-            'constraint' => $constraint,
+            'message' => ucfirst(vsprintf($constraint->getMessage(), array_map(function ($val) {
```
erayd: Don't use ucfirst here, it's not multibyte-safe, and not all languages have the concept of capital letters. Better IMO to define the messages with a capital in the first place, and let anyone providing custom or translation messages handle their own casing.

Also, why do all the format stuff here? More sensible IMO to have the constructor handle it; that way you just instantiate the error with its variable components and you get a finished package, rather than post-processing it as you're doing here.

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -36,23 +37,28 @@ public function __construct(Factory $factory = null)
         $this->factory = $factory ?: new Factory();
     }
 
-    public function addError(JsonPointer $path = null, $message, $constraint = '', array $more = null)
+    public function addError(JsonPointer $path = null, ConstraintError $constraint = null, array $more = array())
     {
         $error = array(
```
erayd: Rather than sticking with the array format, why not roll all this into the `ConstraintError` class? That way you can just pass one object around, and it doesn't remain an unspecced property bag the way it is now. More reliable for anyone needing to interact with it and easier to extend. This is the sort of thing I was intending to do when I was going to overhaul the errors, but you beat me to it with this PR ;-).

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -36,23 +37,28 @@ public function __construct(Factory $factory = null)
...
             'property' => $this->convertJsonPointerIntoPropertyPath($path ?: new JsonPointer('')),
             'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
-            'message' => $message,
-            'constraint' => $constraint,
+            'message' => ucfirst(vsprintf($constraint->getMessage(), array_map(function ($val) {
+                if (is_scalar($val)) {
+                    return $val;
+                }
+
+                return json_encode($val);
```
erayd: I like the idea of json-encoding complex types. Good idea :+1:.

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -36,23 +37,28 @@ public function __construct(Factory $factory = null)
...
+            }, array_values($more)))),
+            'constraint' => array(
```
erayd: Same comment as above - if you feed it to the constructor, then this becomes unnecessary.

src/JsonSchema/Constraints/BaseConstraint.php
shmax: I took the `ucfirst` from elsewhere in the code. If it was safe before, it's safe now. Remember that this bit of code has nothing to do with localized strings. The messages managed by `ConstraintError` are part of our language-agnostic internal workings, which are currently in English.

erayd: Won't this bit end up handling localized strings at some point though (and isn't that a desirable thing)? If not, then you're double-handling the formatting.

If it won't ever handle localized strings, then I may not entirely understand how you envisage this working - if that's the case, can you explain an outline of how you intend things to work?

src/JsonSchema/Constraints/BaseConstraint.php
shmax: `ConstraintError` is just an enum. I'm not sure it's really meant to be built up into a proper class, with it's own members and whatnot. You don't `new` them, for example. You can read up on it here: https://github.com/marc-mabe/php-enum

shmax: As mentioned in the other comment, `ConstraintError` is not used like a conventional class. We could throw it out and use a regular class like you're suggesting, but we would still have a lot of fragmentation of the properties passed in (although it's a little tighter now that I'm sectioning `more` off into the `params` bucket. I'm not sure that using a class would really gain anything other than sort of formalizing the values it manages, but this is the only place in the code where the values are gathered together, so what would be the gain? A consumer would now have a class instance dropped in his lap to figure out.

src/JsonSchema/Constraints/BaseConstraint.php
shmax: Remember that our goal is NOT to do any localizing ourselves, not now, not ever. All I'm shooting for in this PR is to:

1. centralize the error codes and internal error messages so that someone who DOES want to localize has an easy reference
2. Make the arguments that would be passed to a localizing mechanism easily available ( `params` in the error object), already in the correct order and ready to go.
3. Typehint addError somehow that it is impossible for json-schema devs (not localizers) to misuse.

That's pretty much it. So how WOULD someone do localization? Minimally, like this:

```
foreach(  $validator->getErrors() as $error ) {
    $msgId = "msg_{$error->constraint}";
    // insert your favorite localization technology here
    echo $i18n->loc($msgId, $error['params'];
}
```

I guess I would like to hear some feedback from one of the guys that wanted localization to see what he has in mind; maybe we could find a way to add some sugar.

src/JsonSchema/Constraints/BaseConstraint.php
erayd: I would quite like a class, but I can also see where you're coming from here too - I think we had quite different visions for how the errors should eventually end up working.

May I think on this a bit? I like what you're doing here, but I would like to see if I can think of a way to integrate the two approaches, as I see no reason why they can't work together.

Would you consider accepting a PR on your PR branch?

src/JsonSchema/Constraints/BaseConstraint.php
erayd: Aaah, I thought you were intending a localisation interface to eventually come out the far side of this, to support people who were wanting to localise while still not doing that ourselves. If you're wanting to keep the existing error format, then the way you have done it seems like a suitable approach :-).

shmax: I honestly hadn't gotten as far as making it fully glorious for real-world localizers with real use-cases and real technologies. Let's call this phase I, and invite comment from the guy who wanted help with this...

src/JsonSchema/Constraints/BaseConstraint.php
shmax: I was not trying to solve the error system, here. You are still welcome to do that. I was only trying to get localizers unstuck. Why don't we just evaluate this PR in light of that goal, merge it into 6.0 if it accomplishes what it sets out to do, and then you can revamp the error system in another PR all you like and no argument from me (although I will have plenty of arguments)?

src/JsonSchema/Constraints/BaseConstraint.php
erayd: I think we should go with this as you've written it, and maybe do anything else in a different PR, because actual localization interfaces would seem to be out-of-scope for this one.

src/JsonSchema/Constraints/BaseConstraint.php
erayd: Works for me :-)

src/JsonSchema/ConstraintError.php
shmax: I think @erayd and I already came to an understand on this, but just for those follow along at home `ErrorConstraint` is not a localization interface. It's just a  central repository for the natural language strings that we already had floating around in the code. Its only relevance to localization is as a reference.

src/JsonSchema/ConstraintError.php
shmax: Hmm, tricky. It's a good thought, but then I need to figure out how to add an error for the error. I started working on it but I wound up with a huge blob of messy code, and it seems like a more elegant approach would be for `ConstraintError` to throw its own exception, which I could catch in `addError` (which would do its own call to itself). What do you think?

erayd: What's wrong with something like this? No reason for it get complicated.
```PHP
return isset($messages[$name]) ? $messages[$name] : $messages[self:UNKNOWN_ERROR];
```

erayd: Although if you want to do the exception thing, that's a perfectly workable approach too.

erayd: After thinking about it for a bit, I think that maybe throwing an UnknownErrorException and letting the user deal with it (rather than catching it) is a good idea. That way it'll cause the unit tests to fail if we ever throw a nonexistent error, and assuming we did things properly and the new code has test coverage, we should catch it before the changes ship, because Travis will be screaming at us about it.

shmax:
@harej any comment?

harej:
I think this is going in a good direction. The key thing will be having access to the type of error and parameters (i.e. offending inputs), using this information to construct error messages. Then we can build a package to prepare localized errors.

shmax:
Wonderful, thanks much.

shmax:
I opted not to try to work the error-not-found message into `ConstraintError`--it implies that it's a string to be localized, when really the only time it should ever be encountered is in json-schema development. It just wasn't like the others, but I am now throwing a `InvalidArgumentException`.

As long as we're changing the `addError` API, I also swapped the order of $constraint and $path, because I couldn't find any cases of a $constraint not being passed in, but I couldn't make it non-optional and not come before the ones that are optional.

shmax:
@bighappyface please review and merge if you're satisfied

adjenks:
I want this release badly for the i18n.

shmax:
Good grief, has this never been released?

erayd:
@shmax It was merged three years ago, and currently resides in the master branch, but is not backportable for compatibility reasons.

@adjenks If you use the master branch, you can make use of this feature.

PR #365, opened 2017-03-01
Fix infinite recursion on some schemas when setting defaults (#359)

### LINKED ISSUE #359: infinite loop with Constraint::CHECK_MODE_APPLY_DEFAULTS
I've found an issue, with some schema, the validator will go into an infinite loop and eventually terminate when the allowed memory is exhausted, you can reproduce the issue with this code :

```php
<?php

include_once __DIR__."/../vendor/autoload.php";

$validator = new \JsonSchema\Validator();

$data = [];
$schema = (object)['$ref' => 'http://json-schema.org/draft-04/schema'];

echo __LINE__, PHP_EOL;

echo __LINE__, PHP_EOL;

$validator->validate($data, $schema, \JsonSchema\Constraints\Constraint::CHECK_MODE_APPLY_DEFAULTS);

echo __LINE__, PHP_EOL;
```

I'm not sure yet what is causing this. I'll guess a infinite recursion has been introduced in UndefinedConstraint::validateCommonProperties. I'll try to make another reproduction of the bug with a smaller schema

### LINKED ISSUE #377: Handling of nulls when applying defaults
Currently null default values are not correctly applied. I think this happens because validateCommonProperties, in UndefinedConstraint.php, uses isset to check whether a default value has been defined and it cannot distinguish between a null default value and an undefined one.

```php
// e.g.
if (!$this->getTypeCheck()->propertyExists($value, $i) && isset($propertyDefinition->default)) {
```

Moreover, when the subject is an array and default properties or items have been defined, non-null default values are applied even if the properties or the items have a null value. Again, I think this happens because the code uses isset to check whether the property or the item exists.

```php
if (!isset($value[$i]) && isset($propertyDefinition->default)) {
// ...
if (!isset($value[$i]) && isset($itemDefinition->default)) {
```

erayd:
Made this PR to collaborate with @mathroc on #359. The root cause has been located and is not a bug, but the recursion needs to be mitigated, as infinite recursion of a validator is in violation of the spec.

Will backport the fix to 5.2.0 once this PR is merged.

## Changes Relating to #359
 * Add `CHECK_MODE_ONLY_REQUIRED_DEFAULTS`
 * Don't expand nested defaults (fixes #359)

## Other Changes
Seeing as I'm doing this defaults-related PR anyway, I figured I might as well do a little tidying-up of the defaults code, plus fix any other bugs I run into while tracking this one down.

 * Fix another bug with trying to fetch an empty URI
 * Refactor defaults to use `LooseTypeCheck` where appropriate
 * Move default-setting into its own method
 * Rename one of the defaults variables for clarity
 * Add tests:
    + Ensure non-container values aren't treated like containers;
    + Infinite recursion of default objects via `$ref`
    + Default values do not override defined `null`
    + Default values of `null` are correctly applied
 * Fix second test case for `DefaultPropertiesTest::testLeaveBasicTypesAlone`
 * Fix applying default values of `null` / not overwriting `null` (fixes #377)

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -114,6 +114,7 @@ class UndefinedConstraint extends Constraint
         // Apply default values from schema
         if ($this->factory->getConfig(self::CHECK_MODE_APPLY_DEFAULTS)) {
             if (isset($schema->properties) && LooseTypeCheck::isObject($value)) {
+                // $value is an object or assoc array, and properties are defined - treat as an object
                 foreach ($schema->properties as $currentProperty => $propertyDefinition) {
                     if (!LooseTypeCheck::propertyExists($value, $currentProperty) && isset($propertyDefinition->default)) {
                         if (is_object($propertyDefinition->default)) {
@@ -124,7 +125,7 @@ class UndefinedConstraint extends Constraint
                     }
                 }
             } elseif (isset($schema->items) && LooseTypeCheck::isArray($value)) {
-                // $value is an array, and default items are defined - treat as plain array
+                // $value is an array, and items are defined - treat as plain array
                 foreach ($schema->items as $currentProperty => $itemDefinition) {
                     if (!isset($value[$currentProperty]) && isset($itemDefinition->default)) {
                         if (is_object($itemDefinition->default)) {
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -113,38 +113,24 @@ class UndefinedConstraint extends Constraint
 
         // Apply default values from schema
         if ($this->factory->getConfig(self::CHECK_MODE_APPLY_DEFAULTS)) {
-            if ($this->getTypeCheck()->isObject($value) && isset($schema->properties)) {
-                // $value is an object, so apply default properties if defined
+            if (isset($schema->properties) && LooseTypeCheck::isObject($value)) {
                 foreach ($schema->properties as $currentProperty => $propertyDefinition) {
-                    if (!$this->getTypeCheck()->propertyExists($value, $currentProperty) && isset($propertyDefinition->default)) {
+                    if (!LooseTypeCheck::propertyExists($value, $currentProperty) && isset($propertyDefinition->default)) {
                         if (is_object($propertyDefinition->default)) {
-                            $this->getTypeCheck()->propertySet($value, $currentProperty, clone $propertyDefinition->default);
+                            LooseTypeCheck::propertySet($value, $currentProperty, clone $propertyDefinition->default);
                         } else {
-                            $this->getTypeCheck()->propertySet($value, $currentProperty, $propertyDefinition->default);
+                            LooseTypeCheck::propertySet($value, $currentProperty, $propertyDefinition->default);
                         }
                     }
                 }
-            } elseif ($this->getTypeCheck()->isArray($value)) {
-                if (isset($schema->properties)) {
-                    // $value is an array, but default properties are defined, so treat as assoc
-                    foreach ($schema->properties as $currentProperty => $propertyDefinition) {
-                        if (!isset($value[$currentProperty]) && isset($propertyDefinition->default)) {
-                            if (is_object($propertyDefinition->default)) {
-                                $value[$currentProperty] = clone $propertyDefinition->default;
-                            } else {
-                                $value[$currentProperty] = $propertyDefinition->default;
-                            }
-                        }
-                    }
-                } elseif (isset($schema->items)) {
-                    // $value is an array, and default items are defined - treat as plain array
-                    foreach ($schema->items as $currentProperty => $itemDefinition) {
-                        if (!isset($value[$currentProperty]) && isset($itemDefinition->default)) {
-                            if (is_object($itemDefinition->default)) {
-                                $value[$currentProperty] = clone $itemDefinition->default;
-                            } else {
-                                $value[$currentProperty] = $itemDefinition->default;
-                            }
+            } elseif (isset($schema->items) && LooseTypeCheck::isArray($value)) {
+                // $value is an array, and default items are defined - treat as plain array
+                foreach ($schema->items as $currentProperty => $itemDefinition) {
+                    if (!isset($value[$currentProperty]) && isset($itemDefinition->default)) {
+                        if (is_object($itemDefinition->default)) {
+                            $value[$currentProperty] = clone $itemDefinition->default;
+                        } else {
+                            $value[$currentProperty] = $itemDefinition->default;
                         }
                     }
                 }
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -187,6 +187,39 @@ class UndefinedConstraint extends Constraint
     }
 
     /**
+     * Check whether a default should be applied for this value
+     *
+     * @param mixed $schema
+     * @param mixed $parentSchema
+     * @param bool  $requiredOnly
+     *
+     * @return bool
+     */
+    private function shouldApplyDefaultValue($requiredOnly, $schema, $name = null, $parentSchema = null)
+    {
+        // required-only mode is off
+        if (!$requiredOnly) {
+            return true;
+        }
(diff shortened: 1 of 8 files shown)


shmax:
LGTM

mathroc:
hmm in fact there's still an issue when trying to validate a schema, this fails :

```php
<?php

include_once __DIR__."/../vendor/autoload.php";

$validator = new \JsonSchema\Validator();

$data = (object)["propertyOne" => (object)[]];
$schema = (object)['$ref' => 'http://json-schema.org/draft-04/schema'];

echo __LINE__, PHP_EOL;

var_dump($validator->isValid());
echo __LINE__, PHP_EOL;

$validator->validate(
    $data,
    $schema,
    \JsonSchema\Constraints\Constraint::CHECK_MODE_APPLY_DEFAULTS | \JsonSchema\Constraints\Constraint::CHECK_MODE_REQUIRE_DEFAULTS
);

var_dump($validator->isValid());
echo __LINE__, PHP_EOL;
```


> $ php src/test.php
> 21
> bool(true)
> 25
 >
> Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 32 bytes) in /scripts/vendor/justinrainbow/json-schema/src/JsonSchema/Constraints/Constraint.php on line 205

mathroc:
I think `&& !$this->factory->getConfig(self::CHECK_MODE_REQUIRE_DEFAULTS)` should be added to the ifs for array and leaf

erayd:
Thanks - I'll get onto this in the morning; it's 11pm here :-).

Note that I have not yet implemented a recursion fix in this PR - that is coming.

erayd:
>I think && !$this->factory->getConfig(self::CHECK_MODE_REQUIRE_DEFAULTS) should be added to the ifs for array and leaf

Bit more complex than that - need to check whether they are actually required - but god knows what I was thinking when I committed the required-defaults-only feature it in its current state; it's more than a little bit incomplete!

erayd:
@mathroc Sorted. I also renamed the option to `CHECK_MODE_ONLY_REQUIRED_DEFAULTS`, in the interests of being obvious to the end user.

Still need to address the recursion thing.

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -201,6 +174,76 @@ protected function validateCommonProperties(&$value, $schema = null, JsonPointer
...
+            if (
+                $name !== null
+                && isset($schema->required)
+                && is_array($schema->required)
+                && in_array($name, $schema->required)
+            ) {
+                return true;
+            }
+            // draft-03 required is set
+            if (isset($definition->required) && !is_array($definition->required) && $definition->required) {
```
mathroc: `$definition->required` can be something else than a boolean ?

erayd: Yup - in draft-04, it's an array of strings.

mathroc:
looks neat, I'll test it in the morning; it's 11pm here ;)

> Still need to address the recursion thing.

if I got this right, I don't think there is a recursion thing when `CHECK_MODE_ONLY_REQUIRED_DEFAULTS` is set. for the other case, it does not seems that easy, good luck!

erayd:
@mathroc Sounds like we are in timezones exactly 12 hours apart! You're in the UK?

>if I got this right, I don't think there is a recursion thing when CHECK_MODE_ONLY_REQUIRED_DEFAULTS is set. for the other case, it does not seems that easy, good luck!

There could be if you wrote a schema that required the looping defaults. Plus, the spec says that validators aren't allowed to get infinitely recurse, so I still have to deal with it. Thanks for your good wishes! :-)

mathroc:
> @mathroc Sounds like we are in timezones exactly 12 hours apart! You're in the UK?

France, you're in NZ ?

> There could be if you wrote a schema that required the looping defaults

indeed. I wonder: is there a "finite" json document that would be valid according to such a schema ? (even without applying defaults)
maybe if there is an exit option with a `$anyOf`. so now I'm wondering what would happen if there is default values in `$anyOf`'d schemas, would they all be applied ? wouldn't that potentially create invalid documents as a result ? eg:

```json
{
	"$anyOf": [{
		"type" : "object",
		"properties": {
			"a": {
				"type": "string"
			}
		},
		"required" : ["a"],
		"additionalProperties": false
	},{
		"type" : "object",
		"properties": {
			"b": {
				"type": "string"
			}
		},
		"required" : ["b"],
		"additionalProperties": false
	}]
}
```

erayd:
@mathroc 
>France, you're in NZ ?

Yep, I'm a Kiwi :-).

Re your schema question - as I understand it, the schema you have quoted above would be satisfied by either an object containing only a string `b`, or an object containing only a string `a`.

mathroc:
yes but what would be the json document if called with en empty object and the `CHECK_MODE_APPLY_DEFAULTS` flag ?

it should be either an object containing only a string b, or an object containing only a string a, yes. but which one? I've not tested it, but it might also return an object with both `a` and `b` set and (maybe) mark it as invalid. I guess the question is: once a $anyOf schema is validated, does the lib tries the second schema anyway ? and the answer seems to be here : https://github.com/justinrainbow/json-schema/blob/ef3ee8356a2ea9bb44135787dab2f1491dcbb9f7/src/JsonSchema/Constraints/UndefinedConstraint.php#L249 which means it will return an object with `a` set (and I just realized I forgot the default in the previous exemple - fixed)

so now, there is valid json documents for this schema :

```json
{
	"$anyOf": [{
		"type" : "object",
		"properties": {
			"a": {
				"$ref": "#",
				"default": {}
			}
		},
		"required" : ["a"],
		"additionalProperties": false
	},{
		"type" : "object",
		"additionalProperties": false
	}]
}
```

but the current version of lib would recurse infinitely. what should the fixed version return ? `{"a": {}}` or `{}` ? I would prefer `{}` but I can't think of an easy solution, looks like we would have to first try each schema and if none is valid, retry with applied defaults…

and I think that's a schema would have the issue even with `CHECK_MODE_ONLY_REQUIRED_DEFAULTS` at the moment

ps: sorry if I'm being too talkative

erayd:
I'll have to think on it - let me fix the recursion first!

erayd:
And no, you're not being too talkative - you're making me think, which is a good thing here.

erayd:
@mathroc Think I've got the recursion sorted :-). Your example above will return `{"a":{}}`.

@everyone Can anyone else think of an example that will infinitely recurse? Object recursion should be impossible, but if there's a way to do it with arrays...

erayd:
Never mind, my brain cooperated and threw up an array recursion:
```JSON
{
    "items": [
        {
            "$ref": "#",
            "default": []
        }
    ]
}
```

Time to deal with the array cases now...

erayd:
Array recursion is now also fixed.

@mathroc @shmax Please review.

erayd:
Rebased.

erayd:
Travis fail is Travis' fault; the build for PHP-7.1 never ran. Tests should all be passing.

shmax:
I don't entirely follow all this, but I see no obvious black holes. LGTM

erayd:
@shmax Which part(s) are you finding confusing? I tried to make it obvious what I was doing, but if it's problematic for you to follow, maybe I wasn't clear enough - it might be worth my while to add some more comments in the code.

shmax:
Naw, I meant the meta discussion regarding recursion and the finer points of the spec and such; had my attention elsewhere while it was going on and I didn't follow along as closely as I should have. I trust that you guys have it properly sorted, and the code looks fine. LGTM

erayd:
Gotcha - thanks :-)

mathroc:
I finally took the time to look at it and test it, everything looks fine to me, LGTM!

mathroc:
not that it matters that much, but why is this pr for 6.0.0-dev instead of master ? it does not seems to contains BC break

erayd:
@mathroc Because it needs to be in 6.0.0 anyway, and adding it to master first means a merge mess later. I'm backporting non-bc-breaking changes from 6.0.0 as they are merged (see #368).

erayd:
@shmax @mathroc
Thanks for reviewing :-).

I have just added one final commit after your LGTMs to fix #377 (c92dfae), if you also wish to review that. I would have done it as a separate PR, except that the default code has significantly changed in this one, so it was more logical to put it here.

@bighappyface Anything else that needs doing to this PR before merging?

erayd:
Rebased.

PR #366, opened 2017-03-02
Add more unit tests

erayd:
### New Coverage
 * `Validator::coerce`
 * `SchemaStorage::getUriRetriever`
 * `SchemaStorage::getUriResolver`
 * Most of `Iterator\ObjectIterator`
 * `ConstraintError`
 * `Entity\JsonPointer::__construct` (invalid type for `$value`)
 * `Constraints\TypeConstraint::validateTypeNameWording`
 * `Constraints\TypeConstraint::validateType` (throws exception on invalid type)
 * Various parts of `Uri\UriRetriever`
 * Most of `Uri\Retrievers\FileGetContents`
 * `Uri\Retrievers\Curl`

### New Ignores
 * `strlen` in `StringConstraint::strlen` - because mbstring is available on all test platforms, so this line is never hit.
 * `$http_response_header` in `Uri\Retrievers\FileGetContents::retrieve` - because `$http_response_header` is set in the local method scope and is inaccessible to tests.
 * The `\RuntimeException` in `\Uri\Retrievers\Curl::__construct` - because `curl_init` will always be available to the testsuite, so this will never be thrown.

### Roadblocks
 * `FormatConstraint::validateDateTime` (see [#234](https://github.com/justinrainbow/json-schema/pull/234#issuecomment-284917771))
 * `TypeConstraint::validateType` (email case - see #373)

## Why
Because better test coverage is always a good thing. The goal is to get up to 100% coverage in this PR.

Note that in cases where there is a test for an unfixed bug, Travis will fail. This will auto-resolve when the bugfix is merged and this PR is rebased to include it.


erayd:
Note that the `SchemaStorage` test makes two remote HTTP requests for the official spec. Merging #362 will automatically make these local.

If #362 ends up not getting merged, then that test will need redoing to avoid the requests.

erayd:
Yay! This PR now has 100% test coverage of [6.0.0](https://github.com/justinrainbow/json-schema/tree/6.0.0-dev), with the exception of the two items listed in the **Roadblocks** section above.

Note that there are currently two bugs that cause failed tests in this PR - fixes for those are in #370 and #372.

Is there anything else people would like to see tests added for before this is merged?

erayd:
scr-20170308-195202

shmax:
Bravo, very impressive. 👏

erayd:
Thanks :-).

bighappyface:
Cheers

PR #367, opened 2017-03-02
Save composer.lock in the repository

erayd:
Allows faster dev install and ensures everyone is developing with the same versions of project dependencies. Has no effect on which versions are used when this library is a dependency of something else.


erayd:
Caused issues with travis - maybe better to leave this alone for now.

PR #206, opened 2015-12-15
An email is a string, not much else.

jonathan-s:
Fixes #205 Using email as type now works!

Issue #205, opened 2015-12-15
Type email doesn't seem to work


jonathan-s:
> Uncaught exception 'JsonSchema\Exception\InvalidArgumentException' with message 'aoasundqvist.jonathangmail.com is an invalid type for email' in /var/www/site-honshuu/vendor/justinrainbow/json-schema/src/JsonSchema/Constraints/TypeConstraint.php:139"
[15-Dec-2015 18:42:20] WARNING: [pool sdcom] child 4487 said into stderr: "Stack trace:"
[15-Dec-2015 18:42:20] WARNING: [pool sdcom] child 4487 said into stderr: "#0 /var/www/site-honshuu/vendor/justinrainbow/json-schema/src/JsonSchema/Constraints/TypeConstraint.php(73): JsonSchema\Constraints\TypeConstraint->validateType('aoasundqvist.jo...', 'email')"

schema in question uses `"emailAdress": {"type": "email"}`

PR #373, opened 2017-03-08
Revert "An email is a string, not much else."

erayd:
This reverts commit 73ef463f5ee1d50db9f01069c5a060a3d354888a.

'email' is only a valid type attribute in draft-03 (later versions of
the spec explicitly limit the value of type to the core primitive
types), and this code doesn't validate email addresses anyway. IMO we
should be validating it properly or not at all, and noting this went
away after draft-03 my opinion is on the not-at-all side of the fence.

Note that 'email' is *never* defined as a spec type, in any version -
it just slips in under the radar of the draft-03 language which allows
users to put arbitrary things in the type field.

### DIFF
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -206,10 +206,6 @@ class TypeConstraint extends Constraint
             return is_string($value);
         }
 
-        if ('email' === $type) {
-            return is_string($value);
-        }
-
         if ('null' === $type) {
             return is_null($value);
         }


erayd:
See #205 and #206 for background.

erayd:
If we do want to validate email addresses, that should be done using the `format` attribute - this kind of thing is exactly what `format` is intended for. The spec *does* specify 'email' as a valid option for `format`, and says that if it is a string it should conform to [RFC-5322](http://json-schema.org/latest/json-schema-validation.html#RFC5322).

Email validation of `format` per the above is already handled by `FormatConstraint`, but using the older RFC-822 standard.

PR #375, opened 2017-03-08
add enum wrapper

shmax:
Adding an intermediate `Enum` class. This way consumers of the library can extend `ConstraintError` without having references to unfamiliar namespaces exposed in their code, and if we ever decide to go with a different enum implementation we don't have to worry (as much) about breaking their code.

### DIFF
--- a/src/JsonSchema/Enum.php
+++ b/src/JsonSchema/Enum.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace JsonSchema;
+
+abstract class Enum extends \MabeEnum\Enum
+{
+}
--- a/src/JsonSchema/ConstraintError.php
+++ b/src/JsonSchema/ConstraintError.php
@@ -4,7 +4,7 @@ namespace JsonSchema;
 
 use JsonSchema\Exception\InvalidArgumentException;
 
-class ConstraintError extends \MabeEnum\Enum
+class ConstraintError extends Enum
 {
     const ADDITIONAL_ITEMS = 'additionalItems';
     const ADDITIONAL_PROPERTIES = 'additionalProp';


src/JsonSchema/Enum.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace JsonSchema;
+
+class Enum extends \MabeEnum\Enum
```
erayd: Can this please be abstract?

src/JsonSchema/Enum.php
```diff
@@ -0,0 +1,7 @@
...
+{
```
erayd: This should declare an abstract method `getMessage`:
```PHP
abstract public function getMessage();
```

shmax: Hmm, are you sure? Why would we assume that any other uses of Enum also have something to do with messages?

src/JsonSchema/Enum.php
shmax: Probably. Will give it a try when I get home.

src/JsonSchema/Enum.php
erayd: Good point - I assumed based on your summary, but you do mention extending `ConstraintError` explicitly. Let's leave this PR as you originally had it :-).

src/JsonSchema/Enum.php
erayd: If this isn't intended to be extended-only (and might be used directly for something), then it doesn't need to be abstract, and probably shouldn't be. I was suggesting that based on an invalid assumption that this would always be extended for errors.

shmax: Well that's the thing--I don't think that Enum or even Mabe/Enum has any utility without being extended. In fact, the base class itself is indeed abstract: 

https://github.com/marc-mabe/php-enum/blob/master/src/Enum.php#L16

So I think you're right on the money on this one

src/JsonSchema/Enum.php
shmax: You're right. My code makes sense, my comment doesn't. A user would extend `ConstraintError`, not `Mabe/Enum`, so my fear of the unfamiliar namespace being exposed was not correct.

src/JsonSchema/Enum.php
erayd: Gotcha - if that's the case, then yeah, it should be abstract.

src/JsonSchema/Enum.php
erayd: If that's the case, what is the motivation for this PR? There's nothing that uses it other than `ConstraintError`, and if you aren't intending users to extend it either...

shmax: More for internal benefit, I guess. If someone else decides to use Enum for something they can just extend `/JsonSchema/Enum` instead of `/Mabe/Enum`, and if we ever decide to switch to some other Enum implementation we only have to change the code in one place. Just a little future-proofing.

erayd: Fair enough :-).

shmax:
Rebased. Thanks for reviewing.

PR #378, opened 2017-03-10
[BUGFIX] Add provided schema under a dummy / internal URI (fixes #376)

### LINKED ISSUE #376: Error during schema validation
I have the following schema:
`{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "type": "object",
    "oneOf": [
        {
            "properties": {
                "localeId": {
                    "type": "integer"
                },
                "answer": {
                    "type": "string"
                },
                "inputs": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "$ref": "#/definitions/inputs"
                    }
                }
            },
            "required": [
                "localeId",
                "answer"
            ]
        },
        {
            "properties": {
                "answerId": {
                    "type": "integer"
                },
                "inputs": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "$ref": "#/definitions/inputs"
                    }
                }
            },
            "required": ["answerId"]
        }
    ],
    "definitions": {
        "inputs": {
            "properties": {
                "id": {
                    "type": "integer"
                },
                "value": {
                    "type": "string"
                }
            },
            "required": [
                "id",
                "value"
            ]
        }
    }
}`

If I want to validate data, the _JsonSchema\SchemaStorage::resolveRef_ were called with parameter _"#/definitions/inputs"_. The problem is that JsonPointer returned an empty filename  in _JsonSchema\SchemaStorage::addSchema_ and the _uriRetriever_ try to get contents from a empty file which runs into an error.

erayd:
In order to resolve internal `$ref` references within a user-provided schema, `SchemaStorage` needs to know about the schema. As user-supplied schemas do not have an associated URI, use a dummy / internal one instead.

Fixes issue #376.

### DIFF
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -3,7 +3,6 @@
 namespace JsonSchema;
 
 use JsonSchema\Entity\JsonPointer;
-use JsonSchema\Exception\InvalidArgumentException;
 use JsonSchema\Exception\UnresolvableJsonPointerException;
 use JsonSchema\Iterator\ObjectIterator;
 use JsonSchema\Uri\UriResolver;
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -3,6 +3,7 @@
 namespace JsonSchema;
 
 use JsonSchema\Entity\JsonPointer;
+use JsonSchema\Exception\InvalidArgumentException;
 use JsonSchema\Exception\UnresolvableJsonPointerException;
 use JsonSchema\Iterator\ObjectIterator;
 use JsonSchema\Uri\UriResolver;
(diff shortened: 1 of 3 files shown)


src/JsonSchema/SchemaStorage.php
```diff
@@ -3,6 +3,7 @@
 namespace JsonSchema;
 
 use JsonSchema\Entity\JsonPointer;
+use JsonSchema\Exception\InvalidArgumentException;
```
shmax: Is this referenced anywhere?

src/JsonSchema/Validator.php
```diff
@@ -41,6 +41,9 @@ public function validate(&$value, $schema = null, $checkMode = null)
             $this->factory->setConfig($checkMode);
         }
 
+        // add provided schema to SchemaStorage with internal URI to allow internal $ref resolution
+        $this->factory->getSchemaStorage()->addSchema('internal://provided-schema', $schema);
```
shmax: Should this be a class constant?

src/JsonSchema/SchemaStorage.php
erayd: Nope - well spotted; I've removed it.

src/JsonSchema/Validator.php
erayd: Good idea - have made it so.

shmax:
LGTM

PR #381, opened 2017-03-11
Coerce empty string to null

shmax:
Fix for the null coercion portion of https://github.com/justinrainbow/json-schema/issues/379

Coerce empty sting to null for type "null".

@erayd @bighappyface @gtuccini

### DIFF
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -211,7 +211,7 @@ class TypeConstraint extends Constraint
             if ($coerce) {
                 $value = $this->toNull($value);
             }
-            
+
             return is_null($value);
         }
 
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -211,6 +211,7 @@ class TypeConstraint extends Constraint
             if ($coerce) {
                 $value = $this->toNull($value);
             }
+            
             return is_null($value);
         }
 
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -208,6 +208,9 @@ class TypeConstraint extends Constraint
         }
 
         if ('null' === $type) {
+            if ($coerce) {
+                $value = $this->toNull($value);
+            }
             return is_null($value);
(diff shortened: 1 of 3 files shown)


src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -208,6 +208,9 @@ protected function validateType(&$value, $type)
         }
 
         if ('null' === $type) {
+            if ($coerce) {
+                $value = $this->toNull($value);
+            }
             return is_null($value);
```
erayd: Need a newline above `return is_null($value);` to make the style checker happy.

tests/Constraints/CoerciveTest.php
```diff
@@ -128,7 +128,7 @@ public function getValidCoerceTests()
                   "boolean":"true",
                   "object":{},
                   "array":[],
-                  "null":null,
+                  "null":"",
```
erayd: I notice that you have lots of explicit tests in `testValidCoerceCases`, in addition to simply requiring the schema to validate, but that null coercion hasn't been added as one of them. Is there a reason for that?

shmax: I don't understand. Can you provide an example? 

I believe I originally created this file by duplicating BasicTypesTest.

shmax: Ohhh, I think I get what you mean. One sec...

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -211,6 +211,7 @@ protected function validateType(&$value, $type)
             if ($coerce) {
                 $value = $this->toNull($value);
             }
+
```
erayd: The newline is good, but need to remove the unneeded whitespace on this line.

tests/Constraints/CoerciveTest.php
shmax: Added a post-validate test for null.

I also removed what looked like a duplicated block of explicit tests in the same section of code. Not sure if there was any purpose for it...

erayd:
I think we're good - thanks for doing this PR :-).

erayd:
Is this PR still necessary? Seems to me that #384 has made it redundant.

shmax:
Right you are. Closing...

PR #382, opened 2017-03-11
add quiet option

shmax:
Fix for https://github.com/justinrainbow/json-schema/issues/380

Only emit output from CLI when there are errors or when `--quiet` is not specified.

Actually, I'm wondering if maybe we should just skip the `--quiet` option and simply not produce any output at all when validation passes. That's generally how other linux programs work from the command line.

@erayd @bighappyface @evilebottnawi

### DIFF
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -224,7 +224,7 @@ $schema = $refResolver->resolveRef($urlSchema);
 
 if (isset($arOptions['--dump-schema'])) {
     $options = defined('JSON_PRETTY_PRINT') ? JSON_PRETTY_PRINT : 0;
-    output(json_encode($schema, $options) . "\n");
+    echo json_encode($schema, $options) . "\n";
     exit();
 }
 
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -210,7 +210,7 @@ try {
     $urlSchema = $resolver->resolve($pathSchema, $urlData);
 
     if (isset($arOptions['--dump-schema-url'])) {
-        output($urlSchema . "\n");
+        echo $urlSchema . "\n";
         exit();
     }
 } catch (Exception $e) {
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -17,7 +17,6 @@ function __autoload($className)
 {
     $className = ltrim($className, '\\');
     $fileName  = '';
-    $namespace = '';
     if ($lastNsPos = strrpos($className, '\\')) {
         $namespace = substr($className, 0, $lastNsPos);
         $className = substr($className, $lastNsPos + 1);
@@ -112,6 +111,7 @@ Usage: validate-json data.json
 Options:
       --dump-schema     Output full schema and exit
       --dump-schema-url Output URL of schema
+      --quiet           Suppress output unless there are errors
    -h --help            Show this help
 
 HLP;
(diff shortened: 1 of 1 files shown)


evilebottnawi:
> and simply not produce any output at all when validation passes. That's generally how other linux programs work from the command line.

It's a great idea

erayd:
>Actually, I'm wondering if maybe we should just skip the --quiet option and simply not produce any output at all when validation passes. That's generally how other linux programs work from the command line.

I like this a lot, although if we go down this road, having a `--verbose` option probably makes sense.

shmax:
> I like this a lot, although if we go down this road, having a --verbose option probably makes sense.

Yeah, that's better. Let me flip it around...

bin/validate-json
```diff
@@ -221,7 +221,9 @@ try {
     $validator->check($data, $schema);
 
     if ($validator->isValid()) {
-        echo "OK. The supplied JSON validates against the schema.\n";
+        if(isset($arOptions['--verbose'])) {
+            echo "OK. The supplied JSON validates against the schema.\n";
+        }
     } else {
         echo "JSON does not validate. Violations:\n";
         foreach ($validator->getErrors() as $error) {
```
erayd: Looking great :-). 

While you're at it, can we also have a `--quiet` option that suppresses the error output?

shmax: Sure, but what would be the use case?

erayd: scripts or pipelines where you don't care about the error output (and don't want the clutter); you just want to validate. Lots of linux / unix scripts only care about the exit status being zero (OK) or nonzero (not OK), which this script complies with.

shmax: Oh, okay. Sounds good. Comin' up.

shmax:
Added 'quiet' option, and moved all of the option parsing stuff to the top of the file.

bin/validate-json
```diff
@@ -198,21 +210,21 @@ try {
...
-    echo $e->getMessage() . "\n";
+    output("Error loading JSON schema file\n");
+    output($urlSchema . "\n");
+    output($e->getMessage() . "\n");
     exit(2);
 }
 $refResolver = new JsonSchema\SchemaStorage($retriever, $resolver);
 $schema = $refResolver->resolveRef($urlSchema);
 
 if (isset($arOptions['--dump-schema'])) {
```
erayd: Explicit requests for output like this should override `--quiet`.

bin/validate-json
```diff
@@ -198,21 +210,21 @@ try {
     $urlSchema = $resolver->resolve($pathSchema, $urlData);
 
     if (isset($arOptions['--dump-schema-url'])) {
```
erayd: Explicit requests for output like this should override `--quiet`

bin/validate-json
erayd: GitHub highlighted the wrong line - I'm referring to `if (isset($arOptions['--dump-schema']))`

shmax: sgtm

bin/validate-json
erayd: Not sure whether you're intending this one in a different commit to the `--dump-schema` one, or if you missed it - but in the past you've combined this kind of thing in a single commit, so adding a comment just in case.

erayd: Never mind, I see you just posted a commit for that at about the same time as my last comment!

shmax:
My pleasure. Only 28 issues to go!

erayd:
@shmax Any chance you could change the base branch to 6.0.0? Or would you prefer to forward-port this one, rather than backporting to 5.2.0?

shmax:
> Any chance you could change the base branch to 6.0.0?

 SGTM

erayd:
Awesome - thanks :-)

PR #383, opened 2017-03-12
Add option to disable validation of "format" constraint

erayd:
## What
Add `CHECK_MODE_DISABLE_FORMAT`. If set, validation of the `"format"` keyword is ignored completely, and assumed successful.

## Why
From draft-04 onwards, the spec says we should [provide an option to disable format validation](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.2):

> Implementations MAY support the "format" keyword.  Should they choose to do so:
     - they SHOULD implement validation for attributes defined below;
     - they SHOULD offer an option to disable validation for this keyword.

### DIFF
--- a/src/JsonSchema/Constraints/Constraint.php
+++ b/src/JsonSchema/Constraints/Constraint.php
@@ -30,6 +30,7 @@ abstract class Constraint extends BaseConstraint implements ConstraintInterface
     const CHECK_MODE_COERCE_TYPES =     0x00000004;
     const CHECK_MODE_APPLY_DEFAULTS =   0x00000008;
     const CHECK_MODE_EXCEPTIONS =       0x00000010;
+    const CHECK_MODE_DISABLE_FORMAT =   0x00000020;
 
     /**
      * Bubble down the path
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ third argument to `Validator::validate()`, or can be provided as the third argum
 | `Constraint::CHECK_MODE_COERCE_TYPES` | Convert data types to match the schema where possible |
 | `Constraint::CHECK_MODE_APPLY_DEFAULTS` | Apply default values from the schema if not set |
 | `Constraint::CHECK_MODE_EXCEPTIONS` | Throw an exception immediately if validation fails |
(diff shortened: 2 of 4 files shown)


erayd:
@bighappyface / @shmax Please let me know if you want me to slow down - I'm aware I have submitted a large number of PRs recently, and that as a result there is now a backlog of stuff awaiting merging or review. If this is too much, too fast, please say so and I'll hold back for a bit.

shmax:
I don't know about @bighappyface, but I don't mind the pace. We're targeting 6.0 for most of this stuff, anyway, so there's no harm if it takes him a while to get to it.

src/JsonSchema/Constraints/Constraint.php
```diff
@@ -30,6 +30,7 @@
     const CHECK_MODE_COERCE_TYPES =     0x00000004;
     const CHECK_MODE_APPLY_DEFAULTS =   0x00000008;
     const CHECK_MODE_EXCEPTIONS =       0x00000010;
+    const CHECK_MODE_DISABLE_FORMAT =   0x00000020;
```
shmax: Maybe this is just a personal style thing, and it's no big deal, but for some reason I always get a little uncomfortable when I see a variable or constant named in the negative. It makes it the odd duck out, and then you wind up doing odd double-negative logic to check them in the affirmative, eg `if (!disabledThing)`.

Is there any way to flip the polarity and enable it by default, then require the user to explicitly disable it if he doesn't want it?

erayd: Normally, I'd agree with you. I would vastly prefer an on-by-default option.

The reason I did it this way here is because if you pass `$checkMode` directly, rather than going via `Factory::addConfig`, you'll clobber whatever the default is. In this case, that would mean a very high chance of users turning off `"format"` validation by accident (i.e. anybody who uses a custom `$checkMode` and didn't realise the implications of overwriting the default).

Thoughts?

shmax: Good point. I guess this is the best option. 👍

shmax:
LGTM

PR #384, opened 2017-03-13
Improvements to type coercion

### LINKED ISSUE #379: Two possible improvements to type coercion
It would be useful to have a coercible string representation for null. "null" or '' would work just fine.

I think also that the coercion should happen only when the subject doesn't match **any** of the allowed types.
Currently, given the following subject
```json
{
	"value": "true"
}
```
and the following schema
```json
{
	"properties": {
		"value": {
			"type": ["boolean", "string"]
		}
	},
	"type": "object"
}
```
"value" is coerced to boolean, even if its type -- string -- is among the allowed ones.

What do you think?

erayd:
## What
 * Don't coerce already-valid types
 * Add `CHECK_MODE_EARLY_COERCE` to enable previous coercion behavior
 * Don't test further types if we've already validated one successfully
 * Fix reversed arguments from #364 
 * Implement every remaining coercion case from [ajv's matrix](https://github.com/epoberezkin/ajv/blob/master/COERCION.md)
 * Rewrite coercion tests to improve readability and simplify adding future tests.

## Why
See discussion in #379 regarding coercion of already-valid types and `CHECK_MODE_EARLY_COERCE`. Other items are performance / bugfix, or are to improve testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,8 +197,8 @@ Please note that using `CHECK_MODE_COERCE_TYPES` or `CHECK_MODE_APPLY_DEFAULTS` 
 original data.
 
 `CHECK_MODE_EARLY_COERCE` has no effect unless used in combination with `CHECK_MODE_COERCE_TYPES`. If
-enabled, type coercion will occur as soon as possible, even if the value may already be of another valid
-type.
+enabled, the validator will use (and coerce) the first compatible type it encounters, even if the
+schema defines another type that matches directly and does not require coercion.
 
 ## Running the tests
 
--- a/README.md
+++ b/README.md
@@ -186,14 +186,19 @@ third argument to `Validator::validate()`, or can be provided as the third argum
 | `Constraint::CHECK_MODE_NORMAL` | Validate in 'normal' mode - this is the default |
 | `Constraint::CHECK_MODE_TYPE_CAST` | Enable fuzzy type checking for associative arrays and objects |
 | `Constraint::CHECK_MODE_COERCE_TYPES` | Convert data types to match the schema where possible |
+| `Constraint::CHECK_MODE_EARLY_COERCE` | Apply type coercion as soon as possible |
 | `Constraint::CHECK_MODE_APPLY_DEFAULTS` | Apply default values from the schema if not set |
 | `Constraint::CHECK_MODE_ONLY_REQUIRED_DEFAULTS` | When applying defaults, only set values that are required |
 | `Constraint::CHECK_MODE_EXCEPTIONS` | Throw an exception immediately if validation fails |
 | `Constraint::CHECK_MODE_DISABLE_FORMAT` | Do not validate "format" constraints |
 | `Constraint::CHECK_MODE_VALIDATE_SCHEMA` | Validate the schema as well as the provided document |
 
-Please note that using `Constraint::CHECK_MODE_COERCE_TYPES` or `Constraint::CHECK_MODE_APPLY_DEFAULTS`
-will modify your original data.
+Please note that using `CHECK_MODE_COERCE_TYPES` or `CHECK_MODE_APPLY_DEFAULTS` will modify your
+original data.
+
+`CHECK_MODE_EARLY_COERCE` has no effect unless used in combination with `CHECK_MODE_COERCE_TYPES`. If
+enabled, type coercion will occur as soon as possible, even if the value may already be of another valid
+type.
 
 ## Running the tests
 
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -69,8 +69,8 @@ class TypeConstraint extends Constraint
                 $wording[] = self::$wording[$type];
             }
             $this->addError(ConstraintError::TYPE(), $path, array(
-                    'expected' => gettype($value),
-                    'found' => $this->implodeWith($wording, ', ', 'or')
+                    'found' => gettype($value),
+                    'expected' => $this->implodeWith($wording, ', ', 'or')
             ));
         }
     }
@@ -185,6 +185,10 @@ class TypeConstraint extends Constraint
         }
 
         if ('array' === $type) {
+            if ($coerce) {
+                $value = $this->toArray($value);
+            }
+
             return $this->getTypeCheck()->isArray($value);
         }
 
@@ -213,10 +217,18 @@ class TypeConstraint extends Constraint
         }
 
         if ('string' === $type) {
+            if ($coerce) {
+                $value = $this->toString($value);
+            }
+
             return is_string($value);
         }
 
         if ('null' === $type) {
+            if ($coerce) {
+                $value = $this->toNull($value);
+            }
+
             return is_null($value);
         }
 
@@ -232,19 +244,21 @@ class TypeConstraint extends Constraint
      */
     protected function toBoolean($value)
     {
-        if ($value === 'true') {
+        if ($value === 1 || $value === 'true') {
             return true;
         }
-
-        if ($value === 'false') {
+        if (is_null($value) || $value === 0 || $value === 'false') {
             return false;
+        }
+        if ($this->getTypeCheck()->isArray($value) && count($value) === 1) {
+            return $this->toBoolean(reset($value));
         }
 
         return $value;
     }
 
     /**
-     * Converts a numeric string to a number. For example, "4" becomes 4.
+     * Converts a value to a number. For example, "4.5" becomes 4.5.
      *
      * @param mixed $value the value to convert to a number
      *
@@ -255,14 +269,89 @@ class TypeConstraint extends Constraint
         if (is_numeric($value)) {
             return $value + 0; // cast to number
         }
+        if (is_bool($value) || is_null($value)) {
+            return (int) $value;
+        }
+        if ($this->getTypeCheck()->isArray($value) && count($value) === 1) {
+            return $this->toNumber(reset($value));
+        }
 
         return $value;
     }
 
+    /**
+     * Converts a value to an integer. For example, "4" becomes 4.
+     *
+     * @param mixed $value
+     *
+     * @return int|mixed
+     */
     protected function toInteger($value)
     {
-        if (is_numeric($value) && (int) $value == $value) {
-            return (int) $value; // cast to number
+        $numberValue = $this->toNumber($value);
+        if (is_numeric($numberValue) && (int) $numberValue == $numberValue) {
+            return (int) $numberValue; // cast to number
+        }
+
+        return $value;
+    }
+
+    /**
+     * Converts a value to an array containing that value. For example, [4] becomes 4.
+     *
+     * @param mixed $value
+     *
+     * @return array|mixed
+     */
+    protected function toArray($value)
+    {
+        if (is_scalar($value) || is_null($value)) {
+            return array($value);
+        }
+
+        return $value;
+    }
+
+    /**
+     * Convert a value to a string representation of that value. For example, null becomes "".
+     *
+     * @param mixed $value
+     *
+     * @return string|mixed
+     */
+    protected function toString($value)
+    {
+        if (is_numeric($value)) {
+            return "$value";
+        }
+        if ($value === true) {
+            return 'true';
+        }
+        if ($value === false) {
+            return 'false';
+        }
+        if (is_null($value)) {
+            return '';
+        }
+        if ($this->getTypeCheck()->isArray($value) && count($value) === 1) {
+            return $this->toString(reset($value));
+        }
+    }
+
+    /**
+     * Convert a value to a null. For example, 0 becomes null.
+     *
+     * @param mixed $value
+     *
+     * @return null|mixed
+     */
+    protected function toNull($value)
+    {
+        if ($value === 0 || $value === false || $value === '') {
+            return null;
+        }
+        if ($this->getTypeCheck()->isArray($value) && count($value) === 1) {
+            return $this->toNull(reset($value));
         }
 
         return $value;
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -45,11 +45,12 @@ class TypeConstraint extends Constraint
         $type = isset($schema->type) ? $schema->type : null;
         $isValid = false;
         $coerce = $this->factory->getConfig(self::CHECK_MODE_COERCE_TYPES);
+        $earlyCoerce = $this->factory->getConfig(self::CHECK_MODE_EARLY_COERCE);
         $wording = array();
 
         if (is_array($type)) {
-            $this->validateTypesArray($value, $type, $wording, $isValid, $path, false);
-            if (!$isValid && $coerce) {
+            $this->validateTypesArray($value, $type, $wording, $isValid, $path, $coerce && $earlyCoerce);
+            if (!$isValid && $coerce && !$earlyCoerce) {
                 $this->validateTypesArray($value, $type, $wording, $isValid, $path, true);
             }
         } elseif (is_object($type)) {
@@ -57,8 +58,8 @@ class TypeConstraint extends Constraint
 
             return;
         } else {
-            $isValid = $this->validateType($value, $type, false);
-            if (!$isValid && $coerce) {
+            $isValid = $this->validateType($value, $type, $coerce && $earlyCoerce);
+            if (!$isValid && $coerce && !$earlyCoerce) {
                 $isValid = $this->validateType($value, $type, true);
             }
         }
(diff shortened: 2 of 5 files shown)


shmax:
LGTM

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -185,6 +185,10 @@ protected function validateType(&$value, $type, $coerce = false)
         }
 
         if ('array' === $type) {
+            if ($coerce) {
+                $value = $this->toArray($value);
```
shmax: Looks good, but I notice that ajv only does this conversion if `coerceTypes` is set to `array`. Should this be controlled by another flag?

erayd: Hmm, good point. I'm happy either way - would you like a flag?

shmax: Honestly don't know. We can try without for a while, and see if the rationale for having one ever emerges.

erayd: Good suggestion - let's do that then. And if someone asks for it, then we can add it at that point.

I don't supposed you know whether @epoberezkin has stated his rationale for doing it that way in ajv?

shmax: Well, I think this is the issue that spawned the feature: https://github.com/epoberezkin/ajv/issues/158

They discuss it a bit, then come up with the `coerceTypes:'array'` idea

erayd: Hmm... will have to think on this a bit. At a minimum I should write some tests to check that post-coercion validation is happening properly.

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -255,10 +269,27 @@ protected function toNumber($value)
 
     protected function toInteger($value)
     {
-        if (is_numeric($value) && (int) $value == $value) {
+        $numberValue = $this->toNumber($value);
```
shmax: Little risky, isn't it?

```PHP
echo toInteger("5f"); // 5
```

erayd: Ooh, good catch - this is why we have code reviews! Thanks :-)

erayd: Should be sorted now.

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -263,15 +263,23 @@ protected function toNumber($value)
         if (is_numeric($value)) {
             return $value + 0; // cast to number
         }
+        if (is_bool($value) || is_null($value)) {
+            return (int) $value;
+        }
+        if ($this->getTypeCheck()->isArray($value) && count($value) === 1) {
+            reset($value);
```
shmax: `reset` returns the first element in the array, so you can skip a step and  pass its results directly on line 272

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -249,14 +267,46 @@ protected function toNumber($value)
...
+     *
+     * @param mixed $value
+     *
+     * @return mixed
+     */
     protected function toInteger($value)
     {
-        if (is_numeric($value) && (int) $value == $value) {
-            return (int) $value; // cast to number
+        $numberValue = $this->toNumber($value);
```
shmax: what does the call to `toNumber` get you?

src/JsonSchema/Constraints/TypeConstraint.php
erayd: Good point - thanks :-).

src/JsonSchema/Constraints/TypeConstraint.php
erayd: All the other type coersion cases. It means no copy pasta, because other than integer-ness, the two methods are identical.

src/JsonSchema/Constraints/TypeConstraint.php
```diff
@@ -249,14 +267,46 @@ protected function toNumber($value)
...
+
+        return $value;
+    }
+
+    /**
+     * Converts a value to an array containing that value
+     *
+     * @param mixed $value
+     *
+     * @return array
```
shmax: I believe the return type should be `array|mixed`

src/JsonSchema/Constraints/TypeConstraint.php
shmax: is there anything actually functionally different? Because now we're calling `is_numeric` twice whereas before it was only called once

src/JsonSchema/Constraints/TypeConstraint.php
erayd: Is the version of PHPDoc in this project OK with that? I've had issues with types like that in the past.

erayd: (If it's not OK, it should be mixed)

erayd: Never mind, aparently this project doesn't use PHPDoc. I'll change to `array|mixed`.

shmax: just `mixed` is fine, too. I suggested `array|mixed` to emphasize that `array` is primarily what we're interested in

erayd: I like your original suggestion better. If we're not using PHPdoc, then there's no reason to avoid it IMO. If such a reason crops up later, it's a pretty quick / easy change to downgrade the return type to `mixed`.

src/JsonSchema/Constraints/TypeConstraint.php
erayd: Yes - the second `is_numeric` call catches cases that fall through `toNumber` without being successfully coerced.

erayd:
@shmax I think this PR now contains everything we intended it to contain. Is there anything you'd like me to add / change / remove / hide-in-the-corner-and-cry-from-shame-about?

src/JsonSchema/Constraints/TypeConstraint.php
shmax: that's what I mean, though--if you put the code back the way it was then there is only one call to `is_numeric`. So I guess I'm still not clear on what is functionally different about the new code. Are additional values now being coerced to integer (eg. boolean) with this code?

erayd: Yes. The new code also coerces from bool, null, and array, via the call to `toNumber`.

erayd: Basically the logic is to apply all the coercion rules for number, and then see if the result can be further coerced to an int. If it is, then return the cast int - if not, return the original value.

shmax: okie-doke

shmax:
Wow, very impressive, as always! I'll take another look tomorrow when I'm fresh but for now I say LGTM

erayd:
Thanks :-).

erayd:
Hmm, just thought of one other thing - I need to ensure that the double-coercion for array cases is covered.

erayd:
Double-coercion tests done.

erayd:
@shmax How's this?

README.md
```diff
@@ -188,10 +188,15 @@ third argument to `Validator::validate()`, or can be provided as the third argum
...
 | `Constraint::CHECK_MODE_EXCEPTIONS` | Throw an exception immediately if validation fails |
+| `CHECK_MODE_EARLY_COERCE` | Apply type coercion as soon as possible |
 
 Please note that using `Constraint::CHECK_MODE_COERCE_TYPES` or `Constraint::CHECK_MODE_APPLY_DEFAULTS`
 will modify your original data.
 
+`CHECK_MODE_EARLY_COERCE` has no effect unless used in combination with `CHECK_MODE_COERCE_TYPES`. If
+enabled, type coercion will occur as soon as possible, even if the value may already be of another valid
+type.
+
```
shmax: Could you maybe expand on this a bit to make it more clear? I know what you're talking about, but I'm not sure that someone else coming to the project for the first time would.

tests/Constraints/CoerciveTest.php
```diff
@@ -11,274 +11,181 @@
...
-        $checkMode = Constraint::CHECK_MODE_TYPE_CAST | Constraint::CHECK_MODE_COERCE_TYPES;
-
-        $schemaStorage = new SchemaStorage($this->getUriRetrieverMock(json_decode($schema)));
-        $schema = $schemaStorage->getSchema('http://www.my-domain.com/schema.json');
+        // check type conversions
+        $types = array(
+            // toType
+            'string' => array(
+                //    fromType      fromValue       toValue         valid   Test Number
+                array('string',     '"string"',     'string',       true),  // #0
```
shmax: Wow, I love the new grid, much easier to read!

shmax: I usually find it confusing when a code sample uses values that mimic or mirror the type they are meant to be demonstrating. Instead of 'string', can we use something else?

tests/Constraints/CoerciveTest.php
```diff
@@ -11,274 +11,181 @@
...
-            $this->assertEquals($errors, $validator->getErrors(), print_r($validator->getErrors(), true));
+        $this->assertEquals($startType, $type, "Incorrect type '$type': expected '$startType'");
+
+        $validator->validate($data, $schema);
+
+        // check validity
+        if ($valid) {
+            $this->assertTrue(
+                $validator->isValid(),
+                'Validation failed: ' . json_encode($validator->getErrors(), \JSON_PRETTY_PRINT)
```
shmax: 👍

README.md
erayd: Certainly.

tests/Constraints/CoerciveTest.php
```diff
@@ -11,271 +11,201 @@
...
-        $this->assertFalse($validator->isValid(), print_r($validator->getErrors(), true));
+        return $tests;
     }
 
-    /**
-     * @dataProvider getInvalidCoerceTests
-     */
-    public function testInvalidCoerceCasesUsingAssoc($input, $schema, $errors = array())
+    /** @dataProvider dataCoerceCases **/
+    public function testCoerceCases($schema, $data, $startType, $endType, $endValue, $valid, $early = false, $assoc = false)
```
shmax: What if the 7th argument was just something like `$flags`? Then instead of passing in `true` you could pass in `CHECK_MODE_EARLY_COERCE`, which would be a little easier to read, and you don't have to keep adding more arguments to this method signature as more flags are added.

tests/Constraints/CoerciveTest.php
```diff
@@ -11,271 +11,201 @@
...
+                array('array',      '["-45"]',      -45,            true),  // #13
+                array('object',     '{"a":"b"}',    null,           false), // #14
+            ),
+            'boolean' => array(
+                array('string',     '"true"',       true,           true),  // #15
+                array('integer',    '1',            true,           true),  // #16
+                array('boolean',    'true',         true,           true),  // #17
+                array('NULL',       'null',         false,          true),  // #18
+                array('array',      '["true"]',     true,           true),  // #19
+                array('object',     '{"a":"b"}',     null,          false), // #20
```
shmax: extra space?

tests/Constraints/CoerciveTest.php
```diff
@@ -11,271 +11,201 @@
...
+        // check type conversions
+        $types = array(
+            // toType
+            'string' => array(
+                //    fromType      fromValue       toValue         valid   Test Number
+                array('string',     '"string"',     'string',       true),  // #0
+                array('integer',    '45',           '45',           true),  // #1
+                array('boolean',    'true',         'true',         true),  // #2
+                array('boolean',    'false',        'false',        true),  // #3
+                array('NULL',       'null',         '',             true),  // #4
```
shmax: What's with the counting numbers in the comments? They look very nice, but it seems like the kind of thing that would be a pain to maintain going forward. What are they for?

erayd: They are the test number, starting from zero. PHPUnit reports those numbers when anything fails - it makes tracking down a failing test case much faster / easier when you don't have to manually count them all every time, particularly when there are this many tests.

tests/Constraints/CoerciveTest.php
erayd: Yup - well spotted!

tests/Constraints/CoerciveTest.php
```diff
@@ -11,271 +11,201 @@
...
+                );
+            }
+        }
 
-        $this->assertTrue(gettype($value->number) == 'double');
-        $this->assertTrue(gettype($value->integer) == 'integer');
-        $this->assertTrue(gettype($value->negativeInteger) == 'integer');
-        $this->assertTrue(gettype($value->boolean) == 'boolean');
+        // #38 check post-coercion validation (to array)
+        $tests[] = array(
```
shmax: I think it's okay to break out these more complicated ones like this, but since even the complex ones have only a single property, can they be setup before you iterate over them on 83?

tests/Constraints/CoerciveTest.php
erayd: Good idea - will do.

tests/Constraints/CoerciveTest.php
erayd: They can. The reason I did it this way is because complex tests are more likely to be added than simple ones, and adding things to the beginning or midway through requires manually renumbering all the test cases - whereas adding something to the end is easy, and doesn't require renumbering anything.

shmax: I wasn't suggesting adding them to the big grid, only adding them after the grid but before the loop. That way you don't have to do the brain work of trying to figure out what the json string will look like.

tests/Constraints/CoerciveTest.php
erayd: Sure :-).

tests/Constraints/CoerciveTest.php
shmax: Right, that's what I was originally thinking about when I suggested breaking things up into smaller tests. `testEarlyCoerce`, `testArray`, `testNoCoerceNeeded`, etc. Maybe not for the vanilla stuff, but at least the complex ones. This is okay, too, though.

tests/Constraints/CoerciveTest.php
erayd: No problem - done.

tests/Constraints/CoerciveTest.php
erayd: Would you prefer them as separate methods? I like them better all in one place like this, because it's easier to compare them - but I can certainly break to one method per case if you'd rather.

tests/Constraints/CoerciveTest.php
shmax: Oh wait, I get it--you can handle the complex tests the same way you do the others because they have arrays of types, so you can't slot them in the same way as the others.

shmax: Or at least a few of them do. It seems like some of them could still be handled the same way as the ones in the big grid, eg:

before
```PHP
// #39 check post-coercion validation (from array)
        $tests[] = array(
            '{"properties":{"propertyOne":{"type":"number"}}}',
            '{"propertyOne":["ABC"]}',
            'array', null, null, false
        );
```

after
```PHP
// #39 check post-coercion validation (from array)
        $tests['number'][] = array(
            'array',
            '{"propertyOne":["ABC"]}',
            null, null, false
        );
```

I'm sure I have all that in the wrong order, but hopefully you know what I mean

tests/Constraints/CoerciveTest.php
shmax: Up to you. I can live with the numbers.

tests/Constraints/CoerciveTest.php
erayd: Good point - have moved #39 to the grid. I don't think any of the others can move though - if you spot any, please let me know, but I think they all have reasons for being where they are.

tests/Constraints/CoerciveTest.php
erayd: Then I'd like to keep them the way they are please - I find it much easier to debug this way :-).

shmax:
LGTM Fantastic work! Take a victory lap

erayd:
Thanks :-). Really appreciate the work you've done reviewing all this too, and the various suggestions you've made - the result is better because of that.

erayd:
Rebased.

bighappyface:
@erayd conflict?

erayd:
@bighappyface I think we overlapped, and I rebased before you merged the conflict. I have rebased it again.

erayd:
Rebased.

bighappyface:
Once more, please :)

erayd:
Done.

PR #401, opened 2017-03-22
[BUGFIX] Fix autoload to work properly with composer dependencies

erayd:
## What
Replace the autoloader in `validate-json` with the composer one.

## Why
Because the current autoloader cannot handle dependencies, which means that `validate-json` cannot find `MabeEnum\Enum`.

### DIFF
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -6,32 +6,18 @@
  * @author Christian Weiske <christian.weiske@netresearch.de>
  */
 
-/**
- * Dead simple autoloader
- *
- * @param string $className Name of class to load
- *
- * @return void
- */
-function __autoload($className)
-{
-    $className = ltrim($className, '\\');
-    $fileName  = '';
-    if ($lastNsPos = strrpos($className, '\\')) {
-        $namespace = substr($className, 0, $lastNsPos);
-        $className = substr($className, $lastNsPos + 1);
-        $fileName  = str_replace('\\', DIRECTORY_SEPARATOR, $namespace) . DIRECTORY_SEPARATOR;
-    }
-    $fileName .= str_replace('_', DIRECTORY_SEPARATOR, $className) . '.php';
-    if (stream_resolve_include_path($fileName)) {
-        require_once $fileName;
-    }
+// support running this tool from git checkout
+if (is_dir(__DIR__ . '/../vendor')) {
+    set_include_path(__DIR__ . '/..' . PATH_SEPARATOR . get_include_path());
 }
 
-// support running this tool from git checkout
-if (is_dir(__DIR__ . '/../src/JsonSchema')) {
-    set_include_path(__DIR__ . '/../src' . PATH_SEPARATOR . get_include_path());
+// autoload composer classes
+$composerAutoload = stream_resolve_include_path('vendor/autoload.php');
+if (!$composerAutoload) {
+    echo("Cannot load json-schema library\n");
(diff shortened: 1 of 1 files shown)


bin/validate-json
```diff
@@ -6,32 +6,18 @@
...
-        $namespace = substr($className, 0, $lastNsPos);
-        $className = substr($className, $lastNsPos + 1);
-        $fileName  = str_replace('\\', DIRECTORY_SEPARATOR, $namespace) . DIRECTORY_SEPARATOR;
-    }
-    $fileName .= str_replace('_', DIRECTORY_SEPARATOR, $className) . '.php';
-    if (stream_resolve_include_path($fileName)) {
-        require_once $fileName;
-    }
+// support running this tool from git checkout
+if (is_dir(__DIR__ . '/../vendor')) {
```
bighappyface: Would `dirname` work better here to help avoid the path separator headache?

http://php.net/manual/en/function.dirname.php

erayd: Can you clarify? `PATH_SEPARATOR` has nothing to do with filesystem paths - it's the ':' character used as a list separator when defining multiple search paths (ie. what you'll find in the `$PATH` environment variable). Not sure how `dirname` has anything to do with that...

shmax:
Wow, nice! LGTM

bin/validate-json
bighappyface: I edited the comment to refer to directory separator; I didn't realize my phrasing collided with that other constant at the time I wrote it.

Screwing with relative paths in strings to wire things up always seems dirty to me, especially with directory separators differing between platforms.

An example of what I am suggesting:

```shell
$ cd /path/to/json-schema/bin
$ php -a
Interactive shell

php > echo __DIR__;
/path/to/json-schema/bin
php > echo dirname(__DIR__);
/path/to/json-schema
php > echo dirname(__DIR__) . DIRECTORY_SEPARATOR . 'vendor';
/path/to/json-schema/vendor
```

What do you think?

erayd: Works for me, I'll do that now.

bin/validate-json
```diff
@@ -6,32 +6,19 @@
...
-// support running this tool from git checkout
-if (is_dir(__DIR__ . '/../src/JsonSchema')) {
-    set_include_path(__DIR__ . '/../src' . PATH_SEPARATOR . get_include_path());
+// autoload composer classes
+$composerAutoload = stream_resolve_include_path('vendor/autoload.php');
+if (!$composerAutoload) {
+    echo("Cannot load json-schema library\n");
+    exit(1);
 }
+require($composerAutoload);
```
mathroc: just curiosity, why not just `require __DIR__.'/../vendor/autoload.php';` ?

erayd: So that it can be resolved from anywhere in the include path. If you're asking "why use `dirname`", that's because @bighappyface doesn't like the relative path & hardcoded Unix directory separator.

mathroc: ok, it seems a bit odd to me to alter the include_path to access a file you know the absolute path; but well, in a binary it does not matter much

erayd: It seems odd to me too. The impression I got from reading the original code is that the author anticipated the utility script being split from the rest of the package and installed elsewhere, and the include path business is because of that. I don't know the reasons behind that approach being present in the code, so I kept it when I made the change.

PR #403, opened 2017-03-22
Revert 5.2.0 backports prior to 6.0.0 merge

erayd:
## What
Revert commit e3c9bccdc38bbd09bcac0131c00f3be58368b416 to allow 6.0.0-dev to merge smoothly.

This seems like a *much* cleaner approach than messing with `6.0.0-dev` and either having tons of duplicate commits, or losing commit history there. This way, all that needs to happen prior to `6.0.0-dev` being merged, is merging this one to master.

## Why
Because @bighappyface [asked me to solve](https://github.com/justinrainbow/json-schema/pull/390#issuecomment-288563099) the 6.0.0 merge conflict.


erayd:
@bighappyface I'm aware that this is a PR against master, rather than against `6.0.0-dev`, which is what you actually asked me for. But it seems to me that doing it this way makes far more sense, and means that `git blame` doesn't get damaged by distorting the commit history on `6.0.0-dev`.

If you don't like it, let me know, and I'll go back to the other approach.

mathroc:
maybe the v6.0.0 branch should be the master branch ? and backports to v5.x only happen in 5.x.x ?

erayd:
@mathroc It's about to be - that's the whole point of this PR, and is why 5.x.x was created. This PR allows all of 6.0.0-dev to merge into master smoothly, without needing to orphan the 5.2.0 release which had already gone out.

Issue #21, opened 2012-11-28
wrong handling of "required": false


paranoiq:
consider following schema fragmet:

```
{
    "type": "object",
    "properties": {
        "foo": {
            "type": "object",
            "required": false,
            "properties": {
                "bar": {
                    "type": "string",
                    "required": true
                }
            }
        },
        ...
}
```

if "foo" is not present in validated data, validation will fail with message `foo.bar is missing and it is required`

but when "foo" does not exist (and it does not need to!) it is no reason to validate its also non-existing child "bar". "bar" should be validated only if "foo" exists

PR #405, opened 2017-03-23
fix bug when applying defaults

mathroc:
> Warning: First parameter must either be an object or the name of an existing class in vendor/justinrainbow/json-schema/src/JsonSchema/Constraints/UndefinedConstraint.php line 261

to reproduce:

```php
<?php

use JsonSchema\Constraints\Constraint;

include_once __DIR__."/../vendor/autoload.php";

$validator = new \JsonSchema\Validator();

$data = [];
$schema = (object)[
    "type" => "array",
    "items" => (object)[
        "required" => [],
        "type" => "object",
    ]
];

$validator->validate($data, $schema, Constraint::CHECK_MODE_APPLY_DEFAULTS);
```

according to the spec:

> The value of "items" MUST be either a schema or array of schemas.

but here the code supposes it's an array of schemas and it fails when it's trying to use `[]` (the value of `required`) as a schema

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -253,8 +253,14 @@ class UndefinedConstraint extends Constraint
                 }
             }
         } elseif (isset($schema->items) && LooseTypeCheck::isArray($value)) {
+            $items = array();
+            if (LooseTypeCheck::isArray($schema->items)) {
+                $items = $schema->items;
+            } elseif (isset($schema->minItems) && count($value) < $schema->minItems) {
+                $items = array_fill(count($value), $schema->minItems - count($value), $schema->items);
+            }
             // $value is an array, and items are defined - treat as plain array
-            foreach ($schema->items as $currentItem => $itemDefinition) {
+            foreach ($items as $currentItem => $itemDefinition) {
                 if (
                     !array_key_exists($currentItem, $value)
                     && property_exists($itemDefinition, 'default')
--- a/tests/Constraints/DefaultPropertiesTest.php
+++ b/tests/Constraints/DefaultPropertiesTest.php
@@ -149,6 +149,21 @@ class DefaultPropertiesTest extends VeryBaseTestCase
                 '{"items":[{"default":null}]}',
                 '[null]'
             ),
+            array(// #21 items might be a schema (instead of an array of schema)
+                '[{}]',
+                '{"items":{"properties":{"propertyOne":{"default":"valueOne"}}}}',
+                '[{"propertyOne":"valueOne"}]'
+            ),
+            array(// #22 if items is not an array, it does not create a new item
+                '[]',
+                '{"items":{"properties":{"propertyOne":{"default":"valueOne"}}}}',
+                '[]'
+            ),
+            array(// #23 if items is a schema with a default value and minItems is present, fill the array
+                '["a"]',
+                '{"items":{"default":"b"}, "minItems": 3}',
+                '["a","b","b"]'
+            ),
         );
     }
 


erayd:
Nice catch! Appreciate you finding / fixing this :-).

It's 2:15am here, but your bug report has made me think - if `minItems` is set, and `items` is a schema, then we should probably set the default as many times as needed to fulfil that. Thoughts?

The comment on test #21 needs fixing, it says the same thing twice.

I'll review this properly in the morning when I'm not half asleep ;-).

mathroc:
> It's 2:15am here, but your bug report has made me think - if minItems is set, and items is a schema, then we should probably set the default as many times as needed to fulfil that. Thoughts?

yes we could :)

> The comment on test #21 needs fixing, it says the same thing twice.

fixed

> I'll review this properly in the morning when I'm not half asleep ;-).

👍

tests/Constraints/DefaultPropertiesTest.php
```diff
@@ -149,6 +149,16 @@ public function getValidTests()
                 '{"items":[{"default":null}]}',
                 '[null]'
             ),
+            array(// #21 items might be an array of schema (instead of an array of schema)
```
erayd: This comment needs clarifying - it says the same thing twice.

tests/Constraints/DefaultPropertiesTest.php
```diff
@@ -149,6 +149,16 @@ public function getValidTests()
...
+                '[{}]',
+                '{"items":{"properties":{"propertyOne":{"default":"valueOne"}},"required":[]}}',
```
erayd: What is the purpose of `required` in this test?

tests/Constraints/DefaultPropertiesTest.php
```diff
@@ -149,6 +149,16 @@ public function getValidTests()
...
+                '[{"propertyOne":"valueOne"}]'
+            ),
+            array(// #22 if items is not an array, it does not create new item
+                '[]',
+                '{"items":{"properties":{"propertyOne":{"default":"valueOne"}},"required":[]}}',
```
erayd: What is the purpose of `required` in this test?

erayd:
Thanks for fixing this bug :-).

> yes we could :)

Are you intending to do the work on `minItems` (as part of this PR, or in another PR), or would you like me to do it? I'm quite happy either way :-).

tests/Constraints/DefaultPropertiesTest.php
mathroc: they were not, it's just how I found out about this bug and I didn't realized I could remove it from the tests

mathroc:
I've added support for minItems, tell me if you think of something missing (I'll squash everything when it's ready to merge)

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -253,8 +253,14 @@ protected function applyDefaultValues(&$value, $schema, $path)
                 }
             }
         } elseif (isset($schema->items) && LooseTypeCheck::isArray($value)) {
+            $items = array();
+            if (LooseTypeCheck::isArray($schema->items)) {
+                $items = $schema->items;
+            } elseif (isset($schema->minItems) && count($value) < $schema->minItems) {
+                $items = array_fill(count($value), $schema->minItems - 1, $schema->items);
```
erayd: Why the magic number? Shouldn't this be less `count($value)`, rather than less `1`?

src/JsonSchema/Constraints/UndefinedConstraint.php
```diff
@@ -253,8 +253,14 @@ protected function applyDefaultValues(&$value, $schema, $path)
                 }
             }
         } elseif (isset($schema->items) && LooseTypeCheck::isArray($value)) {
+            $items = array();
+            if (LooseTypeCheck::isArray($schema->items)) {
+                $items = $schema->items;
+            } elseif (isset($schema->minItems) && count($value) < $schema->minItems) {
```
erayd: Needs to check and handle the state of `$schema->exclusiveMinimum`.

mathroc: I has no idea this existed, I'll add a test for `exclusiveMinimum`

src/JsonSchema/Constraints/UndefinedConstraint.php
mathroc: absolutely! I thought the second argument of array_fill was `$end_index` but it's `$num`…

mathroc:
should be ok if I understood `exclusiveMinimum` correctly

erayd:
Grrr.... I've sent you on a wild goose chase sorry, `exclusiveMinimum` applies to `minimum` only, not to `minItems` (and therefore should not be in this PR). I should not have tried to review this when I was half asleep; this is my mistake.

Eveything else looks fine; once the `ExclusiveMinimum` stuff goes away (sorry!) this should be merge-ready.

mathroc:
don’t worry, I read its definition to quickly too, it’s gone

mathroc:
rebased

PR #406, opened 2017-03-24
Testcase for minProperties with properties defined + Fix tests

flolivaud:
The `minProperties` is not working if `properties` is defined.

It's probably because of https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/UndefinedConstraint.php#L66


erayd:
Thanks for catching this. What do you think is wrong with that line?

I notice that this PR contains tests, but not a fix. Are you intending to fix this bug, or are you asking us to fix it, and contributing tests?

mathroc:
@flolivaud is a co worker so I can give some more info. 

In https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/UndefinedConstraint.php#L66 I don't think `$this->factory->getSchemaStorage()->resolveRefSchema($schema->properties)` is right: it’s returning a "properties" object instead of a schema. We have not yet tested it, but dropping the `isset($schema->properties)` and always passing `$schema` might be enough

It’s not a blocker for us so I don’t know when we’ll take a look at this, for the moment we're contributing tests. no rush to fix ☺

mathroc:
might be related to /fixed by #406

erayd:
@mathroc I think you're right. I'll check if it's still an issue tomorrow - if it is, then I'll add a fix for this to #411.

erayd:
@mathroc #411 seems to fix it - these tests now pass :-).

@flolivaud Could you please rebase this on 6.0.0-dev?

tests/Constraints/MinMaxPropertiesTest.php
```diff
@@ -72,7 +72,7 @@ public function getInvalidTests()
         return array(
             array(
                 '{
-                  "value": 1
```
erayd: Why is this value being changed, rather than adding the new test independently?

tests/Constraints/MinMaxPropertiesTest.php
```diff
@@ -82,8 +82,26 @@ public function getInvalidTests()
...
+                    "propertyTwo": {
+                      "type": "string"
+                    }
+                  },
+                  "minProperties": 1
+                }'
+            ),
+            array(
                 '{
-                  "value": 1
```
erayd: Why is this value being changed, rather than adding the new test independently?

tests/Constraints/MinMaxPropertiesTest.php
mathroc: because this test was incorrect, the object was found invalid for the wrong reason (because `value` was an integer instead of an object)

tests/Constraints/MinMaxPropertiesTest.php
mathroc: because this test was incorrect, the object was found invalid for the wrong reason (because `value` was an integer instead of an object)

mathroc: both tests might be simplified by not having the `value` wrapping but we chose not to change that to keep the diff simpler

flolivaud:
see #416 for 6.0.0-dev base

flolivaud:
close duplicate #416

mathroc:
@erayd how do you change the base of a PR ?

erayd:
@mathroc At the top-right of the PR, click the 'Edit' button. The select the base branch that you wish to merge into from the dropdown menu just below the PR title.

Issue #407, opened 2017-04-05
\JsonSchema\Constraints\BaseConstraint::arrayToObjectRecursive() doesn't convert empty arrays to objects


shtrom:
When an empty array is given to `\JsonSchema\Constraints\BaseConstraint::arrayToObjectRecursive()`, it is returned identical. This leads to issues later on when using a `SchemaStorage` to contain this schema. Such an empty schema is only really used for tests, but this currently requires workarounds.

The following example code
```php
use \JsonSchema\Validator;
print_r(Validator::arrayToObjectRecursive([ "baz" => [ "bar" => "baz" ]  ])); 
print_r(Validator::arrayToObjectRecursive([]));
```
results in
```
stdClass Object
(
    [baz] => stdClass Object
        (
            [bar] => baz
        )

)
Array
(
)
```

shtrom:
Hum, this seems to only be on the 5.x.x branch. 

Adding cast to `(object)` before (well, after the keyword) the `return` in https://github.com/justinrainbow/json-schema/blob/5.x.x/src/JsonSchema/Constraints/BaseConstraint.php#L146 seems to fix the issue.

shtrom:
Strike that, it's on 6.0.0-dev, too.

erayd:
Thanks - fixed in #409. This will be backported to 5.x.x after merge.

wotnak:
The issue was fixed for passing an empty array directly, but still persists for nested empty arrays.
```php
// Directly passing an empty array.
$array = [];
$object = \JsonSchema\Validator::arrayToObjectRecursive($array);
print_r($object);
// stdClass Object
// (
// )
```

```php
// Passing a nested empty array.
$array = ['shouldBeAnObject' => []];
$object = \JsonSchema\Validator::arrayToObjectRecursive($array);
print_r($object);
// stdClass Object
// (
//     [shouldBeAnObject] => Array
//         (
//         )
// )
```

It could be fixed by passing `JSON_FORCE_OBJECT` flag to `json_encode` used in `arrayToObjectRecursive` method.
https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/BaseConstraint.php#L145
https://www.php.net/manual/en/function.json-encode.php
https://www.php.net/manual/en/json.constants.php#constant.json-force-object
```diff
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
[...]

Issue #408, opened 2017-04-05
SchemaStorage::addSchema() should call BaseConsstraint::arrayToObjectRecursive() on the provide schemas


shtrom:
Without doing this, it might accept Array input from the caller without complaining, but the validation against those schemas will fail later with 

> get_object_vars() expects parameter 1 to be object, array given

at https://github.com/justinrainbow/json-schema/blob/6.0.0-dev/src/JsonSchema/SchemaStorage.php#L118

Calling `BaseConsstraint::arrayToObjectRecursive()` in `addSchema` before https://github.com/justinrainbow/json-schema/blob/6.0.0-dev/src/JsonSchema/SchemaStorage.php#L61, as `Validator::validate()` does https://github.com/justinrainbow/json-schema/blob/6.0.0-dev/src/JsonSchema/Validator.php#L49-L51 should solve the problem, and make handling it in `Validator::validate()` unecessary.

erayd:
Thanks for reporting this.

Your suggested fix causes failed unit tests - I will need to investigate this in more detail to work out what's going on.

erayd:
Fixed in #409. This will be backported to 5.x.x after merge.

DannyvdSluijs:
This is already part of the v6.0.0. release

PR #409, opened 2017-04-05
[BUGFIX] Cast empty schema arrays to object

erayd:
## What
 * Cast empty schema arrays to object.
 * Use `function_exists` instead of checking the PHP version.
 * Move array->object casts into `SchemaConstraint` & `SchemaStorage`.

## Why
 * Fixes bugs #407 & #408
 * `function_exists` allows for the use of polyfills, whereas the version check does not.

### DIFF
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -146,7 +146,7 @@ class BaseConstraint
         $json = json_encode($array);
         if (json_last_error() !== \JSON_ERROR_NONE) {
             $message = 'Unable to encode schema array as JSON';
-            if (version_compare(phpversion(), '5.5.0', '>=')) {
+            if (function_exists('json_last_error_msg')) {
                 $message .= ': ' . json_last_error_msg();
             }
             throw new InvalidArgumentException($message);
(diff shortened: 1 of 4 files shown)


src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -146,12 +146,12 @@ public static function arrayToObjectRecursive($array)
...
             $message = 'Unable to encode schema array as JSON';
-            if (version_compare(phpversion(), '5.5.0', '>=')) {
+            if (function_exists('json_last_error_msg')) {
                 $message .= ': ' . json_last_error_msg();
             }
             throw new InvalidArgumentException($message);
         }
 
-        return json_decode($json);
+        return (object) json_decode($json);
```
shmax: This will only do a shallow cast. Is that intended?

erayd: Yes, it's deliberate. The JSON functions are already handling the deep cast - my understanding of this bug is that it's a shallow-cast problem only (that's the way it was reported).

shmax:
LGTM

src/JsonSchema/Constraints/BaseConstraint.php
shtrom: Yes, it is. `json_decode` already does the deep cast.

PR #411, opened 2017-04-11
[BUGFIX] Split $objectDefinition into $schema and $properties

erayd:
## What
 * Split `$objectDefinition` into `$schema` and `$properties` (fixes bugs in #353);
 * Rename some stuff so that the names make sense.

## Why
Object validation attempts to use a single variable to store both the object schema, and the definition of the object's properties. This causes validation to be incomplete where `"properties"` is not set, but `"additionalProperties"` is.


bighappyface:
@o5 @shmax @mathroc may I have your help giving this another review?

shmax:
Whoops, sorry guys--I'm still paying attention and still interested, but this one just slipped past my radar. Holler if you need anything.

Issue #414, opened 2017-04-12
RFC-3339 Allow The Option of 'T' or space for Date Time.


HunterEl:
Hi There!

I have been using your library for about a year now and it has been fantastic. Thank you. I have been working with a team that would like to use a space instead of a `T` for the Date Time format. If you look at the [RFC-3339](https://tools.ietf.org/html/rfc3339) section `5.6` and the `NOTE` at the bottom of the page:

```  
NOTE: ISO 8601 defines date and time separated by "T".
Applications using this syntax may choose, for the sake of
readability, to specify a full-date and full-time separated by
(say) a space character.
```

The file you have for parsing and validation date time for JSON Schema requires a `T` and does not allow for the option a space. I will submit a PR under this issue number and you can let me know if that works for the direction of your project. It is a minor change to the file `src/JsonSchema/Rfc3339.php` where the regular expression needs to be modified and the date format needs to be altered. 

cheers!

erayd:
Fine with me - thanks for the PR :-). I'll take a look at that now.

bighappyface:
Closed with #415

PR #415, opened 2017-04-12
Issue-414: Allow The Option of T or space for Date time.

HunterEl:
Please see Github Issue: [#414](https://github.com/justinrainbow/json-schema/issues/414)

This PR is to allow for a `T` or space when providing date times to JSON Schema. 

Let me know if there is anything different you would like me to do or if this is not the direction you would like this project to go in. 

Thank you!

### DIFF
--- a/src/JsonSchema/Rfc3339.php
+++ b/src/JsonSchema/Rfc3339.php
@@ -4,7 +4,7 @@ namespace JsonSchema;
 
 class Rfc3339
 {
-    const REGEX = '/^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})(\.\d+)?(Z|([+-]\d{2}):?(\d{2}))$/';
+    const REGEX = '/^(\d{4}-\d{2}-\d{2}[T ]{1}\d{2}:\d{2}:\d{2})(\.\d+)?(Z|([+-]\d{2}):?(\d{2}))$/';
 
     /**
      * Try creating a DateTime instance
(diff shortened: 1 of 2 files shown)


src/JsonSchema/Rfc3339.php
```diff
@@ -23,7 +23,9 @@ public static function createFromString($string)
         $microseconds = $matches[2] ?: '.000000';
         $timeZone = 'Z' !== $matches[3] ? $matches[4] . ':' . $matches[5] : '+00:00';
 
-        $dateTime = \DateTime::createFromFormat('Y-m-d\TH:i:s.uP', $dateAndTime . $microseconds . $timeZone, new \DateTimeZone('UTC'));
+        $dateFormat = strpos($dateAndTime, 'T') === false ? 'Y-m-d H:i:s.uP' : 'Y-m-d\TH:i:s.uP';
```
erayd: Could this check `strpos($dateAndTime, 'T') === 10)` please? That way it won't be tripped up by someone passing a 'T' elsewhere in the string.

erayd: (Although that said, I don't think that this is a huge issue, because timezone codes aren't currently allowed, so things like 'PST' will not be an issue at present. So if you'd rather leave it that way it is, that's fine too.)

HunterEl: absolutely.

erayd:
Also, could you please change the base for this PR to 6.0.0-dev. I'll backport it to 5.3.0 once it's merged.

HunterEl:
@erayd done.

erayd:
@bighappyface Your turn now :-).

erayd:
Backported for 5.2.1 in f375f0b.

PR #419, opened 2017-05-03
[BUGFIX] Split "uri" format into "uri" & "uri-reference", fix meta-schema bug

### LINKED ISSUE #418: "format":"uri" still OK after recent changes?
Hi guys..

First, thanks for the active work on this project, it's very nice to see that much activity going on here ,-)

An update from `5.1.0` to `5.2.0` broke one of our projects unit tests and i'm not sure if our test expectations are wrong - or the changes here ;-)

### The schema

Given this simple schema:

```json
{
  "title": "Example",
  "type": "object",
  "properties": {
    "url": {
      "title": "Url",
      "type": [
        "string",
        "null"
      ],
      "format": "uri"
    }
  },
  "additionalProperties": false
}

```
### The problem

Our unit tests expects the string `jjj--no-url` **not** to be a valid url.. but apparently now, it is ;-)
This was changed recently in commit 5e92df010b5022ee2e4c00cf2f48bba7de05c686

For us, this is kinda of a problem, we don't regard (and who would?) `jjj--no-url` as a valid url..

So if i pass the string `jjj--no-url` as an url, it will land in this condition of the `if`:
https://github.com/justinrainbow/json-schema/blob/5.2.0/src/JsonSchema/Constraints/FormatConstraint.php#L96

### The question

Isn't the check to lean? If we have a look at the line:

```php
$validURL = filter_var('scheme://host/' . $element, FILTER_VALIDATE_URL, FILTER_NULL_ON_FAILURE);
```
So this basically means that *any* string will pass (as long it doesn't have invalid filename parts) as it gets prefixed with a scheme and a host (thus seen as a relative part)..

I checked this, even the string `a` will pass.

IMHO, this makes the `uri` check obsolete and wrong. If we prepend scheme and host *for* the user to pass the url check, the user really *didn't* provide a *correct* url.. 

So is this what is really expected, that *any* string passes the `uri` format?

Thanks ;-)

erayd:
## What
 * Return `"uri"` format to strict validation of only fully-qualified URI.
 * Add `"uri-reference"` format to validate URI references.

## Why
 * [RFC-3986 §1.2.3](https://tools.ietf.org/html/rfc3986#section-1.2.3) says that URI references are not a subset of URI.
 * The draft-03 & draft-04 meta-schemas use the `"uri"` format erroneously - this is a [meta-schema bug](https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/177#issuecomment-293051367).
 * Fixes #418.

### DIFF
--- a/src/JsonSchema/ConstraintError.php
+++ b/src/JsonSchema/ConstraintError.php
@@ -28,6 +28,7 @@ class ConstraintError extends Enum
     const FORMAT_STYLE = 'styleFormat';
     const FORMAT_TIME = 'timeFormat';
     const FORMAT_URL = 'urlFormat';
+    const FORMAT_URL_REF = 'urlRefFormat';
     const INVALID_SCHEMA = 'invalidSchema';
     const LENGTH_MAX = 'maxLength';
     const LENGTH_MIN = 'minLength';
@@ -77,6 +78,7 @@ class ConstraintError extends Enum
             self::FORMAT_STYLE => 'Invalid style',
             self::FORMAT_TIME => 'Invalid time %s, expected format hh:mm:ss',
             self::FORMAT_URL => 'Invalid URL format',
+            self::FORMAT_URL_REF => 'Invalid URL reference format',
             self::LENGTH_MAX => 'Must be at most %d characters long',
             self::INVALID_SCHEMA => 'Schema is not valid',
             self::LENGTH_MIN => 'Must be at least %d characters long',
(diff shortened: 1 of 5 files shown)


erayd:
Travis failures are caused by #420 (dependency resolution), and are unrelated to this PR.

bighappyface:
@erayd please rebase with the fix per #420

erayd:
Done.

bighappyface:
@erayd rebase once more, please.

erayd:
@bighappyface Done.

PR #432, opened 2017-06-06
Added property name in draft-3 required error

sunspikes:
In case of required error with draft-3, the error message is missing property names. (Similar to #91)

Creating a new PR against the updated master.

### DIFF
--- a/src/JsonSchema/ConstraintError.php
+++ b/src/JsonSchema/ConstraintError.php
@@ -43,7 +43,6 @@ class ConstraintError extends Enum
     const NOT = 'not';
     const ONE_OF = 'oneOf';
     const REQUIRED = 'required';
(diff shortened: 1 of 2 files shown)


sunspikes:
@erayd Actually as i mentioned, after creating the PR #430 I've deleted my branch (As you can see in the PR it shows unknown repository). Not sure if i can check out the PR rebase it and push it again without having my original branch.

erayd:
@sunspikes I missed that comment - think I typed my reply before yours loaded for me. Noting that, let's just go with this PR.

For future reference, if you have an open PR, don't delete the branch it's attached to until after it's been merged ;-). If you accidentally delete your branch, GitHub allows you to restore it from the PR (there's a button at the bottom of the PR for this).

Issue #435, opened 2017-06-20
Can't find the correct way to use SchemaStorage


johannes-see:
Due to a lack of understanding on my part and also the rather slim documentation, I am forced to ask about the proper way to use SchemaStorage here.

I want to have a parent schema in place, located at `https://myurl.com/schema/parentschema.schema.json` and then multiple schemata, each inheriting the constraints from `parentschema` at the top level. I.e., the parentschema looks something like
```
{
  "param1": { "type": "integer" },
  "param2": { "type": "string" }
}
```

and the subschemas all look something like
```
{
  "$ref": "https://myurl.com/schema/parentschema.schema.json",
  "required": ["param1"]
}
```
adding only their own respective required structure to the schema.

I am trying to use SchemaStorage like

```
$schema = file_get_contents(__DIR__ . "/../schema/subschema.schema.json");

$schemaStorage = new SchemaStorage();
$schemaStorage->addSchema(__DIR__ . "/../schema/parentschema.schema.json", json_decode($schema));

$fac = new Factory($schemaStorage);
$validator = new Validator($fac);
```

johannes-see:
Nevermind, I found it, and I was entirely too quick to open up an issue.
The documentation could be a bit more generous though.

Some further explanation of SchemaStorage for others running into the same sort of trouble:
* its the way to explain to the validator how to resolve references to outside the schema
* you initialize a SchemaStorage and then add reference addresses to it
* you add reference addresses with by `$schemaStorage->add("/address/written/out", $objectTheAddressHasToPointTo)`
* you initialize the validator with a SchemaStorage
* when the validator comes across a reference to "address/written/out", it already has all the needed structure loaded

erayd:
If you are confused by something, you are most welcome to open issues to ask - I'm aware the documentation is terrible at the moment. Better documentation is on my list of things to do for this project, and is fairly important, because what we have at the moment is fairly useless.

Regarding your original question - what you are trying to do IS NOT SUPPORTED by the JSON Schema standard - there is no official mechanism through which inheritance is possible. If you use `$ref` somewhere, all other properties of the object where `$ref` appears [MUST be ignored](https://tools.ietf.org/html/draft-wright-json-schema-01#section-8):

> An object schema with a "$ref" property MUST be interpreted as a "$ref" reference.  The value of the "$ref" property MUST be a URI Reference.  Resolved against the current URI base, it identifies the URI of a schema to use.  All other properties in a "$ref" object MUST be ignored.

[...]

PR #437, opened 2017-06-21
[BUGFIX] Ignore $ref siblings & abort on infinite-loop references

### LINKED ISSUE #436: Sibling properties of $ref are not ignored
According to the spec:

> An object schema with a "$ref" property MUST be interpreted as a "$ref" reference.  The value of the "$ref" property MUST be a URI Reference.  Resolved against the current URI base, it identifies the URI of a schema to use.  All other properties in a "$ref" object MUST be ignored.

The current behavior is to process all properties regardless of whether `$ref` is present, which is a bug and contrary to the spec.

This needs to be fixed for 6.0.0, and should not be backported - it's a bugfix, but will break known behavior that some people may be depending on.

Big thanks to @johannes-see for noticing this in #435 :+1:.

erayd:
## What
 * Refactor to ignore all other sibling properties where `$ref` is present (fixes #436).
 * Add check for infinitely-looping references

## Why
 * Because the spec states that all sibling properties of `$ref` [MUST be ignored](https://tools.ietf.org/html/draft-wright-json-schema-01#section-8)
 * Because infinitely looping is both undesirable and explicitly prohibited by the spec.

This PR will **not** be backported. #436 is definitely a bug, but fixing it results in a significant change in behavior which some users may be relying on (e.g. as a hack to implement inheritance), and as such it should be merged as major-version change.

Thanks to @johannes-see for finding the bug (#435).

### DIFF
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -94,7 +94,7 @@ class SchemaStorage implements SchemaStorageInterface
     /**
      * {@inheritdoc}
      */
-    public function resolveRef($ref)
+    public function resolveRef($ref, $resolveStack = array())
     {
         $jsonPointer = new JsonPointer($ref);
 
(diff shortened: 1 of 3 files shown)


erayd:
@shmax @bighappyface @mathroc
Do you think a separate PR for 5.x.x that catches infinitely-looping references is a good idea? It may have a significant impact on performance, as doing this in 5.x.x will require a large number of equality checks between objects that may be identical but are not the same instance.

The check in this PR doesn't have the same performance implications, because it only needs to check whether objects are the same instance - but this approach will not work in 5.x.x, because 5.x.x deliberately preserves properties from both the referenced schema and the referee, and returns an entirely new object.

mathroc:
I don't think it's very important to backport the infinitely-looping check to 5.x, working on 6.x and releasing seems a more appropriate use of time. especially as an upgrade to 6.x should be painless, users hitting this bug should be able to upgrade easily once released (or use dev-master before that)

shmax:
I agree with @mathroc. If people haven't complained of this bug before now then they're probably not hitting it, and there's no harm in waiting for 6.

bighappyface:
Agreed on focusing on v6.x.x

erayd:
OK - I won't worry about fixing the loop in 5.x.x then :-).

This PR is ready for review whenever you guys feel like it.

shmax:
👍

PR #442, opened 2017-07-27
Added custom error codes

jamesadler:


erayd:
Not sure what you're trying to achieve with this PR, but if you are intending to reopen this or submit another one, please ensure that the target branch is 6.0.0-dev. New changes and bugfixes all go into that branch - anything that doesn't break the API will be backported to 5.x.x after it's merged.

PR #448, opened 2017-09-30
Add proper recursive handling for $ref resolution base

erayd:
# What
Implement recursive resolution of `$ref` base.

# Why
Because the existing code doesn't actually do this properly, and is various combinations of missing and not implemented in this area.

Note that this patch does *not* distinguish between schema and non-schema contexts; it just resolves `$ref` wherever it may occur within a schema. Limiting this to schema-context only may be added later.

`ObjectIterator` is *not* removed, despite now being redundant, as it's a handy class for people who wish to implement their own preprocessor for whatever reason.

### DIFF
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -5,7 +5,6 @@ namespace JsonSchema;
 use JsonSchema\Constraints\BaseConstraint;
 use JsonSchema\Entity\JsonPointer;
 use JsonSchema\Exception\UnresolvableJsonPointerException;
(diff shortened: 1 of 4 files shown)


erayd:
@shmax Any chance of a code review?

erayd:
Have rebased this on #449 - the latest version of php-cs-fixer introduces some dangerous changes, and also causes CI failures on travis. #449 should be merged prior to this PR.

shmax:
LGTM 💯

erayd:
Rebased.

PR #449, opened 2017-09-30
Update php-csfixer rules to address problem in 2.7 & new multiline rule

erayd:
## What
 * Configure to ignore processing of `yoda_style` rules
 * Explicitly set multiline comment mode
 * Update comment syntax for multiline `//` prefix
 * Explicitly run PHP-5.3 Travis tests on `precise`

## Why
 * `yoda_style` in 2.7 is dangerous and may result in logic errors. In some cases, it also results in invalid syntax.

 * multiline comments prefixed with `//` now seem to be misaligned, and this cannot be disabled, so have changed the relevant comment.

 * `trusty` is now the default PHP testing distro on Travis, and it does not support PHP-5.3.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -11,6 +11,7 @@ $config
         '@PSR2' => true,
         '@Symfony' => true,
         // additionally
+        'align_multiline_comment' => array('comment_type' => 'phpdocs_like'),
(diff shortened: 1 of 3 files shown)


shmax:
LGTM

PR #452, opened 2017-10-04
[BUGFIX] Don't resolve a schema id against itself

erayd:
## What
Add a missing condition to prevent resolving a schema id against itself. This prevents potential double-resolution of relative paths.

## Why
Because this is a bug, and double-resolution of relative paths results in an incorrect dereferencing result.


dkarlovi:
Probably a good idea to add a test for this.

shmax:
LGTM

erayd:
> Probably a good idea to add a test for this.
I agree. Done :-).

PR #455, opened 2017-10-05
Don't add a `file://` prefix to URI that already have a scheme

### LINKED ISSUE #451: Experiencing "Error: JSON schema not found at file://"
When updating our dependencies from 5.2.1 to 5.2.2 today, we experienced this error:

```
Error: JSON schema not found at file://
```

Ref https://github.com/wp-cli/wp-cli/pull/4393 and https://travis-ci.org/wp-cli/wp-cli/jobs/282809129

Let me know what situational details I can provide to help reproduce it. `justinrainbow/json-schema` is a dependency of Composer, which we use within our application.

erayd:
## What
Add regex filter to URI check when expanding `$ref`.

## Why
Some URI types do not pass validation against `\FILTER_VAR_URL` (e.g. `phar://`), but are still valid and still have a scheme. This patch catches those situations via a simple regex.

Fixes #451.

### DIFF
--- a/src/JsonSchema/Uri/UriResolver.php
+++ b/src/JsonSchema/Uri/UriResolver.php
@@ -77,7 +77,11 @@ class UriResolver implements UriResolverInterface
     public function resolve($uri, $baseUri = null)
     {
         // treat non-uri base as local file path
-        if (!is_null($baseUri) && !filter_var($baseUri, \FILTER_VALIDATE_URL)) {
+        if (
+            !is_null($baseUri) &&
+            !filter_var($baseUri, \FILTER_VALIDATE_URL) &&
+            !preg_match('|^[^/]+://|u', $baseUri)
+        ) {
             if (is_file($baseUri)) {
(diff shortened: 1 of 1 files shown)


src/JsonSchema/Uri/UriResolver.php
```diff
@@ -77,7 +77,11 @@ public function generate(array $components)
     public function resolve($uri, $baseUri = null)
     {
         // treat non-uri base as local file path
-        if (!is_null($baseUri) && !filter_var($baseUri, \FILTER_VALIDATE_URL)) {
+        if (
+            !is_null($baseUri) &&
+            !filter_var($baseUri, \FILTER_VALIDATE_URL) &&
+            !preg_match('|^[^/]+://|u', $baseUri)
```
bighappyface: Pipe `|` delimiter so that you do not need to escape the forward slash `/` characters, correct?

erayd:
Yup.

PR #456, opened 2017-10-08
Fix realpath() based logic for windows compatiblity.

david-garcia-garcia:
This dependency is totally breaking Magento 2.2.0 release on Windows. Tried to keep original logic as untouched as posible.


erayd:
Could you please check whether #455 already fixes this issue? As I understand it, this PR is intended to address exactly the same problem.

david-garcia-garcia:
#455 Fixes this. Closing PR.

erayd:
@david-garcia-garcia Sweet, thanks for checking / confirming :-).

PR #464, opened 2017-11-08
marc-mabe/php-enum versions

marc-mabe:
I'm very happy you are planning to use my small library :)

As you are using base functionality only you can support much more versions of it. This helps a lot in cases you have to deal with more libraries and all have different dependent versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "marc-mabe/php-enum": "2.3.1",
+        "marc-mabe/php-enum":"^2.0 || ^3.0",
         "icecave/parity": "1.0.0"
     },
     "require-dev": {


wimleers:
Is the intent to make `json-schema` work with both the `2.x` and `3.x` branches of https://github.com/marc-mabe/php-enum?

wimleers:
(Asking this because it's related to #569.)

shmax:
@wimleers  Looks like we sort of forgot to follow up on this one. We only use the most basic, vanilla features of php-enum. I had a quick look at the "BC break" tags listed in the 3.0 release, and I didn't see anything that would impact us. On the other hand, with 3.0 he raises the minimal required version PHP version to 5.6, and we're still on 5.3. So, we can't fully switch to 3.0 without creating PHP dependency issues, and if we stay on 2.0 then we don't always get the latest and greatest. Any suggestions?

marc-mabe:
PR is rebased now

> Is the intent to make json-schema work with both the 2.x and 3.x branches of https://github.com/marc-mabe/php-enum?

yes

marc-mabe:
any update on this ?

shmax:
@erayd I'm not really familiar with the version branch strategy. Which one do we target with this?

marc-mabe:
@shmax `"^2.0 || ^3.0"` is the same as `(>= 2.0.0 && < 3.0.0) || (>= 3.0.0 && < 3.0.0)` in this case the same as simplified `>= 2.0.0 && < 4.0.0`.
Normally (if not configured otherwise) composer will take the latest possible version matching this condition and conditions of other depending projects using this same dependency.

erayd:
> Which one do we target with this?

@shmax - exactly what @marc-mabe said above. That constraint should be fine for what we are doing, so long as you're happy that the syntax is the same for the bits you use - I'm assuming that because you approved the PR, that is the case.

shmax:
I was talking more about master vs one of our version branches (eg. 5.x.x, 6.0.dev). Master it is, then.

erayd:
@shmax Always master. Merging PRs directly into 5.x.x causes problems. This one will probably make it into my next backport release though.

Issue #463, opened 2017-11-08
Support for php7


leroy0211:
When will be there support for PHP 7? 

Because PHP already flagged PHP 5.5 as End Of Life. And php 5.6 will be End Of Life in January 2019

bighappyface:
See the Travis build configs. PHP 7 and 7.1 are included.

leroy0211:
That's weird, when I try to require this package, it will throw a php version error that the latest version allowed is `5.5.9`. Maybe there's something wrong with my environment, let me check that first.

leroy0211:
That's weird, when I try to require this package, it will throw a php version error that the latest version allowed is `5.5.9`. Maybe there's something wrong with my environment, let me check that first.

PR #465, opened 2017-11-10
override new phpcs rule

mathroc:


.travis.yml
```diff
@@ -10,6 +10,7 @@ matrix:
   fast_finish: true
   include:
     - php: 5.3
+      dist: precise
```
mathroc: because https://docs.travis-ci.com/user/reference/trusty#PHP-images

erayd:
Won't forcing this to post, rather than simply disabling it, potentially introduce logic errors in the future? Changing the order of an increment / decrement operation in any situation that assigns the result somewhere is asking for trouble, as they are not equivalent.

Can we turn this rule off completely so that it leaves increment / decrement alone? I don't trust php-cs-fixer to recognize assignment situations, and it's introduced problems with this kind of assumption in the past.

erayd:
Note the comments above regarding assignment also include using the result as input to another logic operation, regardless of whether a variable is explicitly set.

.php_cs.dist
```diff
@@ -24,6 +24,7 @@ $config
         'pre_increment' => false,
         'trailing_comma_in_multiline_array' => false,
         'simplified_null_return' => false,
+        'increment_style' => false,
```
erayd: Thanks :-)

mathroc:
that’s true, it disabled it now, unfortunately it seems there is issue when running `composer install` with some version of php. might be related to #463 ?

erayd:
> ...unfortunately it seems there is issue when running composer install with some version of php. might be related to #463 ?

Hooray, more dependency issues! Will take a look, hang on.

.travis.yml
erayd: This is already fixed in 6.0.0-dev.

mathroc:
my bad, I even saw the comment https://github.com/justinrainbow/json-schema/pull/390#issuecomment-339419771 a few days ago and didn’t remember…, so that’s much better now

I don’t have much time to help more so I just wanted to do this to get some work of your plate (even if it’s not much), thx for the all work ;)

erayd:
All good - I appreciate your taking care of this :-).

PR #470, opened 2017-11-28
Don't use PHP's fmod

### LINKED ISSUE #469: False negatives with real values of multipleOf
I use the `multipleOf` property to limit numbers to two decimal places. For example:

"turnover": {
  "type": "number",
  "multipleOf": 0.01
}

When passed a value of `5000000` it fails the constraint and returns an error.  That number is valid for the constraint so this is a false negative.

PeterMead:
This fixes #469.

### DIFF
--- a/src/JsonSchema/Constraints/NumberConstraint.php
+++ b/src/JsonSchema/Constraints/NumberConstraint.php
@@ -71,7 +71,7 @@ class NumberConstraint extends Constraint
     private function fmod($number1, $number2)
     {
         $number1 = abs($number1);
-        $modulus = fmod($number1, $number2);
+        $modulus = ($number1 - round($number1 / $number2) * $number2);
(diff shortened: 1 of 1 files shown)


PeterMead:
Well I didn't want to criticise the existing code, but yes that function is a bit of a mess.  That's mostly because it's trying to work around the issues with PHP's fmod but if we're not using that any more we can simplify the function.  I think it's still too complicated for a one liner and if it's used in two places it's better to keep the code in one place.

bighappyface:
Per #469 I think it would be nice to have a test demonstrating the false negative.

bighappyface:
@erayd should I hold out for a test on this one?

erayd:
@bighappyface I'm happy with it as-is. It would be handy to have a test at some point, but I don't think we should be delaying this fix for it, noting the lack of resources.

PR #471, opened 2017-11-28
Don't cast multipleOf to be an integer for the error message

### LINKED ISSUE #468: Error message for multipleOf no longer shows decimal places
I use the `multipleOf` property to limit numbers to two decimal places.  For example:
```lang=json
"turnover": {
  "type": "number",
  "multipleOf": 0.01
}
```

Prior to [v5.2.0s](https://github.com/justinrainbow/json-schema/releases/tag/5.2.0), and specifically #364, the error message was:
`"Must be a multiple of 0.01"`

It is now:
`"Must be a multiple of 0"`

PeterMead:
This fixes #468.


erayd:
Good catch - thanks!

shmax:
Thanks for fixing my goof!

PR #476, opened 2017-12-30
Fix: Mark check() and coerce() as deprecated

localheinz:
This PR

* [x] marks `Validator::check()` and `Validator::coerce()` as deprecated

### DIFF
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -217,7 +217,7 @@ if (isset($arOptions['--dump-schema'])) {
 
 try {
     $validator = new JsonSchema\Validator();
-    $validator->check($data, $schema);
+    $validator->validate($data, $schema);
 
     if ($validator->isValid()) {
         if(isset($arOptions['--verbose'])) {
--- a/demo/demo.php
+++ b/demo/demo.php
@@ -6,7 +6,7 @@ $data = json_decode(file_get_contents('data.json'));
 
 // Validate
 $validator = new JsonSchema\Validator();
-$validator->check($data, (object) array('$ref' => 'file://' . realpath('schema.json')));
+$validator->validate($data, (object) array('$ref' => 'file://' . realpath('schema.json')));
 
 if ($validator->isValid()) {
     echo "The supplied JSON validates against the schema.\n";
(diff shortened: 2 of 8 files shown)


bin/validate-json
```diff
@@ -217,7 +217,7 @@ if (isset($arOptions['--dump-schema'])) {
 
 try {
     $validator = new JsonSchema\Validator();
-    $validator->check($data, $schema);
+    $validator->validate($data, $schema);
```
localheinz: Probably makes sense to use `Validator::validate()` here, too.

demo/demo.php
```diff
@@ -6,7 +6,7 @@
 
 // Validate
 $validator = new JsonSchema\Validator();
-$validator->check($data, (object) array('$ref' => 'file://' . realpath('schema.json')));
+$validator->validate($data, (object) array('$ref' => 'file://' . realpath('schema.json')));
```
localheinz: Probably makes sense to use `Validator::validate()` here, too.

tests/Constraints/SelfDefinedSchemaTest.php
```diff
@@ -71,6 +71,6 @@ public function testInvalidArgumentException()
     {
         $v = new Validator();
         $this->setExpectedException('\JsonSchema\Exception\InvalidArgumentException');
-        $v->check(json_decode('{}'), json_decode(''));
+        $v->validate(json_decode('{}'), json_decode(''));
```
erayd: You cannot use `validate()` in this situation, as the first argument is a reference. If you want to change this, you'll need to decode the document into an intermediate variable first, and then pass the variable.

tests/ValidatorTest.php
```diff
@@ -36,7 +36,7 @@ public function testBadAssocSchemaInput()
         $validator->validate($data, $schema);
     }
 
-    public function testCheck()
+    public function testDeprecatedCheckStillValidates()
```
erayd: Method name needs clarification - validation in this test is supposed to fail, but the new name falsely implies that success is the intended outcome.

localheinz: How about I rename both to `test.*DelegatesToValidate()`?

localheinz:
@erayd 

Updated!

localheinz:

PR #477, opened 2017-12-30
Enhancement: Implicitly enable no_unused_imports fixer

localheinz:
This PR

* [x] implicitly enables the `no_unused_imports` fixer
* [x] runs `composer style-fix`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -15,7 +15,6 @@ $config
         'array_syntax' => array('syntax' => 'long'),
         'binary_operator_spaces' => false,
         'concat_space' => array('spacing' => 'one'),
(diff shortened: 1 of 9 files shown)


.php_cs.dist
```diff
@@ -15,7 +15,6 @@ $config
         'array_syntax' => array('syntax' => 'long'),
         'binary_operator_spaces' => false,
         'concat_space' => array('spacing' => 'one'),
-        'no_unused_imports' => false,
```
localheinz: Since `no_unused_imports` is enabled as part of the `@Symfony` rule set, we implicitly enable it by removing this rule here.

localheinz:

PR #478, opened 2017-12-30
Fix: Remove unused argument

localheinz:
This PR

* [x] removes an unused argument

Somewhat related to #477.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,7 +1,7 @@
 <?php
 
 $finder = new PhpCsFixer\Finder();
-$config = new PhpCsFixer\Config('json-schema', 'json-schema style guide');
+$config = new PhpCsFixer\Config('json-schema');
 $finder->in(__DIR__);
 
 /* Based on ^2.1 of php-cs-fixer */


.php_cs.dist
```diff
@@ -1,7 +1,7 @@
 <?php
 
 $finder = new PhpCsFixer\Finder();
-$config = new PhpCsFixer\Config('json-schema', 'json-schema style guide');
```
localheinz: The `$description` argument was removed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2244.

erayd: Why have you also removed `$name`, when only `$description` has been deprecated?

.php_cs.dist
```diff
@@ -1,7 +1,7 @@
...
+$config = new PhpCsFixer\Config('json-schema');
```
localheinz: The name is still here, @erayd.

erayd: My mistake, sorry! I looked at the diff and thought I saw the whole line removed.

localheinz: All good, @erayd, thank you!

localheinz:

PR #479, opened 2017-12-30
Fix: No need to specify path to bin directory

localheinz:
This PR

* [x] removes `vendor/bin` from scripts defined in `composer.json`

💁‍♂️ For reference, see https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands:

>Note: Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are easily accessible. In this example no matter if the `phpunit` binary is actually in `vendor/bin/phpunit` or `bin/phpunit` it will be found and executed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,10 +57,10 @@
         }
     },
     "scripts": {
-        "test" : "vendor/bin/phpunit",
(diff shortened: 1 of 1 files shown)


localheinz:

PR #480, opened 2017-12-30
Fix: No need to update composer itself

localheinz:
This PR

* [x] removes a run of `composer self-update` from the `before_install` section

💁‍♂️ That's already happing implicitly on Travis, no need to do it twice. See for example https://travis-ci.org/justinrainbow/json-schema/jobs/301016931#L476-L479.


localheinz:

PR #482, opened 2017-12-30
Enhancement: Add .editorconfig

localheinz:
This PR

* [x] adds an `.editorconfig`

💁‍♂️ For reference, see http://editorconfig.org.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_size = 2
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 /docs               export-ignore
 /tests              export-ignore
+.editorconfig       export-ignore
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_size = 2
```
localheinz: I believe this configuration reflects what is *mostly* used within this project.

erayd: While these settings may reflect current usage, they aren't necessarily desirable, and we should probably have a discussion around indentation particularly.

I'd rather like to see the entire project move to a 4-space indent, unless anyone else objects. What we have now is largely historical, and comes from a very old codebase with none of the original maintaners remaining.

localheinz: Sounds good, @erayd!

bighappyface: @localheinz shall we move to 4 space indention and move this forward?

localheinz: @bighappyface 

Can do! Been busy today with tax preparation. Taking a look tomorrow!

localheinz: @bighappyface 

Updated the indentation! 👍

PR #484, opened 2017-12-30
Fix: Remove rule for ignoring bin directory

localheinz:
This PR

* [x] removes a rule for ignoring the `bin` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 /vendor
-/bin
-!/bin/validate-json
 coverage
 .buildpath
 .project


.gitignore
```diff
@@ -1,6 +1,4 @@
 /vendor
-/bin
-!/bin/validate-json
```
localheinz: Since this file is checked in, but I couldn't find any reason why *anything* would end up in `bin/`, I think it's best to remove both rules here.

erayd: I suspect this was to avoid accidental checkins of e.g. JSON, PHP etc. files that people have put in the `bin` directory for testing.

I'd prefer that the `-/bin` rule remain, unless there's good reason to nuke it, as it prevents some obvious opportunities for making mistakes. `-/bin/validate-json` can certainly go though, as the file is already checked in.

erayd: I suspect this was to avoid accidental checkins of e.g. JSON, PHP etc. files that people have put in the `bin` directory for testing.

I'd prefer that the `-/bin` rule remain, unless there's good reason to nuke it, as it prevents some obvious opportunities for making mistakes. `-/bin/validate-json` can certainly go though, as the file is already checked in.

localheinz: Not sure about that, @erayd. To be honest, the [`.ignore`](https://plugins.jetbrains.com/plugin/7495--ignore) plugin for PhpStorm alerted me initially:

screen shot 2017-12-31 at 13 24 18

Apparently, this seems a bit like an error on the side of the plugin, as we have a rule that excludes `bin/validate-json`. However, if we apply this patch

```diff
diff --git a/.gitignore b/.gitignore
index e5ad919..7132403 100644
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /vendor
 /bin
-!/bin/validate-json
 coverage
 .buildpath
 .project
```

then it looks even more like an error that `bin/validate-json` is checked in. 

Then maybe I'll just close the PR for now.

erayd: Maybe just leave it as-is then :-). If your IDE suggests removing already-checked-in files because they are mentioned in .gitignore, then IMO the rule preventing that behaviour from impacting a legitimate file is worth keeping.

PR #485, opened 2017-12-30
Fix: Case mismatch

localheinz:
This PR

* [x] fixes cases where cases mismatch


localheinz:

PR #486, opened 2017-12-30
Fix: Remove unused argument from method call

localheinz:
This PR

* [x] removes an unused argument from a method call


localheinz:

PR #489, opened 2017-12-30
Fix: Remove unused parameter

localheinz:
This PR

* [x] removes an unused parameter


localheinz:

PR #490, opened 2017-12-31
Fix: Keep rules sorted in .php_cs.dist

localheinz:
This PR

* [x] keeps rules in `.php_cs.dist` sorted

💁‍♂️ This ideally makes it more obvious where to add new rules (in order).


localheinz:

PR #491, opened 2017-12-31
 Enhancement: Enable no_alias_function and trailing_comma_in_multiline_array fixer

localheinz:
This PR

* [x] enables the `no_alias_function` as well as the `trailing_comma_in_multiline_array` fixer
* [x] runs `composer style-fix`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/2.9#usage:

> **no_alias_functions** [`@Symfony:risky`]
>
>Master functions shall be used instead of aliases.
>
>Risky rule: risky when any of the alias functions are overridden.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -29,6 +29,7 @@ $config
         'phpdoc_types_order' => array('null_adjustment' => 'none', 'sort_algorithm' => 'none'),
     ))
     ->setFinder($finder)
+    ->setRiskyAllowed(true)
 ;
 
 return $config;
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -14,6 +14,7 @@ $config
         'array_syntax' => array('syntax' => 'long'),
         'binary_operator_spaces' => false,
         'concat_space' => array('spacing' => 'one'),
+        'no_alias_functions' => true,
         'no_useless_else' => true,
         'no_useless_return' => true,
         'ordered_imports' => true,
--- a/src/JsonSchema/Constraints/TypeConstraint.php
+++ b/src/JsonSchema/Constraints/TypeConstraint.php
@@ -134,10 +134,10 @@ class TypeConstraint extends Constraint
             return implode($delimiter, $elements);
         }
         $lastElement  = array_slice($elements, -1);
-        $firsElements = join($delimiter, array_slice($elements, 0, -1));
+        $firsElements = implode($delimiter, array_slice($elements, 0, -1));
         $implodedElements = array_merge(array($firsElements), $lastElement);
 
-        return join(" $listEnd ", $implodedElements);
+        return implode(" $listEnd ", $implodedElements);
     }
 
     /**


.php_cs.dist
```diff
@@ -15,18 +15,20 @@ $config
...
         'phpdoc_order' => true,
         'phpdoc_summary' => false,
         'pre_increment' => false,
-        'trailing_comma_in_multiline_array' => false,
         'simplified_null_return' => false,
+        'trailing_comma_in_multiline_array' => false,
         'yoda_style' => null,
-        'increment_style' => false,
     ))
+    ->setRiskyAllowed(true)
```
erayd: I find this concerning - there have been a number of times in the past when php-cs-fixer has shipped changes that resulted in it altering the execution logic of the project code, rather than making purely cosmetic changes. Thankfully, this caused the testsuite to fail, so nothing broken made it through review, but still. Anything that php-cs-fixer labels risky, or has the potential to change execution logic, worries me.

.php_cs.dist
```diff
@@ -15,18 +15,20 @@ $config
...
         'no_useless_else' => true,
         'no_useless_return' => true,
         'ordered_imports' => true,
         'phpdoc_no_package' => false,
         'phpdoc_order' => true,
         'phpdoc_summary' => false,
         'pre_increment' => false,
-        'trailing_comma_in_multiline_array' => false,
         'simplified_null_return' => false,
+        'trailing_comma_in_multiline_array' => false,
```
erayd: While we're moving it, can we please turn this rule on? A trailing comma is desirable, as it makes git blame more obvious.

.php_cs.dist
erayd: To clarify - php-cs-fixer has altered logic with *no warning, and no changes in configuration*. This behaviour is IMO unacceptable. I don't have an issue with the join -> implode changes it's made this time, but if we're going to commit this, I'd like some certainty it won't make changes in the future that break things (e.g. by replacing deprecated function names that are required for PHP 5.3 compatibility).

.php_cs.dist
localheinz: Sure, can do!

.php_cs.dist
localheinz: That's why I would suggest to pin `friendsofphp/php-cs-fixer` to at least a specific minor version, what do you think?

erayd: I think pinning is a very good idea. It doesn't entirely solve the issue, but it will help.

@mathroc - php-cs-fixer is yours, how do you feel about this?

bighappyface:
@mathroc please see the conversation on the initial commit of this PR - https://github.com/justinrainbow/json-schema/pull/491/commits/47406002f7da3100b842bac1a15901f20606c96c#r159148409

mathroc:
@erayd @bighappyface I think there's been a misunderstanding, why do you think php-cs-fixer is mine ?

erayd:
@mathroc @bighappyface This is my error sorry - looks like I tagged the wrong person. The person who was supposed to have been tagged was @alcohol.

alcohol:
Uhm, mine? What?

erayd:
@alcohol Yours, because you added it to this project in the first place. I was wanting to know what your thoughts are on whether pinning it is an acceptable solution to the problem of php-cs-fixer making unexpected and unannounced changes to the execution logic, as I assume you had reasons for setting it up the way you did originally.

@localheinz Testing this PR against the latest version would be helpful. It's been close to a year since this PR was initially opened, and I don't have time to do that at present (plus I'm not terribly familiar with php-cs-fixer anyway).

alcohol:
Oh, I see.

Having read up on the discussion (including the hidden comments), I can add the following: pinning is fine. Though it would be a good idea to occasionally check for updates using `composer outdated --direct` and determining what changes have been made and whether or not they warrant an update (and if yes, what would have to be configured additionally etc).

localheinz:
@erayd 

Just rebased and pushed, let me know what I can do!

alcohol:
You did not actually pin php-cs-fixer to a minor version yet though?

I would suggest changing `"friendsofphp/php-cs-fixer": "^2.1",` to `"friendsofphp/php-cs-fixer": "~2.2.20",`

localheinz:
@alcohol 

No, that was never my intention with this PR, but can do in a separate PR.

bighappyface:
@alcohol @localheinz PHP-CS-Fixer has been pinned per #537

localheinz:
@bighappyface 

Rebased!

bighappyface:
@localheinz please rebase once more. I merged #535.

localheinz:
@bighappyface 

Done! 🤓

bighappyface:
@erayd @shmax any objections to merging?

localheinz:
Closing in favour of sorting out other issues with the `php-cs-fixer` requirements and configuration first.

PR #494, opened 2018-02-13
[BUGFIX] Apply defaults in $ref'ed property / item definitions

### LINKED ISSUE #473: Default value not applied when it's defined through $ref
**Schema (YML):**
```yml
definitions:
  limit:
    type: integer
    minimum: 1
    maximum: 100
    default: 10

test:
  type: object
  properties:
    query:
      type: string
    limit:
      $ref: "#/definitions/limit"
    offset:
      type: integer
      default: 0
```

**JSON:**
```json
{
  "query": "test"
}
```

**Validation:**
```php
$validator->validate($request, $schema, Constraint::CHECK_MODE_APPLY_DEFAULTS);
```

**Request after validation:**
```json
{
  "query": "test",
  "offset": 0
}
```

It seems validator doesn't apply default value in $refs.
Version: 5.2.6
PHP: 7.1

erayd:
## What
Dereference property and array item definitions prior to checking for a default value. Fixes #473.

## Why
Because the spec says:
 1. We should be ignoring siblings of `$ref` (so `default` isn't valid here anyway); and
 2. There might be a `default` defined in the reference target.

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -265,6 +265,7 @@ class UndefinedConstraint extends Constraint
             }
             // $value is an array, and items are defined - treat as plain array
             foreach ($items as $currentItem => $itemDefinition) {
+                $itemDefinition = $this->factory->getSchemaStorage()->resolveRefSchema($itemDefinition);
                 if (
                     !array_key_exists($currentItem, $value)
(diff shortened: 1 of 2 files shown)


erayd:
@shmax Any chance of a code review?

shmax:
LGTM!

erayd:
Thanks guys :-)

PR #505, opened 2018-04-13
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ For reference, see https://github.com/localheinz/composer-normalize (which uses `justinrainbow/json-schema` internally to normalize `composer.json` using the underlying JSON schema).


localheinz:

PR #507, opened 2018-04-17
Add support for "const"

### LINKED ISSUE #506: Support for "const"?
I just noticed that the `Validator` class does not appear to support the `const` keyword as specified in [section 6.1.3](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.3) of the spec. Is this intentional?

Given a few pointers where to start, I'd also be prepared to open a PR to add support for this feature.

```
$schema = [
        "properties" => [
                "foo" => [
                        "type" => "string",
                        "const" => "foo"
                ]
        ]
];

$doc = ["foo" => "bar"];

$validator->validate($doc, $schema);

// expected: false, actual: true
var_dump($validator->isValid());
```

martin-helmich:
As discussed in #506, this PR adds support for the `const` keyword as first specified in [section 6.24 of draft 06](https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-6.24).

Fixes #506

### DIFF
--- a/src/JsonSchema/Constraints/ConstConstraint.php
+++ b/src/JsonSchema/Constraints/ConstConstraint.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the JsonSchema package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace JsonSchema\Constraints;
+
+use JsonSchema\ConstraintError;
+use JsonSchema\Entity\JsonPointer;
+
+/**
+ * The ConstConstraint Constraints, validates an element against a constant value
+ *
+ * @author Martin Helmich <martin@helmich.me>
+ */
+class ConstConstraint extends Constraint
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function check(&$element, $schema = null, JsonPointer $path = null, $i = null)
+    {
+        // Only validate const if the attribute exists
+        if ($element instanceof UndefinedConstraint && (!isset($schema->required) || !$schema->required)) {
+            return;
+        }
+        $const = $schema->const;
+
+        $type = gettype($element);
+        $constType = gettype($const);
+
+        if ($this->factory->getConfig(self::CHECK_MODE_TYPE_CAST) && $type == 'array' && $constType == 'object') {
+            if ((object) $element == $const) {
+                return;
+            }
+        }
+
+        if ($type === gettype($const)) {
+            if ($type == 'object') {
+                if ($element == $const) {
+                    return;
+                }
+            } elseif ($element === $const) {
+                return;
+            }
+        }
+
+        $this->addError(ConstraintError::CONSTANT(), $path, array('const' => $schema->const));
+    }
+}
--- a/tests/Constraints/CoerciveTest.php
+++ b/tests/Constraints/CoerciveTest.php
@@ -165,6 +165,13 @@ class CoerciveTest extends VeryBaseTestCase
             'integer', 'boolean', true, true
         );
 
+        // #51 check boolean coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"boolean","const":false}}}',
+            '{"propertyOne":"false"}',
+            'string', 'boolean', false, true
+        );
+
         foreach ($types as $toType => $testCases) {
             foreach ($testCases as $testCase) {
                 $tests[] = array(
--- a/tests/Constraints/CoerciveTest.php
+++ b/tests/Constraints/CoerciveTest.php
@@ -144,6 +144,27 @@ class CoerciveTest extends VeryBaseTestCase
             'string', 'integer', 42, true
         );
 
+        // #48 check boolean coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"boolean","const":false}}}',
+            '{"propertyOne":"false"}',
+            'string', 'boolean', false, true
+        );
+
+        // #49 check boolean coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"boolean","const":true}}}',
+            '{"propertyOne":"true"}',
+            'string', 'boolean', true, true
+        );
+
+        // #50 check boolean coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"boolean","const":true}}}',
+            '{"propertyOne":1}',
+            'integer', 'boolean', true, true
+        );
+
         foreach ($types as $toType => $testCases) {
             foreach ($testCases as $testCase) {
                 $tests[] = array(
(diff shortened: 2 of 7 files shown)


tests/Constraints/ConstTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function getValidTests()
+    {
+        return array(
+            array(
+                '{"value":"bar"}',
+                '{
+                  "type":"object",
+                  "properties":{
+                    "value":{"type":"string","const":"bar"}
```
shmax: I'm a little curious about the potential use cases covered by this implementation. Ten minutes ago I had never heard of the `const` keyword, but from what I've read since then its main utility comes from interacting with the `$data` concept proposed in v5. Given that we don't currently support `$data` or relative JSON pointers, how much use is there in supporting a keyword that guarantees that an input value is precisely equal to some other value?

martin-helmich: I did not find any mention of the `$data` concept in the recent spec v7. From my understanding, `const` is presumably nothing more than a more expressive alternative to a one-element `enum`:

>**6.1.3. const**
>The value of this keyword MAY be of any type, including null.
>
>An instance validates successfully against this keyword if its value is equal to the value of the keyword.

My use case for `const` is when using a "constant" property as a type discriminator when using polymorphic types:

```
{
  "$oneOf": [
    {
      "properties": {
        "paymentType": {"type": "string", "const": "directDebit"},
        "iban": {"type": ...}
       }
    },
    {
      "properties": {
        "paymentType": {"type: "string", "const": "creditCard"},
        "cardNumber": {...}
      }
  ]
}
```

In this case, documents like `{"payment": {"paymentType": "directDebit", "iban": "DE1234567890"}}` and `{"payment": {"paymentType": "creditCard", "cardNumber": "123456789"}}` would match the schema, while something like `{"payment": {"paymentType": "creditCard", "iban": "DE1234567890"}}` would not.

shmax: I see. Okay. 👍

shmax:
Does this play nicely with coercion? It would be great if you could add a section somewhere that does a few checks (possibly in `CoerciveTests.php`...)

martin-helmich:
@shmax Done; I've added a few test cases on how I think `const` should work with coercion enabled; they all passed without any further changes.

Btw, I noticed that the `$tests = []` assignment (formerly line 133) in the `CoerciveTests.php` caused test cases 39-45 to never be executed; I've re-enabled them, as they were all passing.

tests/Constraints/CoerciveTest.php
```diff
@@ -130,7 +130,20 @@ public function dataCoerceCases()
...
 
-        $tests = array();
+        // #46 check coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"string","const":"42"}}}',
+            '{"propertyOne":42}',
+            'integer', 'string', "42", true
+        );
+
+        // #46 check coercion with "const"
```
shmax: // #47

martin-helmich: Fixed

shmax:
> Btw, I noticed that the $tests = [] assignment (formerly line 133) in the CoerciveTests.php caused test cases 39-45 to never be executed; 

Egad, it's been that way for a year! Nice catch, thanks much.

Not to be a pest, but can you add one more coercive test to check boolean type? Just to be sure--php is fine with `==` checks on string/numbers, but not string/boolean.

martin-helmich:
>Not to be a pest, but can you add one more coercive test to check boolean type? Just to be sure--php is fine with == checks on string/numbers, but not string/boolean.

:+1: Done

tests/Constraints/CoerciveTest.php
```diff
@@ -130,7 +130,41 @@ public function dataCoerceCases()
...
+
+        // #49 check boolean coercion with "const"
+        $tests[] = array(
+            '{"properties":{"propertyOne":{"type":"boolean","const":true}}}',
+            '{"propertyOne":"true"}',
+            'string', 'boolean', true, true
+        );
+
+        // #50 check boolean coercion with "const"
+        $tests[] = array(
```
shmax: I was hoping to see a false...

```php
("true" == true) // true
("false" == false) // false
```

src/JsonSchema/Constraints/ConstConstraint.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if ($this->factory->getConfig(self::CHECK_MODE_TYPE_CAST) && $type == 'array' && $constType == 'object') {
+            if ((object) $element == $const) {
+                return;
+            }
+        }
+
+        if ($type === gettype($const)) {
+            if ($type == 'object') {
+                if ($element == $const) {
```
erayd: This does not correctly implement deep comparison of objects - it results in members being compared using `==`.

Could you please change this implementation to ensure type-strict comparison of object properties. You will probably need to iterate them; PHP lacks a suitable native comparison operator that can do it all in one go.

tests/Constraints/CoerciveTest.php
martin-helmich: Here you go...

shmax: Beautiful, thank you

tests/Constraints/CoerciveTest.php
```diff
@@ -130,7 +130,41 @@ public function dataCoerceCases()
...
+            '{"properties":{"propertyOne":{"type":"boolean","const":true}}}',
+            '{"propertyOne":1}',
+            'integer', 'boolean', true, true
+        );
```
erayd: Please also add some deep object tests. With and without mismatched property types.

src/JsonSchema/Constraints/ConstConstraint.php
martin-helmich: Following that argumentation, the `enum` implementation (which was my inspiration for this class) should probably be changed as well -- it uses the same comparison.

erayd: If enum is similarly broken, then yes - I agree with you that it needs fixing.

erayd:
@shmax I reckon 6.0.0 (which this PR is against), and I'll also backport it for 5.3.0.

erayd:
@bighappyface I notice that you have merged this, without a fix for the deep object comparison issue I raised above. Are you intending for that to be fixed in a separate PR? @martin-helmich seems to have abandoned this one without making the necessary change.

Raphy:
Is that normal the version 5.2.8 and 5.X.X do not have this feature ?!
Browsing files from these version and I can't find this feature :(

erayd:
@Raphy That is correct, yes. This is a draft-06 feature, but is currently incomplete, and not backported.

This library officially supports draft-03 and draft-04, but does not get fully support newer versions of the spec. Draft-06 is largely similar to draft-04, but there are a few differences, and this is one of them.

PR #515, opened 2018-05-21
Fix missing deref.

### LINKED ISSUE #514: Array of items doesn't expand, and so validation check is wrong.
TL:DR version

There is a missing expansion of a reference when the type is an array of items. PR incoming.

----------------------------

Original version

https://stackoverflow.com/questions/50390920/how-to-correctly-validate-array-of-objects-using-justinrainbow-jsonschema/50407155#50407155

I have code that correctly validates an article returned from an endpoint that returns single articles. I'm pretty sure it's working correctly as it gives a validation error when I deliberately don't include a required field in the article.

I also have this code that tries to validate an array of articles returned from an endpoint that returns an array of articles. However, I'm pretty sure that isn't working correctly, as it always says the data is valid, even when I deliberately don't include a required field in the articles.

How do I correctly validate an array of data against the schema?

The full test code is below as a standalone runnable test. Both of the tests should fail, however only one of them does.

```
<?php

declare(strict_types=1);

error_reporting(E_ALL);

require_once __DIR__ . '/vendor/autoload.php';


// Return the definition of the schema, either as an array
// or a PHP object
function getSchema($asArray = false)
{
    $schemaJson = <<< 'JSON'
{
  "swagger": "2.0",
  "info": {
    "termsOfService": "http://swagger.io/terms/",
    "version": "1.0.0",
    "title": "Example api"
  },
  "paths": {
    "/articles": {
      "get": {
        "tags": [
          "article"
        ],
        "summary": "Find all articles",
        "description": "Returns a list of articles",
        "operationId": "getArticleById",
        "produces": [
          "application/json"
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Article"
              }
            }
          }
        },
        "parameters": [
        ]
      }
    },
    "/articles/{articleId}": {
      "get": {
        "tags": [
          "article"
        ],
        "summary": "Find article by ID",
        "description": "Returns a single article",
        "operationId": "getArticleById",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "articleId",
            "in": "path",
            "description": "ID of article to return",
            "required": true,
            "type": "integer",
            "format": "int64"
          }
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "$ref": "#/definitions/Article"
            }
          }
        }
      }
    }
  },
  "definitions": {
    "Article": {
      "type": "object",
      "required": [
        "id",
        "title"
      ],
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        },
        "title": {
          "type": "string",
          "description": "The title for the link of the article"
        }
      }
    }
  },
  "schemes": [
    "http"
  ],
  "host": "example.com",
  "basePath": "/",
  "tags": [],
  "securityDefinitions": {
  },
  "security": [
    {
      "ApiKeyAuth": []
    }
  ]
}
JSON;

    return json_decode($schemaJson, $asArray);
}

// Extract the schema of the 200 response of an api endpoint.
function getSchemaForPath($path)
{
    $swaggerData = getSchema(true);
    if (isset($swaggerData["paths"][$path]['get']["responses"][200]['schema']) !== true) {
        echo "response not defined";
        exit(-1);
    }

    return $swaggerData["paths"][$path]['get']["responses"][200]['schema'];
}

// JsonSchema needs to know about the ID used for the top-level
// schema apparently.
function aliasSchema($prefix, $schemaForPath)
{
    $aliasedSchema = [];

    foreach ($schemaForPath as $key => $value) {
        if ($key === '$ref') {
            $aliasedSchema[$key] = $prefix . $value;
        }
        else if (is_array($value) === true) {
            $aliasedSchema[$key] = aliasSchema($prefix, $value);
        }
        else {
            $aliasedSchema[$key] = $value;
        }
    }
    return $aliasedSchema;
}


// Test the data matches the schema.
function testDataMatches($endpointData, $schemaForPath)
{
    // Setup the top level schema and get a validator from it.
    $schemaStorage = new \JsonSchema\SchemaStorage();
    $id = 'file://example';
    $swaggerClass = getSchema(false);
    $schemaStorage->addSchema($id, $swaggerClass);
    $factory = new \JsonSchema\Constraints\Factory($schemaStorage);
    $jsonValidator = new \JsonSchema\Validator($factory);

    // Alias the schema for the endpoint, so JsonSchema can work with it.
    $schemaForPath = aliasSchema($id, $schemaForPath);

    // Validate the things
    $jsonValidator->check($endpointData, (object)$schemaForPath);

    // Process the result
    if ($jsonValidator->isValid()) {
        echo "The supplied JSON validates against the schema definition: " . \json_encode($schemaForPath) . " \n";
        return;
    }

    $messages = [];
    $messages[] = "End points does not validate. Violations:\n";
    foreach ($jsonValidator->getErrors() as $error) {
        $messages[] = sprintf("[%s] %s\n", $error['property'], $error['message']);
    }

    $messages[] = "Data: " . \json_encode($endpointData, JSON_PRETTY_PRINT);

    echo implode("\n", $messages);
    echo "\n";
}



// We have two data sets to test. A list of articles.

$articleListJson = <<< JSON
[
  {
      "id": 19874
  },
  {
      "id": 19873
  }
]
JSON;
$articleListData = json_decode($articleListJson);


// A single article
$articleJson = <<< JSON
{
  "id": 19874
}
JSON;
$articleData = json_decode($articleJson);


// This passes, when it shouldn't as none of the articles have a title
testDataMatches($articleListData, getSchemaForPath("/articles"));


// This fails correctly, as it is correct for it to fail to validate, as the article doesn't have a title
testDataMatches($articleData, getSchemaForPath("/articles/{articleId}"));
```

Danack:
Fixes #514

I realise this could do with some tests.

Please can you direct me to a similar existent test to copy and modify, or at least say where an appropriate test might go.

### DIFF
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -64,6 +64,10 @@ class CollectionConstraint extends Constraint
      */
     protected function validateItems(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
+        if (is_array($schema->items) && array_key_exists('$ref', $schema->items)) {
+            $schema->items = $this->factory->getSchemaStorage()->resolveRefSchema((object)$schema->items);
+        };
+
         if (is_object($schema->items)) {
             // just one type definition for the whole array
 
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -65,8 +65,8 @@ class CollectionConstraint extends Constraint
     protected function validateItems(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
         if (is_array($schema->items) && array_key_exists('$ref', $schema->items)) {
-            $schema->items = $this->factory->getSchemaStorage()->resolveRefSchema((object)$schema->items);
-        };
+            $schema->items = $this->factory->getSchemaStorage()->resolveRefSchema((object) $schema->items);
+        }
 
         if (is_object($schema->items)) {
             // just one type definition for the whole array


Danack:
broken test appears not related to change.

erayd:
> Please can you direct me to a similar existent test to copy and modify, or at least say where an appropriate test might go.

Tests for this should probably go in `ArraysTest.php`. You'll see many example tests in the same folder.

tarunlalwani:
@Danack , @erayd , seems like the PR is not needed as such. The issue was in the usage of library.

The validation was run like

```php
$jsonValidator->check($endpointData, (object)$schemaForPath);
```

Which worked for a single object but not for a array type schema. And the usage should have been like

```php
$jsonValidator->check($endpointData, (object)$schemaForPath);
```

See updated answer on 

https://stackoverflow.com/a/50449472/2830850

Probably a thought, that you should have some check that people don't mix `stdClass` and `array` or something that could be just documented to avoid such issues in future

erayd:
@tarunlalwani Well spotted - thanks. I've had a play with the code in #514, and it seems that `aliasSchema` is causing the problem by returning assoc arrays instead of objects. If it's adjusted to return an object, then the validator behaves as expected.

Danack:
I'll close this PR as it was based on me being dumb.

Issue #517, opened 2018-05-23
Const & enum deep comparison is not strictly typed


erayd:
The `const` implementation from #507, and the existing `enum` implementation, [does not properly compare objects](https://github.com/justinrainbow/json-schema/pull/507#discussion_r182204595). Objects are compared in a loosely-typed manner (`==`), which also results in a loosely-typed comparison of all members.

This behavior is incorrect, and results in spec violations. For example, `"string" == true` returns true, despite being clearly invalid - both the type and content are completely different.

shmax:
Would you like me to take a crack at this? I feel like you need some video game time...

erayd:
Please! I'm flat out right now...

shmax:
Righto, then. I have some time on Sunday.

PR #518, opened 2018-05-29
Strict Enum/Const Object Checking

shmax:
Fix for #517 

@erayd correctly deduced that the non-strict object comparisons happening in `EnumConstraint` and `ConstConstraint` were opening type comparison holes on their member values. For example, the following test would pass:

```
array(
	'{
		"value": {
			"foo": "12"
		}
	}',
	'{
		"type": "object",
		"properties": {
			"value": {
				"type": "any", 
				"enum": [
					6, 
					"foo", 
					[], 
					true, 
					{
						"foo": 12
					}
				]
			}
		}
	}'
)
```

PHP rules governing object type equality checking are somewhat infamous. From the [docs](http://php.net/manual/en/language.oop5.object-comparison.php):

> When using the comparison operator (==), object variables are compared in a simple manner, namely: Two object instances are equal if they have the same attributes and values (values are compared with ==), and are instances of the same class.

> When using the identity operator (===), object variables are identical if and only if they refer to the same instance of the same class.

It provides no mechanism for doing strict member-by-member comparisons, so for this PR I drew in a library dedicated to this kind of thing, [Parity](https://github.com/IcecaveStudios/parity).


shmax:
@erayd @martin-helmich please review

erayd:
Needs a code style fix, but otherwise looks good :+1:. Thanks!

src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -53,7 +53,7 @@ public function addError(ConstraintError $constraint, JsonPointer $path = null,
             'pointer' => ltrim(strval($path ?: new JsonPointer('')), '#'),
             'message' => ucfirst(vsprintf($message, array_map(function ($val) {
                 if (is_scalar($val)) {
-                    return $val;
+                    return var_export($val, true);
```
shmax: Extra credit: I noticed that the const error string was producing this message for bool type:

> Does not have a value equal to 1

instead of 

> Does not have a value equal to true

shmax:
So what do we do with this?

erayd:
@shmax I don't think the hhvm build is related, but it needs fixing. I'm happy for that to be in a different PR. However, the failing tests on all builds introduced in your latest commit do need fixing.

shmax:
Oh, heh, didn't realize the other tests had broken. Thanks much.

Issue #533, opened 2018-09-25
Checking for oneOf/anyOf with different types lose value when coercing types


LordWeedlle:
When checking for a schema which get a `oneOf` or an `anyOf` we validate type for each schema possible in `JsonSchema/Constraints/TypeConstraint::validateType()`.
But inside this function, when using `Constraint::CHECK_MODE_COERCE_TYPES` option, we cast the `$value` (passed by reference to the function) to the type defined in schema which result in a `NULL` value set to `$value` this variable isn't castable to the required type.
When then checking for the good type, `$value` is `NULL` so it cannot validates.

Here is an example:

> We wants to validate that we send a string `'all'` or an array containing some strings.

```PHP
$value = [
    'Monday'
];

$schema = [
    'type' => 'object',
    'properties' => [
        'daysOfWeek' => [
            'oneOf' => [
                [
                    'type' => 'string',
                    'enum' => 'all'
                ],
                [
                    'type'     => 'array',
                    'minItems' => 1,
                    'items'    => ['type' => 'string']
                ]
            ]
        ]
];

$validator  = new Validator;
// Cast to object
$value = json_decode(json_encode($value));

$validator->validate($value, $schema, Constraint::CHECK_MODE_COERCE_TYPES);
```

When validating the validator will cast $value to `string` in JsonSchema/Constraints/TypeConstraint.php:l222 (`$value = $this->toString($value);`). $value is then gonna equals `NULL` as the array cannot be casted to string. When then checking for second oneOf schema, it cannot validates.

erayd:
@shmax Do you have time to look at this? I'm flat out at present; I'm unlikely to be able to investigate this until after October 19th, although I'm happy to do code reviews.

erayd:
Thanks - you're a legend :-).

PR #535, opened 2018-09-26
return original value when no cast

shmax:
Fix for #533 

This change aligns `TypeConstraint::toString` with the other casting methods in that if it can't figure out a cast to do, then it simply returns the original value.

### DIFF
--- a/tests/Constraints/CoerciveTest.php
+++ b/tests/Constraints/CoerciveTest.php
@@ -39,7 +39,7 @@ class CoerciveTest extends VeryBaseTestCase
                 array('array',      '[45]',         '45',           true),  // #5
                 array('object',     '{"a":"b"}',    null,           false), // #6
                 array('array',      '[{"a":"b"}]',  null,           false), // #7
-                array('array',      '[1,2]',  		array(1,2),     false), // #8
+                array('array',      '[1,2]',  		array(1, 2),     false), // #8
             ),
             'integer' => array(
                 array('string',     '"45"',         45,             true),  // #9
@@ -173,7 +173,7 @@ class CoerciveTest extends VeryBaseTestCase
             'string', null, null, false
         );
 
-		foreach ($types as $toType => $testCases) {
+        foreach ($types as $toType => $testCases) {
             foreach ($testCases as $testCase) {
                 $tests[] = array(
                     sprintf('{"properties":{"propertyOne":{"type":"%s"}}}', strtolower($toType)),
--- a/tests/Constraints/CoerciveTest.php
+++ b/tests/Constraints/CoerciveTest.php
@@ -39,53 +39,47 @@ class CoerciveTest extends VeryBaseTestCase
                 array('array',      '[45]',         '45',           true),  // #5
                 array('object',     '{"a":"b"}',    null,           false), // #6
                 array('array',      '[{"a":"b"}]',  null,           false), // #7
+                array('array',      '[1,2]',  		array(1,2),     false), // #8
             ),
             'integer' => array(
-                array('string',     '"45"',         45,             true),  // #8
-                array('integer',    '45',           45,             true),  // #9
-                array('boolean',    'true',         1,              true),  // #10
-                array('boolean',    'false',        0,              true),  // #11
-                array('NULL',       'null',         0,              true),  // #12
-                array('array',      '["-45"]',      -45,            true),  // #13
-                array('object',     '{"a":"b"}',    null,           false), // #14
-                array('array',      '["ABC"]',      null,           false), // #15
+                array('string',     '"45"',         45,             true),  // #9
+                array('integer',    '45',           45,             true),  // #10
+                array('boolean',    'true',         1,              true),  // #11
+                array('boolean',    'false',        0,              true),  // #12
+                array('NULL',       'null',         0,              true),  // #13
+                array('array',      '["-45"]',      -45,            true),  // #14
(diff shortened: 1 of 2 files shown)


shmax:
@erayd @LordWeedlle

tests/Constraints/CoerciveTest.php
```diff
@@ -39,55 +39,49 @@ public function dataCoerceCases()
                 array('array',      '[45]',         '45',           true),  // #5
                 array('object',     '{"a":"b"}',    null,           false), // #6
                 array('array',      '[{"a":"b"}]',  null,           false), // #7
+                array('array',      '[1,2]',  		array(1,2),     false), // #8
```
shmax: This is the only new code in this file. Everything else is just redoing the numbers (why do we have numbers in the comments, again?)

LordWeedlle: I presume numbers are there for debug purpose, so when phpunit tells you testXXX with dataset #23 failed, you don't have to count arrays to find the one failing XD

erayd: Yep, that's what they are for.

shmax: But that's awful, isn't it? Shouldn't PHPUnit be giving us line numbers? I just tried it locally, and it does report line numbers, but they're wrong. Is this really considered idiomatic test code, or are we just cargo culting?

shmax: > We can set names for the datasets

That sounds more reasonable, and a little more in line with what you see in javascript testing (`describe` blocks, and such). Still don't understand why PHPUnit can come up with the comment to echo back to us and not the line number, but I'm not interested enough to pursue it. 

Anyway, are we all set here? Seems like a pretty straightforward fix...

erayd: > Still don't understand why PHPUnit can come up with the comment to echo back to us and not the line number...

It's not echoing the comment. It's giving the array index of the test. It has no idea which line the test was originally defined on unfortunately, it can only tell us the line on which the assertion failed. The comments are just to make life easier for us when we need to find which test it's complaining about.

> Anyway, are we all set here? Seems like a pretty straightforward fix...

Yep, looks good to me. Thanks :-).

shmax: > It's not echoing the comment.

Then how is it useful to you?

erayd: > Then how is it useful to you?

It lets me find the exact test in the source file when PHPUnit complains. If it says test 15 from file X failed, I can go straight to test 15 without having to count them, because I can just look for the comment.

shmax: Right, but you said it's not echoing the comment. If it's not echoing the comment, then there's nothing for you to eyeball. If it is echoing the comment, and you're only seeing that comment, then it seems like a line number could be derived, somehow.

erayd: I think you are misunderstanding how this works.

PHPUnit reports the array index of the test, but doesn't say anything about what that test is, or which line number it's defined on. In order to locate where the failing test is defined, one must manually search through the source file and find the test with the correct array index.

The comments are manually-added references to the array index of each test. That way, instead of having to manually count through the entire array definition in order to locate which test PHPUnit just complained about, you can simply look for the comment with the correct number.

Ideally, we'd use associative indexes with a useful description in them, which would render the comments redundant - I wanted to do this at one point, but there was a technical reason why it wasn't possible at that time - I do not recall why, but it may have been a compatibility conflict. If you would like to look into this again, you are welcome to do so - it may well be that whatever the issue was, has been resolved since I last looked into it for this project.

shmax: Oh, is _that_ what's going on? Egad, it's even worse than I thought. Well, shouldn't be too tough to fix. Here's one idea:

```
protected function _addTest($config) {
  $config[] = debug_backtrace()[0]['line'];
  return $config;
}
```

Use it like this:
```
$tests[] = $this->_addTest(array(
            '{"properties":{"propertyOne":{"type":["number", "string"]}}}',
            '{"propertyOne":42}',
            'integer', 'integer', 42, true
        ));
```

Then you modify your batch test method to expect the line number:
```
public function testCoerceCases($schema, $data, $startType, $endType, $endValue, $valid, $lineNumber, $extraFlags = 0, $assoc = false)
    {
```

From there I suppose you could just write some kind of helper assert method that appends the line number.

erayd: > Oh, is *that* what's going on? Egad, it's even worse than I thought.

Yep, it's complete shit, and I'm responsible for introducing it. Sorry ;-).

Your idea is a good one; I hadn't considered using the backtrace. I think we should look at providing that info as part of an assoc index first (rather than using a special function), because that would be IMO a better place for that information if it's now usable, and will result in PHPUnit printing it alongside the error. If we still can't set an assoc index though, then we'll need to modify the batch tests as you mention above.

shmax: Sure, either idea sounds like an improvement. 👍

LordWeedlle:
I didn't took time to thank you for your rapidity and your work @shmax @erayd so I do it now =)
Thank you !
It's always a pleasure to use a repository where people are reactive !

shmax:
> thank you for your rapidity and your work @shmax @erayd

Pleasure to be of service. 😄

PR #544, opened 2018-11-06
feat: Exclude certain endpoints from media type validation

estahn:
- Fixes justinrainbow/json-schema#543 and justinrainbow/json-schema#540

### DIFF
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -33,6 +33,14 @@ class UriRetriever implements BaseUriRetrieverInterface
     );
 
     /**
+     * @var array A blacklist for media type ceheck exclusion
+     */
+    protected $mediaTypeBlacklist = array(
+        'http://json-schema.org/',
+        'https://json-schema.org/'
+    );
+
+    /**
      * @var null|UriRetrieverInterface
      */
     protected $uriRetriever = null;
@@ -43,6 +51,16 @@ class UriRetriever implements BaseUriRetrieverInterface
      * @see loadSchema
      */
     private $schemaCache = array();
+
+    /**
+     * Adds an endpoint to the media type validation blacklist
+     *
+     * @param string $endpoint
+     */
+    public function addBlacklistedEndpoint($endpoint)
+    {
+        $this->mediaTypeBlacklist[] = $endpoint;
+    }
 
     /**
      * Guarantee the correct media type was encountered
@@ -65,9 +83,10 @@ class UriRetriever implements BaseUriRetrieverInterface
             return;
         }
 
-        if (substr($uri, 0, 23) == 'http://json-schema.org/') {
-            //HACK; they deliver broken content types
-            return true;
+        for ($i = 0, $iMax = count($this->mediaTypeBlacklist); $i < $iMax; $i++) {
+            if (stripos($uri, $this->mediaTypeBlacklist[$i]) === 0) {
+                return true;
+            }
         }
 
         throw new InvalidSchemaMediaTypeException(sprintf('Media type %s expected', Validator::SCHEMA_MEDIA_TYPE));
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -33,7 +33,7 @@ class UriRetriever implements BaseUriRetrieverInterface
     );
 
     /**
-     * @var array A blacklist for media type ceheck exclusion
+     * @var array A blacklist for media type check exclusion
      */
     protected $mediaTypeBlacklist = array(
         'http://json-schema.org/',
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -33,9 +33,9 @@ class UriRetriever implements BaseUriRetrieverInterface
     );
 
     /**
-     * @var array A blacklist for media type check exclusion
+     * @var array A list of endpoints for media type check exclusion
      */
-    protected $mediaTypeBlacklist = array(
+    protected $allowedInvalidContentTypeEndpoints = array(
         'http://json-schema.org/',
         'https://json-schema.org/'
     );
@@ -53,13 +53,13 @@ class UriRetriever implements BaseUriRetrieverInterface
     private $schemaCache = array();
 
     /**
-     * Adds an endpoint to the media type validation blacklist
+     * Adds an endpoint to the media type validation exclusion list
      *
      * @param string $endpoint
      */
-    public function addBlacklistedEndpoint($endpoint)
+    public function addInvalidContentTypeEndpoint($endpoint)
     {
-        $this->mediaTypeBlacklist[] = $endpoint;
+        $this->allowedInvalidContentTypeEndpoints[] = $endpoint;
     }
 
     /**
@@ -83,8 +83,8 @@ class UriRetriever implements BaseUriRetrieverInterface
             return;
         }
 
-        for ($i = 0, $iMax = count($this->mediaTypeBlacklist); $i < $iMax; $i++) {
-            if (stripos($uri, $this->mediaTypeBlacklist[$i]) === 0) {
+        foreach ($this->allowedInvalidContentTypeEndpoints as $endpoint) {
+            if (strpos($uri, $endpoint) === 0) {
                 return true;
             }
         }
--- a/tests/Uri/UriRetrieverTest.php
+++ b/tests/Uri/UriRetrieverTest.php
@@ -330,13 +330,35 @@ EOF;
         $this->assertEquals('454f423bd7edddf0bc77af4130ed9161', md5(json_encode($schema)));
     }
 
-    public function testJsonSchemaOrgMediaTypeHack()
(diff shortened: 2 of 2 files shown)


estahn:
@erayd I'm flexible on the naming of the `attribute` and `method` name. I just didn't have better ideas. Let me know your thoughts.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -32,6 +32,14 @@ class UriRetriever implements BaseUriRetrieverInterface
         '|^https?://json-schema.org/draft-(0[34])/schema#?|' => 'package://dist/schema/json-schema-draft-$1.json'
     );
 
+    /**
+     * @var array A blacklist for media type ceheck exclusion
```
erayd: Typo in comment ('ceheck')

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -32,6 +32,14 @@ class UriRetriever implements BaseUriRetrieverInterface
...
+     */
+    protected $mediaTypeBlacklist = array(
```
erayd: More descriptive name please - perhaps something like 'allowInvalidContentType'.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -44,6 +52,16 @@ class UriRetriever implements BaseUriRetrieverInterface
      */
     private $schemaCache = array();
 
+    /**
+     * Adds an endpoint to the media type validation blacklist
+     *
+     * @param string $endpoint
+     */
+    public function addBlacklistedEndpoint($endpoint)
```
erayd: Again with the naming. I don't feel that this is a blacklist - it's a list of exceptions.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -65,9 +83,10 @@ public function confirmMediaType($uriRetriever, $uri)
             return;
         }
 
-        if (substr($uri, 0, 23) == 'http://json-schema.org/') {
-            //HACK; they deliver broken content types
-            return true;
+        for ($i = 0, $iMax = count($this->mediaTypeBlacklist); $i < $iMax; $i++) {
```
erayd: Refactor to use `foreach` please.

src/JsonSchema/Uri/UriRetriever.php
bighappyface: spelling

estahn: fixed.

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -65,9 +83,10 @@ public function confirmMediaType($uriRetriever, $uri)
...
+            if (stripos($uri, $this->mediaTypeBlacklist[$i]) === 0) {
```
erayd: Should be case-sensitive.

tests/Uri/UriRetrieverTest.php
```diff
@@ -339,6 +339,28 @@ public function testJsonSchemaOrgMediaTypeHack()
...
+    /**
+     * @expectedException \JsonSchema\Exception\InvalidSchemaMediaTypeException
+     */
+    public function testJsonSchemaOrgMediaTypeBlacklistUnknown()
+    {
+        $mock = $this->getMock('JsonSchema\Uri\UriRetriever', array('getContentType'));
+        $mock->method('getContentType')->willReturn('Application/X-Fake-Type');
+        $retriever = new UriRetriever();
+
+        $retriever->confirmMediaType($mock, 'http://iglucentral.com');
```
erayd: Use example.com for example domains please.

bighappyface:
@estahn is _blacklist_ the correct term? Wouldn't a list of allowed endpoints be a whitelist?

tests/Uri/UriRetrieverTest.php
```diff
@@ -339,6 +339,28 @@ public function testJsonSchemaOrgMediaTypeHack()
...
+    }
+
+    public function testJsonSchemaOrgMediaTypeBlacklistAdded()
+    {
+        $mock = $this->getMock('JsonSchema\Uri\UriRetriever', array('getContentType'));
+        $mock->method('getContentType')->willReturn('Application/X-Fake-Type');
+        $retriever = new UriRetriever();
+        $retriever->addBlacklistedEndpoint('http://iglucentral.com');
```
erayd: Use example.com for example domains please.

estahn:
> @estahn is _blacklist_ the correct term? Wouldn't a list of allowed endpoints be a whitelist?

Indeed, but since the list contains excluded endpoints it would be a blacklist. I'm happy to rename it to something else, e.g. `excludedMediaTypeCheckEndpoints`.

tests/Uri/UriRetrieverTest.php
estahn: fixed.

tests/Uri/UriRetrieverTest.php
estahn: fixed.

src/JsonSchema/Uri/UriRetriever.php
estahn: @erayd Can you elaborate? Which part needs to be case-sensitive?

src/JsonSchema/Uri/UriRetriever.php
estahn: @erayd As I said I'm not happy with the name either and open for suggestions. `allowInvalidContentType` doesn't match the content either I believe.

How about `allowedInvalidContentTypeEndpoints`?

erayd: That works :+1:

src/JsonSchema/Uri/UriRetriever.php
bighappyface: @estahn ☝️

src/JsonSchema/Uri/UriRetriever.php
erayd: @estahn Most of a URL is case-sensitive; only the domain and scheme are not. As this is a feature which does prefix matching on a full URL, we should be checking case. So `stripos()` is not appropriate here.

An alternative would be to parse the URL properly and separate out the origin, if you would like to distinguish between the sensitive and insensitive parts.

estahn: @erayd I see. Would you consider something like `strtolower` is ok, given it's more a theoretical issue?

erayd: @estahn No, I do not consider that to be OK. It's not just a theoretical issue; this has practical implications (it's a security problem, and this library is widely used). `strtolower` doesn't solve anything, it just transfers the case-insensitivity to a different place.

erayd: @estahn Yes, using `strpos` is OK :-).

src/JsonSchema/Uri/UriRetriever.php
estahn: done

tests/Uri/UriRetrieverTest.php
```diff
@@ -330,7 +330,7 @@ public function testRetrieveSchemaFromPackage()
         $this->assertEquals('454f423bd7edddf0bc77af4130ed9161', md5(json_encode($schema)));
     }
 
-    public function testJsonSchemaOrgMediaTypeHack()
+    public function testJsonSchemaOrgMediaTypeBlacklistDefault()
```
erayd: Naming - not a blacklist.

tests/Uri/UriRetrieverTest.php
```diff
@@ -339,6 +339,28 @@ public function testJsonSchemaOrgMediaTypeHack()
         $this->assertTrue($retriever->confirmMediaType($mock, 'http://json-schema.org/'));
     }
 
+    /**
+     * @expectedException \JsonSchema\Exception\InvalidSchemaMediaTypeException
+     */
+    public function testJsonSchemaOrgMediaTypeBlacklistUnknown()
```
erayd: Naming - not a blacklist, and nothing to do with json-schema.org.

tests/Uri/UriRetrieverTest.php
```diff
@@ -339,6 +339,28 @@ public function testJsonSchemaOrgMediaTypeHack()
...
+    {
+        $mock = $this->getMock('JsonSchema\Uri\UriRetriever', array('getContentType'));
+        $mock->method('getContentType')->willReturn('Application/X-Fake-Type');
+        $retriever = new UriRetriever();
+
+        $retriever->confirmMediaType($mock, 'http://example.com');
+    }
+
+    public function testJsonSchemaOrgMediaTypeBlacklistAdded()
```
erayd: Naming - not a blacklist, and nothing to do with json-schema.org.

tests/Uri/UriRetrieverTest.php
estahn: done

tests/Uri/UriRetrieverTest.php
estahn: done

erayd:
@bighappyface Your turn :-).

PR #545, opened 2018-11-14
Fix min/maxProperties check for invalid array input

integer:
For invalid array input validation falls with `get_object_vars() expects parameter 1 to be object, array given` error.

### DIFF
--- a/src/JsonSchema/Constraints/TypeCheck/StrictTypeCheck.php
+++ b/src/JsonSchema/Constraints/TypeCheck/StrictTypeCheck.php
@@ -31,6 +31,10 @@ class StrictTypeCheck implements TypeCheckInterface
 
     public static function propertyCount($value)
     {
+        if (!is_object($value)) {
+            return 0;
+        }
+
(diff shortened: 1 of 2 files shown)


erayd:
@shmax Is array -> object type coercion ever a thing?

shmax:
Not that I know of; in any case he didn't add his code to the coercive tests. In the test he did add he's passing an array to a schema expecting an object, so it seems natural that the test should fail, but I don't know if he's trying to fix validation so it psases, or if he's fixing some kind of exception that throws instead of an expected reported error.

erayd:
@shmax Thanks - just wanted to check that in case this code would ever be expected to count an array rather than ignoring it.

@integer Looks pretty good. Your test shouldn't include a `type` keyword though; it's important that it fails validation for the right reason.

integer:
@erayd I simplified test. 

This pull request covers invalid input which our QA team tried. This input was not valid (Is is OK) but whole program stops with PHP error. With my patch validation fails and program is still running.

Issue #547, opened 2018-11-21
Array properties with array items listed in enums does not validate properly


malinink:
For example if we have array property:
```
{"data": ["a", "b"]}
```
And validation like:
```
{
    "type": "object",
    "properties": {
        "data": {
            "type": "array",
            "items": {
                "type": "string",
                "enum": ["b", "c"]
            }
        }
    }
}
```

Validation wrongly finished with success.
I will provide PR with test for that particular case.

erayd:
Thanks - good catch! That should definitely not be passing validation.

erayd:
Fixed in master with #498, backported for 5.2.8 in #558.

rbchasesc:
@erayd Finally got to writing some test cases for these. Can be found in #559

malinink:
@erayd  tests are OK, there is no need of that Issue and PR.

erayd:
@bighappyface Could you please close #547 (fixed in #498) and #549 (superseded by #559), and merge #559.

PR #537, opened 2018-10-17
Fix: Pin friendsofphp/php-cs-fixer to minor version

localheinz:
❗️ Blocks #491.

This PR

* [x] pins `friendsofphp/php-cs-fixer` to a compatible minor version
* [x] removes unknown fixers

Follows https://github.com/justinrainbow/json-schema/pull/491#issuecomment-430517542.

PR #498, opened 2018-03-04
Remove large array type-specific performance optimisation

erayd:
## What
 * Fixes #441
 * Reduces complexity

## Why
Because the current implementation of this optimisation logic contains several serious bugs, and the project is short of maintainers. As this optimisation doesn't apply to most use-cases anyway, removing it completely and validating via the regular code path seems like the best option.

Issue #441, opened 2017-07-14
anyOf Warning


allentsai:
Hello,  I would first like to thank you all for the hard work you've put into this project, and add the disclaimer that this could very well be user error, but my schema validates with other validators, and previous versions, but I get a warning with v5.2.1.

Stripped down to what is causing the error:
Base schema:
````
{
	"$schema": "http://json-schema.org/draft-04/schema#",
	"type": "object",
	"properties": {
		"location": {
			"anyOf": [
				{
					"$ref": "./postalList.json"
				},
				{
					"$ref": "./area.json"
				}
			]
		}
	},
	"additionalProperties": false,
	"required": [
		"location"
	]
}
````
postalList.json:
````
{
	"$schema": "http://json-schema.org/draft-04/schema#",
	"type": "object",
	"properties": {
		"countryLimit": {
			"type": "string"
		},
		"type": {
			"enum": ["postalList"]
		},
		"data": {
			"type": "array",
			"items": {
				"type": "string",
				"minLength": 2
			},
			"minItems": 1,
			"uniqueItems": true
		}
	},
	"additionalItems": false
}
````
area.json:
````
{
	"$schema": "http://json-schema.org/draft-04/schema#",
	"type": "object",
	"properties": {
		"type": {
			"enum": ["area"]
		},
		"data": {
			"type": "array",
			"items": {
				"$ref": "./areas.json"
			},
			"minItems":1,
			"uniqueItems": true
		}
	},
	"additionalProperties": false
}
````

areas.json:
````
{
	"$schema": "http://json-schema.org/draft-04/schema#",
	"type": "object",
	"properties": {
		"id": {
			"type": "number",
			"minimum": 0
		},
		"name": {
			"type": "string",
			"minLength": 1
		},
		"abbr": {
			"type": "string",
			"pattern": "[A-Z]+",
			"minLength": 1
		}
	},
	"required": ["id", "name", "abbr"]
}
````
Test Code:
````
$validator = new \JsonSchema\Validator();
$value = json_encode(
    ["location" => ["type" => "area", "data" => [["id" => 1, "name" => "area name", "abbr" => "AB"]]]]
);

$result = $validator->validate(json_decode($value), (object)['$ref' => 'file://' . realpath('./test.json')]);
````

This gives me the follow error repeated twice:
````
[...]

PR #558, opened 2019-01-10
Backports for 5.2.8

erayd:
Bugfixes from [6.0.0](https://github.com/justinrainbow/json-schema/tree/master) that can be backported to 5.2.8 without breaking backwards-compatibility.

## Backported PRs
 * #470 Don't use PHP's fmod
 * #498 Remove type-specific performance optimisation (#441, #547)
 * #520 Add hhvm to allow_failures
 * #528 Comment (phpdoc) fix in bin/validate-json
 * #537 Pin php-cs-fixer to minor version
 * #544 Exclude some endpoints from media type validation
 * #545 Check invalid array input to min / max properties

## Skipped PRs
 * #471 Error message change - out of scope for 5.x.x (errors were rewritten for v6)
 * #507 Add support for "const" - out of scope for 5.x.x
 * #518 Strict Enum/Const Object Checking - out of scope for 5.x.x
 * #535 Return original value (this type of string coercion doesn't exist in 5.x.x)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -11,11 +11,9 @@ $config
         '@PSR2' => true,
         '@Symfony' => true,
         // additionally
-        'align_multiline_comment' => array('comment_type' => 'phpdocs_like'),
         'array_syntax' => array('syntax' => 'long'),
         'binary_operator_spaces' => false,
         'concat_space' => array('spacing' => 'one'),
-        'increment_style' => false,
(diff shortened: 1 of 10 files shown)


erayd:
@bighappyface Could you please merge this when you have a moment?

bighappyface:
@erayd thanks! See https://github.com/justinrainbow/json-schema/releases/tag/5.2.8

PR #565, opened 2019-02-15
Meida-Type : clean media type from unused data

AliObeidat:
We have case , with this media type : 
application/json; charset=utf-8
This media type is not valid for **confirmMediaType** function in **UriRetriever** class , so before we retrieve it , we need to be sure , we have media type without any extra data.

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/AbstractRetriever.php
+++ b/src/JsonSchema/Uri/Retrievers/AbstractRetriever.php
@@ -29,6 +29,11 @@ abstract class AbstractRetriever implements UriRetrieverInterface
      */
     public function getContentType()
     {
-        return $this->contentType;
+        if(empty($this->contentType)){
+            return $this->contentType;
+        }
+
+        // return cleaned content type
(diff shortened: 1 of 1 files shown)


DannyvdSluijs:
@AliObeidat thank you very much for you effort. Sorry this was left dangling. In an attempt to cleanup this repo we are trying to filter the issues and see which ones might be closed. Is it safe to assume this is a rather old pull request, which sadly was left unanswered. 

I've checked around a bit and discovered #350 which was closed due to inactivity and also my PR (#709) are kind of related.  Also RFC8259 has the following footnote on the charset:
> Note:  No "charset" parameter is defined for this registration. Adding one really has no effect on compliant recipients.
And searching the web for the header value `application/json; charset=utf-8` leads to results on message boards and SCM solutions where the consensus seems to aim to the correct of the header on the serving side.

Is it safe to assume since you haven't followed up on this PR since 2019 we can close it for now? Feel free to close it yourself or comments if helpful.

PR #567, opened 2019-02-26
Don't run checks which assume a defined instance against undefined

### LINKED ISSUE #566: Issue with "not" in geojson schema
I'm trying to validate a simple object against the geojson schema but it fails on the "not" in the `crs` property.

This is the test I wrote, which currently fails.

```php
public function testValidateWithNonExistedPropertyUsingNot()
{
    $schema = json_decode(file_get_contents('http://json.schemastore.org/geojson'), true);
    $data = json_decode('{"type": "FeatureCollection", "features": []}', true);

    $validator = new Validator();

    $this->assertTrue($validator->isValid());
}
```

It looks like it has something do to with "not" on properties that are not required/present in the data.

erayd:
Adds undefined check to the end of required clause.

Validation tests for an undefined instance should be limited to those
tests which are actually applicable (e.g. required, default). Tests
which assume an instance previously attempted to validate an internal
default when the instance is undefined, rather than ignoring it, which
is incorrect behavior.

Closes #566.

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -154,7 +154,7 @@ class UndefinedConstraint extends Constraint
                     $this->addError(ConstraintError::REQUIRED(), $path, array('property' => $propertyName));
                 }
             } else {
-                // If the value is both undefined and not required, skip remaining checks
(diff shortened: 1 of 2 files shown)


erayd:
@shmax Could you please review this when you have some time.

PR #563, opened 2019-02-04
Tests on php 7.3

samnela:

PR #575, opened 2019-06-02
Tests on php 7.3

SignpostMarv:
follows on from #563:

* adjusts php-cs-fixer config to preserve existing behaviour, rather than applying new rules added in php-cs-fixer
* splits the 7.0 run in two, so one can get feedback on php-cs-fixer before phpunit has finished running & so xdebug isn't loaded while php-cs-fixer runs.


SignpostMarv:
you're welcome :)

PR #585, opened 2019-08-16
 patternProperties now adds error if no match

piotrkochan:


erayd:
This PR violates the spec. `patternProperties` does not mandate a match.

PR #589, opened 2019-09-23
5.x.x: Update validate-json to use spl_autoload_register

### LINKED ISSUE #574: The validate-json binary autoload deprecation warning in PHP 7.2+
When running the validate-json command in cli using PHP 7.2 or later, the following deprecation warning shows:

```
PHP Deprecated:  __autoload() is deprecated, use spl_autoload_register() instead in
~/.composer/vendor/justinrainbow/json-schema/bin/validate-json on line 16
```

sanmai:
Fixes #574

- `spl_autoload_register` exists since PHP 5.1, so this shouldn't be a problem backward-compatibility wise.
- There's was a more drastic change (#401 & 9fe7822) in the master to fix the same root issue, but it wasn't backported to 5.x. Hence the need for this fix.

### DIFF
--- a/bin/validate-json
+++ b/bin/validate-json
@@ -13,7 +13,7 @@
  *
  * @return void
  */
-function __autoload($className)
+spl_autoload_register(function ($className)
 {
     $className = ltrim($className, '\\');
     $fileName  = '';
@@ -26,7 +26,7 @@ function __autoload($className)
     if (stream_resolve_include_path($fileName)) {
         require_once $fileName;
     }
-}
+});
 
 // support running this tool from git checkout
 if (is_dir(__DIR__ . '/../src/JsonSchema')) {


erayd:
Thank you @sanmai :-).

maks-rafalko:
Guys, could you please tag a new release with this fix?

erayd:
@maks-rafalko Yes. I want to roll a few other backported fixes in too.

sanmai:
A release with just this fix will be very appreciated because it is a major deprecation since PHP 7.2. Almost everyone these days is getting PHP 7.3 from the start, e.g. because Debian stable includes one. Thank you very much for considering this request.

erayd:
@sanmai Are you requesting a release ASAP, or are you specifically wanting a release with that fix *only* in it? Because I'm doing the backports now, and should be tagging stuff in a few minutes.

erayd:
@sanmai I've just tagged v5.2.9 with all current backports merged in.

Please let me know if you need an additional release that contains *only* the autoload fix - that seems like a rather strange thing to ask for, so I'd prefer not to do it without a good reason.

maks-rafalko:
Thank you!

sanmai:
> Please let me know if you need an additional release that contains only the autoload fix

This wasn't necessary absolutely. I meant that if you're busy with other things tagging a release with just this fix would be a big help. But else any release works. Thank you again.

PR #590, opened 2019-09-23
Fix travis PHP 5.4 and 5.5

sanmai:
Like #583

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ matrix:
     - php: 5.3
       dist: precise
     - php: 5.4
+      dist: trusty
     - php: 5.5
+      dist: trusty
     - php: 5.6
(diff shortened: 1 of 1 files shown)


erayd:
Have just reversed the merge for this sorry - it modifies the 5.x.x branch, and I didn't initially notice this. Excepting very rare situations where a patch affects *only* the 5.x.x branch, we don't allow that - patches must apply in the master branch first, and are then backported via another PR (helps keep everything very consistent).

I will roll a backport release for you this week which includes #583.

sanmai:
No questions about this PR. I'm more concerned with #589 if you ask.

erayd:
@sanmai I have merged that one - thank you for catching that; `_autoload()` has been deprecated for a while!

PR #593, opened 2019-10-17
Fix composer warning

andytruong:
Small one, just hide composer warning:

```
Deprecation warning: require-dev.json-schema/JSON-Schema-Test-Suite is invalid, it should not contain uppercase characters. Please use json-schema/json-schema-test-suite instead. Make sure you fix this as Composer 2.0 will error.
```


guilliamxavier:
Looks like this was eventually done in #614

PR #596, opened 2019-11-11
Update .gitattributes

reedy:


erayd:
No sorry - these files are used for CI. We won't be removing them from archives.

reedy:
Why are they used by CI? Does CI not use the git repo?

erayd:
@reedy Travis should, but whatever it's doing caused it to think the php-cs-fixer config was missing when your change was applied.

The `bin` folder is also supposed to be distributed; it should remain regardless of CI issues.

erayd:
Although taking another look at that, it could also be that the pin is broken for some reason. The changes it's making don't look like the same mess it made last time the config was insufficient.

PR #597, opened 2019-11-12
Add .gitattributes to .gitattributes

reedy:


erayd:
Thanks for this. Looks like there's definitely a pin failure, as per my comment on your other PR.

reedy:
> Thanks for this. Looks like there's definitely a pin failure, as per my comment on your other PR.

I'll have a look at it in more detail later on today (ie after I've had some sleep) and see if I can help

erayd:
Will merge once that's sorted; we're configured with a merge block on any PR that fails CI. In this case, the PR is fine.

Issue #598, opened 2019-11-12
php-cs-fixer is not properly pinned


erayd:
php-cs-fixer is still trying to apply new, breaking changes to the source - in this case it's removing parts of the documentation comments - despite #537 which was explicitly intended to prevent this.

/CC @localheinz

erayd:
/CC @SignpostMarv also, as you were the last person to touch this in #575.

SignpostMarv:
Is this regarding that annoying habit php-cs-fixer has of removing non-starred lines in multi-line docblock tags ?

erayd:
@SignpostMarv  Based on a quick look at what it's doing, it seems to be removing argument lines with a technically invalid type - which are deliberately the way they are to be more descriptive (e.g. `string|null`).

localheinz:
@erayd 

Can you give an example?

SignpostMarv:
@erayd [this seems to be fine](https://gist.github.com/SignpostMarv/35ebb5220b76b115539b38d06d6d1dc2). There'll be rules to configure to change the behaviour. However....

`@param tags` with no description that are directly match the are redundant.
```diff
  10) tests\Constraints\TypeTest.php (no_superfluous_phpdoc_tags, phpdoc_align)
      ---------- begin diff ----------
--- Original
+++ New
@@ -69,8 +69,7 @@
     /**
      * Helper to assert an error message
      *
-     * @param string         $expected
-     * @param TypeConstraint $actual
+     * @param string $expected
      */
     private function assertTypeConstraintError($expected, TypeConstraint $actual)
     {

```
mid-line indentation is a matter of taste, although additionally I'm going to hazard a guess most of these should be more specific than `mixed`.
```diff
   5) src\JsonSchema\Constraints\UndefinedConstraint.php (no_superfluous_phpdoc_tags, phpdoc_align)
      ---------- begin diff ----------
--- Original
+++ New
@@ -55,10 +55,9 @@
     /**
      * Validates the value against the types
      *
-     * @param mixed       $value
-     * @param mixed       $schema
-     * @param JsonPointer $path
-     * @param string      $i
+     * @param mixed  $value
+     * @param mixed  $schema
+     * @param string $i
      */

```
As with the single-typed redundant line removal, `T|null` is a redundant line when the actual parameter matches the specified type, i.e. `JsonPointer|null $path`
```diff
[...]

PR #599, opened 2019-11-13
Fix: Disable no_superfluous_phpdoc_tags fixer

localheinz:
This PR

* [x] disables the `no_superfluous_phpdoc_tags` fixer, which is implicitly enabled by using the `@Symfony` rules

Related to #598.

💁‍♂ Personally, I would not use the `@Symfony` ruleset, but instead configure fixers one by one. I do this with [`localheinz/php-cs-fixer-config`](https://github.com/localheinz/php-cs-fixer-config), which allows me to re-use a configuration for `friendsofphp/php-cs-fixer` across projects. Together with https://dependabot.com, this works excellent, by the way, in keeping projects updated with a consistent configuration.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -14,6 +14,7 @@ $config
         'array_syntax' => array('syntax' => 'long'),
         'binary_operator_spaces' => false,
         'concat_space' => array('spacing' => 'one'),
+        'no_superfluous_phpdoc_tags' => false,
         'no_useless_else' => true,
         'no_useless_return' => true,
         'ordered_imports' => true,


erayd:
Thanks for this, and for your other PR. I don't have an opinion one way or the other on the `@Symfony` ruleset - would you like to continue with exceptions from it, or would you rather change to the approach you've outlined above? I'm quite happy with either.

My main consideration is that it doesn't do unexpected things to the source without warning; provided that criteria is met, then the configuration can be whatever people prefer.

localheinz:
@erayd 

Perhaps #600 is even more critical?

I'd be happy to help sorting out the fixer configuration!

PR #600, opened 2019-11-13
Fix: Do not allow installation of multiple versions of friendsofphp/php-cs-fixer

localheinz:
This PR

* [x] disallows installation of multiple versions of `friendsofphp/php-cs-fixer`
* [x] removes unknown fixers from the configuration for `php-cs-fixer`
* [x] installs `friendsofphp/php-cs-fixer` using a separate `composer.json` to work around PHP version incompatibilities

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "icecave/parity": "1.0.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "~2.2.20||~2.15.1",
+        "friendsofphp/php-cs-fixer": "~2.2.20",
         "json-schema/JSON-Schema-Test-Suite": "1.2.0",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -32,7 +32,7 @@
         "icecave/parity": "1.0.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "~2.2.20||~2.15.1",
```
localheinz: Allowing the installation of different versions will result in different results when running

```
$ composer style-fix
```

which is probably not what we want.

erayd:
@localheinz Is this currently a work-in-progress, or are you waiting for me to review it?

localheinz:
@erayd 

This is waiting for your review!

🤓

PR #604, opened 2019-11-22
Fixed a problem with oneOf matches, as if there are multiple matches,…

rawveg:
… this would - by default - throw a failure, which is incorrect

### DIFF
--- a/src/JsonSchema/Constraints/UndefinedConstraint.php
+++ b/src/JsonSchema/Constraints/UndefinedConstraint.php
@@ -365,7 +365,7 @@ class UndefinedConstraint extends Constraint
                     // other schema options in the OneOf field.
                 }
             }
-            if ($matchedSchemas !== 1) {
+            if ($matchedSchemas <= 0) {
                 $this->addErrors(array_merge($allErrors, $startErrors));
                 $this->addError(ConstraintError::ONE_OF(), $path);
             } else {


erayd:
According to the spec, `oneOf` must match *exactly* one subschema. Your change would make `oneOf` behave like `anyOf`, which is incorrect.

rawveg:
However, when you have the following in the Schema:-

```
                "oneOf": [
                  {
                    "required": [
                      "type",
                      "use_backend",
                      "conditional",
                      "and"
                    ]
                  },
                  {
                    "required": [
                      "type",
                      "use_backend",
                      "conditional",
                      "or"
                    ]
                  },
                  {
                    "required": [
                      "type",
                      "use_backend",
                      "conditional",
                      "and",
                      "or"
                    ]
                  }
                ]
```

And the JSON to match **exactly one rule**:-

```
{
    "type": "use_backend",
    "use_backend": "randombackend-419617",
    "conditional": "if",
    "and": [
        {
            "index": 41
        },
        {
            "index": 41,
            "negate": true
        }
    ],
    "or": [
        {
            "index": 1,
            "negate": true
        },
        {
            "index": 1
        },
        {
            "index": 1
        }
    ]
}
```
Matching fails and $matchedSchemas instead of being 1, is actually 3 (even though there is only one exact match). Thoughts?

erayd:
You are attempting to use conditionals, which are not part of the draft-04 spec. This library does not yet support newer versions 

If you rewrite your schema using draft-04 syntax, then it should work.

erayd:
Actually, my mistake - your data has conditionals, not the schema. Stand by.

erayd:
The problem is that your data contains both `and` and `or` properties. This is a match for all three conditions:

 - require `or` to be present
 - require `and` to be present
 - require both of `and` / `or` to be present

The validator is correctly reporting that your data violates the `oneOf` constraint.

PR #606, opened 2019-12-20
Further travis tweaks

GrahamCampbell:
1. Since we've pinned HHVM to 3.18, and the tests are passing, we should probably take it out of allowed failures now.

2. Nightly was failing to install dependencies (because, quite rightly, people have marked their code as not working on PHP 8 yet). We can instruct composer to ignore PHP version requirements.

3. No interaction mode is already enabled due to travis setting an env variable that enforces it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,14 @@ matrix:
       dist: trusty
   allow_failures:
     - php: nightly
-    - php: hhvm-3.18
 
 before_install:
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Ping @erayd. :)

NB Test failures are unrelated, and should be fixed in a separate PR - the failures are just code style due to php-cs-fixer not being pinned.

erayd:
Thanks @GrahamCampbell :-).

PR #609, opened 2020-01-09
suppress php-7.4 warning in CollectionConstraint.php: not an array

lzsiga:
In some cases php-7.4 warns that in CollectionContraint.php (lines 93 and 122) `$schema->items` used as array, but it is actually just 'true' (boolean)
Solution: check it beforehand with 'is_array'

Note: this problem happens when I use draft-07 for schema validation: eg
`validate-json json-schema-draft-07.json json-schema-draft-07.json`

### DIFF
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -90,7 +90,7 @@ class CollectionConstraint extends Constraint
         } else {
             // Defined item type definitions
             foreach ($value as $k => &$v) {
-                if (array_key_exists($k, $schema->items)) {
+                if (is_array($schema->items) && array_key_exists($k, $schema->items)) {
                     $this->checkUndefined($v, $schema->items[$k], $path, $k);
                 } else {
                     // Additional items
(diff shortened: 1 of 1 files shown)


erayd:
This library doesn't support draft-07 (it tops out at draft-04). Attempting to use 07 schemas with it may work if you stick with 04 syntax, but you can't trust it if you are actaully using 07 features, as keywords which were not present in 04 are simply ignored as per the 04 spec.

I'm conflicted about whether we should address this at all, because a boolean is [not a valid type](https://json-schema.org/draft-04/json-schema-validation.html#rfc.section.5.3.1) for `items` anyway:

> The value of "items" MUST be either an object or an array. If it is an object, this object MUST be a valid JSON Schema. If it is an array, items of this array MUST be objects, and each of these objects MUST be a valid JSON Schema.

Catching this kind of thing is why schema validation exists; if that feature is enabled then it should be flagged there before validation even starts (although I'm inclined to say yes, because the performance impact of this particular check is minimal). However, if we do check for this in validation, then the outcome should be a validation failure (because the schema is invalid), not simply ignoring the issue as this PR currently does.

I would also prefer it if this check were done once, [where `validateItems()` is called from](https://github.com/justinrainbow/json-schema/blob/3921012b92bdda17ec9b5a9f62cd25b2d0591b70/src/JsonSchema/Constraints/CollectionConstraint.php#L53), rather than multiple times inside `validateItems()`.

lzsiga:
Thank you for your reply, please give me some time to understand the details. (Note: I've already found out I was wrong: this warning isn't php-7.4 specific.)

lzsiga:
Hi,
For now, let it sink, I might come back later with better a suggestion.
Thank you!

PR #611, opened 2020-01-27
Update gitattributes to ignore more files

KasperFranz:


KasperFranz:
Seems like a later PR was accepted which implemented the above items

PR #612, opened 2020-01-29
Fix coding style via PHP-CS-Fixer

peter279k:
# Changed log
- Fixing coding style via `php-cs-fixer`.

### DIFF
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -57,10 +57,9 @@ class CollectionConstraint extends Constraint
     /**
      * Validates the items
      *
-     * @param array            $value
-     * @param \stdClass        $schema
-     * @param JsonPointer|null $path
-     * @param string           $i
+     * @param array     $value
+     * @param \stdClass $schema
+     * @param string    $i
      */
     protected function validateItems(&$value, $schema = null, JsonPointer $path = null, $i = null)
     {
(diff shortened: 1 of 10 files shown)


erayd:
This PR just runs php-cs-fixer, but does not fix the problem with php-cs-fixer and removes wanted content as a result.

Php-cs-fixer has a track record of making undesirable changes without warning when updated - that needs to be fixed first. Most of the existing codebase should be compliant with the desired behavior.

If you would like to re-style things I'm happy to consider that, but it's really important that we don't lose data / change execution logic as a result, both of which php-cs-fixer has tried to do to this project in the past. We also need a good reason for the style change.

erayd:
See #599 and #600 - perhaps you guys could collaborate on those?

peter279k:
@localheinz, do you have any idea about this :)?

localheinz:
@peter279k 

Thanks for reaching out - I believe we should sort out #600 first, and then see about enabling or configuring rules as we see fit, ideally one by one (unless we feel brave and want to enable and configure a bunch of them at once).

peter279k:
> @peter279k
> 
> Thanks for reaching out - I believe we should sort out #600 first, and then see about enabling or configuring rules as we see fit, ideally one by one (unless we feel brave and want to enable and configure a bunch of them at once).

Hi @localheinz, I agree with you and we can focus on #600 firstly and configuring rules on next steps.

PR #619, opened 2020-02-26
Fix PHP 8 deprecated warnings

matbech:
This fixes issue https://github.com/justinrainbow/json-schema/issues/618


phpdeveloper9:
With sample code, why json result not validating field name from json schema?

erayd:
@phpdeveloper9 Please open a new issue that fully describes your problem, and includes example schema / data that can be used to replicate it.

Issue #617, opened 2020-02-26
Specify supported JSON Schema version


janlucaklees:
I would love to see the actual JSON Schema version that is supported by the library somewhere in the readme. Sure I can go look in the tests and infer that only Draft-4 of the JSON Schema is supported, but you need to be very desperate to do that.

Context: I spent a whole lot of time figuring out why 'contains' is not working for an array. Not sure, but maybe with `Constraint::CHECK_MODE_VALIDATE_SCHEMA` set, it should warn you about properties in your JSON Schema that are not supported.

I would be open to "implement" those changes myself and open a PR for them to be merged.

erayd:
Good suggestion. A PR for this would be appreciated :-).

yannickroger:
@erayd @janlucaklees I think we can close that one 😉

janlucaklees:
@yannickroger Yess, thank you for reminding me!

PR #620, opened 2020-03-12
Added note about supported Draft versions

janlucaklees:
When I got started working with this library I got frustrated when my Schema didn't work as expected. It wasn't until I looked at the test files that I realized I was using features of a Draft that was not supported by this library. A note like the one I added would have saved me a lot of time trying to debug my Schema.

So I am adding it now for everyone that comes after me.

This PR is a result of #617

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Latest Stable Version](https://poser.pugx.org/justinrainbow/json-schema/v/stable.png)](https://packagist.org/packages/justinrainbow/json-schema)
 [![Total Downloads](https://poser.pugx.org/justinrainbow/json-schema/downloads.png)](https://packagist.org/packages/justinrainbow/json-schema)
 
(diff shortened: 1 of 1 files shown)


yannickroger:
Hi @erayd 
Could this be merged (you asked for it in the issue). We just lost a day of work because of a draft-07 schema. We read the documentation and didn't see any limitation. I think this is very useful.

janlucaklees:
@yannickroger I would love to merge this too, but I get the feeling that because of the failed tests, this PR doesnt get so much attention. Unfortunately. I don't really get what error is occuring. I suspect it is something with the php linter, but looking at the logs I am not even sure where the error is. It shouldn't matter however, as I just touched the Readme.

I'll still try to keep this up to date.

erayd:
@janlucaklees Yep, just the linter.

@yannickroger Thanks for reminding me about this PR - it's now merged 🙂.

PR #621, opened 2020-03-12
Fixed php-cs-fixer config

GrahamCampbell:


erayd:
Hang on - why is this closed?

PR #641, opened 2020-08-26
fix typo

staabm:
`strpos` expects args in the order $haystack, $needle.

additionally `strpos` can never return `-1` but will return `false` when nothing can be found.

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -188,7 +188,7 @@ class FormatConstraint extends Constraint
         // which will fail the above string comparison because the passed
         // $datetime may be '2000-05-01T12:12:12.123Z' but format() will return
         // '2000-05-01T12:12:12.123000Z'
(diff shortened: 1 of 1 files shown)


guilliamxavier:
The fix is correct, but actually this block is dead code (see https://github.com/justinrainbow/json-schema/pull/234#issuecomment-284917771, "[it] was simply merged in error after the problem was already solved"), so it should probably be deleted (in a second commit)

/ping @erayd

DannyvdSluijs:
Looking into the comment from @guilliamxavier he is correct. The "typo" was introduced in ffc856db on 18/02/2016 whereas in 6474265b on 03/03/2016 the calling code of the `validateDateTime` with an `format` argument containing an `u` was removed making this check void. So removing the code block would be best.

@staabm thank you very much for your efforts though. Sadly the PR was never followed up. In an attempt to cleanup this repo we are trying to filter the issues and see which ones might be closed. Based on the above conclusion is it safe to assume this can be closed? Feel free to close it yourself with some comments if helpful. A new PR #710 was just created to remove the dead code.

staabm:
Thank you guys for the analysis

PR #646, opened 2020-11-17
URI Retriever: follow http redirects using Curl by default

soyuka:
The CurlURIRetriever will fallback to the FileGetContents one with relative paths as I understood that file paths and actual URIs were using the same retriever. 

Please let me know if my assumption is false.

### DIFF
--- a/src/JsonSchema/Uri/Retrievers/Curl.php
+++ b/src/JsonSchema/Uri/Retrievers/Curl.php
@@ -41,6 +41,7 @@ class Curl extends AbstractRetriever
         curl_setopt($ch, CURLOPT_URL, $uri);
         curl_setopt($ch, CURLOPT_HEADER, true);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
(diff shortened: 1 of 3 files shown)


src/JsonSchema/Uri/Retrievers/Curl.php
```diff
@@ -36,44 +40,47 @@ public function __construct()
...
      * @param string $response cURL HTTP response
      *
      * @return bool Whether the Content-Type header was found or not
      */
-    protected function fetchContentType($response)
+    private function fetchContentType($headers, $curlInfo)
     {
-        if (0 < preg_match("/Content-Type:(\V*)/ims", $response, $match)) {
-            $this->contentType = trim($match[1]);
+        if (0 < preg_match_all("/content-type: (.+?(?=;|$))/ims", $headers, $match, PREG_SET_ORDER)) {
```
soyuka, line 82: https://www.phpliveregex.com/p/xYE#tab-preg-match-all

DannyvdSluijs:
@soyuka We are working on bringing back this library in an active and maintained state.
I was looking into this PR and was kind of confused over what it is that is being solved, are your proposing a fix for an issue of are you adding a new feature? In its current state it seems to be doing a mix of both.

Could you explain a bit more in detail what this PR aims to do?

DannyvdSluijs:
Closing the PR for now due to lack of feedback. You're very welcome to get back and reopen this PR with additional feedback.

Issue #649, opened 2021-01-01
Bug: Unescaped slash throws error, when pattern contains slash


a-r-m-i-n:
Here you don't escape slashes in expression: https://github.com/justinrainbow/json-schema/blob/fa4d2d3c1e40a222ded125b679891086c9a6c7d6/src/JsonSchema/Constraints/FormatConstraint.php#L200

Here you do: https://github.com/justinrainbow/json-schema/blob/fa4d2d3c1e40a222ded125b679891086c9a6c7d6/src/JsonSchema/Constraints/StringConstraint.php#L43

The first check should do the same here. 

Thanks for this great lib! And a happy new year! 🥳

erayd:
Thanks for reporting this. A PR that fixes it would be very welcome 🙂.

ktomk:
would you consider to file a PR on this? @a-r-m-i-n

ktomk:
Just asking, as I may also pick a take on it. For that I would be also interested in the "throws error" details, specifically how to provoke an error and what message of that error mentioned is. /cc @a-r-m-i-n 

The `str_replace` looks fishy to me as `preg_match` is in use. Me thinks better of `preg_quote` for that.

a-r-m-i-n:
Here we go. I just copied the preg_match from the second place. If you want to replace the used function, feel free. I think this will at least solve the issue I currently have, when I try to use Composer's JSON schema. The "name" property there, got a check pattern for "abc/def", including the unescaped / which is also used as delimiter before.

erayd:
> The `str_replace` looks fishy to me as `preg_match` is in use. Me thinks better of `preg_quote` for that.

That method uses `preg_match` *to check whether `$regex` is a valid regular expression*, by attempting to compile it and checking for an error condition. It does not use a regular expression to validate a string. Using `preg_quote` would defeat the entire purpose of the `validateRegex` method.

ktomk:
Thanks!

PR #650, opened 2021-01-06
Streamline validation of Regex

a-r-m-i-n:
And always escape used delimiter in passed pattern.

Resolves: #649

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -197,7 +197,7 @@ class FormatConstraint extends Constraint
 
     protected function validateRegex($regex)
     {
-        return false !== @preg_match('/' . $regex . '/u', '');
+        return false !== @preg_match('#' . str_replace('#', '\\#', $regex) . '#u', '');
     }
 
     protected function validateColor($color)


src/JsonSchema/Constraints/FormatConstraint.php
```diff
@@ -197,7 +197,7 @@ protected function validateDateTime($datetime, $format)
 
     protected function validateRegex($regex)
     {
-        return false !== @preg_match('/' . $regex . '/u', '');
+        return false !== @preg_match('#' . str_replace('#', '\\#', $regex) . '#u', '')
```
ohader, line 200: Terminating semi-colon is missing...
Besides that, please use `preg_quote` instead of of `str_replace`here...

ohader (changes requested):

src/JsonSchema/Constraints/FormatConstraint.php
erayd: Do **not** use `preg_quote` here.

The `validateRegex` method is used to *validate* a regular expression by attempting to compile it. Escaping the entire regular expresson entirely defeats the point of that logic, and would result in any string being considered a valid expression.

a-r-m-i-n: But the semicolon has been indeed accidentally removed. I've used github code editor on smartphone ;)

a-r-m-i-n: Fixed the semicolon

ohader: @erayd Thanks for pointing it out

erayd (approved):
Thanks 😁.

guilliamxavier:
I just got hit by that. Could this bugfix be backported to `5.x.x` and released as 5.2.11? 😄 Thanks in advance

guilliamxavier:
Wait, there's another place. I'm preparing a PR

guilliamxavier:
See #653

PR #659, opened 2021-03-19
Fix php warning, docblocks, and peephole optimizations

robocoder:

PR #660, opened 2021-03-19
Add codesniffer; fix phpcs errors wrt PSR12

robocoder:
This PR assumes you bump the PHP requirement to PHP 7.1+.


DannyvdSluijs:
Raising to PHP 7.1 should be okay (as per https://github.com/justinrainbow/json-schema/pull/706#issuecomment-1882962806). However I think it is unexpected to have this "big" change be part of (what looks like) a housekeeping PR.

@robocoder I had tagged you in #659 as well which is now closed by you without a comment. Can we make the assumption you're no longer pursuing any PR's on this project? Meaning #660 and #661 should also be closed?

robocoder:
Yeah, that's to be expected when PRs collect dust for 3 yrs.

PR #662, opened 2021-04-16
Fix CI (php-cs-fixer, HHVM)

guilliamxavier:


guilliamxavier:
Now there's HHVM segfaulting on Travis...

Seldaek:
Thanks

Issue #76, opened 2013-10-10
Not proper "oneOf" handling


antonCPU:
```json
{
    "oneOf": [
        {
            "type": "object",
            "properties": {
                "key": {
                    "type": "string"
                },
                "test1": {
                    "type": "string"
                }
            },
            "required": ["key", "test1"]
        },
        {
            "type": "object",
            "properties": {
                "key": {
                    "type": "string"
                },
                "test2": {
                    "type": "string"
                }
            },
            "required": ["key", "test2"]
        }
    ]
}
```

This schema will pass for:
```json
{
     "test2" : "ok"
}
```
Because of **array_unique()** usage in `JsonSchema\Constraints\Constraint::getErrors()`.
When the second schema will be checked a list of errors will contain "key" and "test1" errors. During checking will be added another error with "key" required, but **array_unique()** will erase it, so this schema will pass a criteria in `JsonSchema\Constraints\Underfined::validateOfProperties()` as a result.

jojo1981:
Tested this issue with version: 2.0.0 and it works like supposed.

    
    [
        {
            "property": "key",
            "message": "The property key is required",
            "constraint": "required"
        },
        {
            "property": "test1",
            "message": "The property test1 is required",
            "constraint": "required"
        },
        {
            "property": "",
            "message": "Failed to match exactly one schema",
            "constraint": "oneOf"
        }
    ]
    
@bighappyface I think this one can also be closed?

bighappyface:
@jojo1981 thanks!

PR #663, opened 2021-04-16
Keep errors order for oneOf like anyOf/allOf

guilliamxavier:


guilliamxavier:
Wait, I just noticed that this is basically reverting #95 😓 Unfortunately @gwagner did not add a test at the time (just changed the order of the existing one), so I have to try guessing now...

guilliamxavier:
Okay I have found out: the original issue was #76, and #95 was actually only a workaround that hid the real cause, which was properly fixed in #98 (with tests!), cf https://github.com/justinrainbow/json-schema/pull/98#issuecomment-41925662

So I say this PR is good 😉

PR #686, opened 2022-04-05
Fix type validation failing for "any" and false-y type wording

Deuchnord:
Fixing a small issue that makes type validation fail when type is `any` or a false-y value, due to `isset()` returning `false` for `null` variables.


Seldaek:
This should be rebased on master actually, and then we do a backport to 5.x.x branch from master. Fix looks good to me though.

Seldaek:
Released as 5.2.12

PR #688, opened 2022-04-19
Fix: Order

localheinz:
This pull request

- [x] keeps rules for `friendsofphp/php-cs-fixer` ordered by name


localheinz:
Thank you, @Seldaek!

PR #690, opened 2022-06-06
Prevents infinite recursion

azambon:
I've come across a case where it was possible to generate a recursive schema. In that instance this method goes into an infinite recursion.
I'm not 100% sure whether this is an issue with json-schema itself or with the fr3d/swagger-assertions layer that I'm using on top of it, but in any case adding an anti-infinite-recursion check here seems reasonable to me.

### DIFF
--- a/src/JsonSchema/SchemaStorage.php
+++ b/src/JsonSchema/SchemaStorage.php
@@ -79,13 +79,21 @@ class SchemaStorage implements SchemaStorageInterface
      *
      * @param mixed  $schema
      * @param string $base
+     * @param array  $stack
      */
-    private function expandRefs(&$schema, $base = null)
+    private function expandRefs(&$schema, $base = null, array $stack = [])
     {
+        if (in_array($schema, $stack)) {
+            //Prevent infinite recursion
+            return;
+        }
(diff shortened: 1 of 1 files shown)


erayd:
Recursive schemas are a legitimate and deliberate part of the spec - validation should follow that recursion however deep is necessary to validate the document.

With that said, recursion should never be *infinite* - it should always be bounded by the depth of the document being validated. This library achieves that by using lazy expansion of references; they shouldn't be actually expanded until the expansion is required during validation. This mechanism inherently prevents infinite recursion.

Are you able to provide a minimal example document and schema that triggers infinite recursion? If so, then that is a bug that needs to be fixed - but should probably be via fixing whatever is wrong with the lazy-load logic, rather than introducing something new.

azambon:
I managed to get back to this, at last. It turned out that the issue was NOT due to the json-schema component, but it appears to be caused by the fr3d/swagger-assertions package that uses json-schema. Hence the confusion.

So we can close this. Sorry for the mistake.

erayd:
No worries - great to hear you got to the bottom of it 😁

PR #693, opened 2022-08-31
Another attempt to make it work under PHP 5.3 and PHP 8+

sanmai:
Build is passing. I left few comments along the way, please ask if anything isn't clear.

Commit history is a mess. If you're not comfortable squash-merging this PR, I'll be happy to merge everything into one commit.


composer.json
```diff
@@ -34,7 +34,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.2.20 || ~2.19.0",
         "json-schema/json-schema-test-suite": "1.2.0",
-        "phpunit/phpunit": "^4.8.35"
+        "sanmai/phpunit-legacy-adapter": "^6.4 || ^8.2.1"
```
sanmai, line 37: https://packagist.org/packages/sanmai/phpunit-legacy-adapter/dependents?order_by=downloads

tests/Constraints/VeryBaseTestCase.php
```diff
@@ -87,4 +87,22 @@ private function getJsonSchemaDraft04()
...
+    }
+
+    public static function assertIsArrayCompat($actual, $message = '')
+    {
+        if (\is_callable(array('PHPUnit\Framework\TestCase', 'assertIsArray'))) {
+            parent::assertIsArray($actual, $message);
+        } else {
+            static::assertInternalType('array', $actual, $message);
+        }
+    }
```
sanmai, line 107: Target PHP versions do not support traits, therefore we can't use [off-the-shelf polyfill](https://github.com/PHPUnitGoodPractices/polyfill).

.github/workflows/continuous-integration.yml
```diff
@@ -36,15 +37,12 @@ jobs:
           - php-version: "5.3"
             dependencies: lowest
             experimental: false
-#          - php-version: "8.0"
-#            dependencies: highest
-#            experimental: false
-#          - php-version: "8.1"
-#            dependencies: lowest-ignore
```
sanmai, line 43: We have to drop this because it leads to installing dependencies broken on PHP 8.1.

composer.json
```diff
@@ -29,12 +29,14 @@
...
         "marc-mabe/php-enum":"^2.0 || ^3.0 || ^4.0",
-        "icecave/parity": "1.0.0"
+        "icecave/parity": ">=1.0.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.2.20 || ~2.19.0",
         "json-schema/json-schema-test-suite": "1.2.0",
-        "phpunit/phpunit": "^4.8.35"
+        "phpunit/phpunit": "^4.8.36 || ^5.7.27 || ^6.5.14 || ^7.5.20 || ^8.5.29 || ^9.5.24",
+        "sanmai/phpunit-legacy-adapter": "^6.4 || ^8.2.1",
```
sanmai, line 38: There's a certain bias since I'm the author of this package. In my defense, it is [used by several high-profile packages with millions of installs](https://packagist.org/packages/sanmai/phpunit-legacy-adapter/dependents?order_by=downloads).

tests/Constraints/VeryBaseTestCase.php
soyuka: what about replacing `assertInternalType` with an `assertTrue(is_array($data), $message)` instead?

composer.json
```diff
@@ -29,12 +29,14 @@
...
+        "phpspec/prophecy-phpunit": "^1.1 || ^2.0.1"
```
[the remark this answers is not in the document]
sanmai: IIRC most recent PHPUnit does not depend on this library, where in fact it is used by this library.

But I digress: it has been two years; my memory is foggy.

tests/Constraints/VeryBaseTestCase.php
sanmai: Following the same logic we should use `assertTrue(is_array())` instead of `assertIsArray()`; what do you think?

sanmai:
I can't remember why I needed this.

Issue #699, opened 2023-05-24
Why do we support higher version json-schema?


hzb1115:
As we all know, Json-schema has been upgraded to draft-06 draft-07.... But why do we only support draft-4?  
And I want to know if we can support a higher version Json-schema.

erayd:
The master branch supports almost all of draft-06.

> But why do we only support draft-4?

Because nobody on this project has the time to implement it. It's basically on life support.

DannyvdSluijs:
@hzb1115 in an effort to clean up this repo and see if we can move forward if the above answer your question, could you close the issue?

erayd:
@DannyvdSluijs Given the time elapsed since I answered this, with no further reply, I'm happy to unilaterally close it.

PR #706, opened 2023-12-14
feat: Add support for Draft 6 in NumberConstraint

llupa:
I am opening this PR because I first encountered it as reported in https://github.com/api-platform/core/issues/6041. I have read and seen either via commit history or issues (such as #699) that this repository has very limited _people_ capabilities.

I tried to keep this as _short_ as possible to offer a Draft 4 and Draft 6 compliance for Numbers->Range case without changing too much. Ref: [Draft 6 Backward-incompatible changes](https://json-schema.org/draft-06/json-schema-release-notes#backwards-incompatible-changes). 

I don't know if an `is_bool` is fine vs `filter_var`, I am more than happy to apply changes as per your feedback.

### DIFF
--- a/src/JsonSchema/Constraints/NumberConstraint.php
+++ b/src/JsonSchema/Constraints/NumberConstraint.php
@@ -26,7 +26,8 @@ class NumberConstraint extends Constraint
     public function check(&$element, $schema = null, JsonPointer $path = null, $i = null)
     {
         // Verify minimum
-        if (isset($schema->exclusiveMinimum)) {
+        if (isset($schema->exclusiveMinimum) && filter_var($schema->exclusiveMinimum, FILTER_VALIDATE_BOOLEAN)) {
+            // Draft 4 schema
             if (isset($schema->minimum)) {
                 if ($schema->exclusiveMinimum && $element <= $schema->minimum) {
                     $this->addError(ConstraintError::EXCLUSIVE_MINIMUM(), $path, array('minimum' => $schema->minimum));
@@ -36,12 +37,18 @@ class NumberConstraint extends Constraint
             } else {
                 $this->addError(ConstraintError::MISSING_MINIMUM(), $path);
             }
+        } elseif (isset($schema->exclusiveMinimum) && filter_var($schema->exclusiveMinimum, FILTER_VALIDATE_INT)) {
+            // Draft 6 schema
+            if ($element <= $schema->exclusiveMinimum) {
+                $this->addError(ConstraintError::EXCLUSIVE_MINIMUM(), $path, array('exclusiveMinimum' => $schema->exclusiveMinimum));
+            }
         } elseif (isset($schema->minimum) && $element < $schema->minimum) {
             $this->addError(ConstraintError::MINIMUM(), $path, array('minimum' => $schema->minimum));
         }
 
         // Verify maximum
-        if (isset($schema->exclusiveMaximum)) {
+        if (isset($schema->exclusiveMaximum) && filter_var($schema->exclusiveMaximum, FILTER_VALIDATE_INT)) {
+            // Draft 4 schema
             if (isset($schema->maximum)) {
                 if ($schema->exclusiveMaximum && $element >= $schema->maximum) {
                     $this->addError(ConstraintError::EXCLUSIVE_MAXIMUM(), $path, array('maximum' => $schema->maximum));
@@ -50,6 +57,11 @@ class NumberConstraint extends Constraint
                 }
             } else {
                 $this->addError(ConstraintError::MISSING_MAXIMUM(), $path);
+            }
+        } elseif (isset($schema->exclusiveMaximum) && filter_var($schema->exclusiveMaximum, FILTER_VALIDATE_INT)) {
+            // Draft 6 schema
+            if ($element >= $schema->exclusiveMaximum) {
+                $this->addError(ConstraintError::EXCLUSIVE_MAXIMUM(), $path, array('exclusiveMaximum' => $schema->exclusiveMaximum));
             }
         } elseif (isset($schema->maximum) && $element > $schema->maximum) {
             $this->addError(ConstraintError::MAXIMUM(), $path, array('maximum' => $schema->maximum));
--- a/tests/Constraints/MinimumMaximumTest.php
+++ b/tests/Constraints/MinimumMaximumTest.php
@@ -28,12 +28,25 @@ class MinimumMaximumTest extends BaseTestCase
                 }'
             ),
             array(
-                '{"value": 3}',
                 '{
-                    "type": "object",
-                    "properties": {
-                        "value": {"type": "integer", "minimum": 3, "exclusiveMinimum": true}
-                    }
+                  "value":16
+                }',
+                '{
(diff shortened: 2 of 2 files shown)


erayd:
Thank you very much for this PR! As you've noticed, we are indeed extremely short on people with available time.

A few points:
1. Could you please set this PR against the master branch, not 5.x.x. The 5.x.x branch is draft-04 only, and draft-06 stuff goes in master.
2. Draft-06 and draft-04 are mutually exclusive regarding the `exclusiveMinimum` and `exclusiveMaximum` keywords unfortunately, as both versions of the spec mandate a type, but the type differs.
3. I appreciate what you are trying to do with the version inference thing (detecting draft-4 vs draft-06 behaviour by looking at the type). It's not really appropriate (because it results in things passing validation which shouldn't), but given that we do this elsewhere already with other keywords, I'm inclined to say that we should just accept it here too. Users that want strict validation of the type against the spec version can enable `CHECK_MODE_VALIDATE_SCHEMA`. With that said, this behaviour does need to be documented in the README, as it's clearly not spec-compliant.
4. I'm unsure whether we should also backport this into 5.x.x, and currently leaning towards not, particularly noting that the master branch is essentially feature-complete for draft-06 - it can be (and is) used in production now for most use-cases. My primary concern is that backporting this could result in validation behaviour changing for existing deployments in violation of the v4 spec, and I really don't want to risk that on something as widely-used as this library is. Happy to be convinced otherwise if you have a good argument for why we should backport it regardless.

@Seldaek Do you have an opinion on (4)? If this violation would trouble composer (or if you aren't sure), then I won't backport it. You're too important to break ;-)

llupa:
Hey! Thank you for looking at this so soon. Since you used numbers, I will use numbers too 😄 

1. I first encountered this issue in API Platform, their schema generator moved on to _greener pastures_ and suddenly my regression tests started breaking. The tests broke in a **dependabot** PR and that had me look deeper. I don't know if you read the linked issue or not, but I can also just override my property to Draft 4 and be done with it. That did not feel like the right solution.
2. The reason I sent it to 5.x.x is because API Platform does not require `master`. To add I have been working with Symfony for 10+ years now so my first instinct was to make it _backward compatible_ and _forward compatible_ but as I am writing this I am noticing that I did not trigger any deprecation for Draft 4 checks. I have no strong opinion where this should go and I will do whatever is your final decision 👌 
3. I can open a different PR for `CHECK_MODE_VALIDATE_SCHEMA` too.

From @erayd comment above it looks like it is safe to require master as a _version_ rather than `^5`, would this be too much for API Platform @soyuka ?

Seldaek:
I'm really not familiar enough with the draft 04 vs 06 changes, but I doubt this impacts Composer. We only have a few integer types defined and none of them have additional restrictions. 

From a "requiring master" perspective, it'd be good if we changed https://github.com/justinrainbow/json-schema/blob/master/composer.json#L41 to be 6.x-dev @erayd - that way projects requiring the draft 06 could require this package as `^6@dev` which is much better than pinning to dev-master, as dev-master will eventually become v7. `^6@dev` will require dependents of dependents to also allow this project in a dev version until there is stable release, so it's also not super convenient, but at least it's semantically correct from a versioning point of view.

llupa:
@erayd I can update this PR against master now. Do you want to keep both Draft 4 and 6 or just move to 6 and never look back?

erayd:
> Do you want to keep both Draft 4 and 6 or just move to 6 and never look back?

Both, please. Many people use v4 still.

Seldaek:
> Requiring PHP 8 as a minimum version for the v6 branch will never happen; it would break too many dependent projects.

@erayd I am sure this is about more than Composer, but from my/Composer perspective, PHP 7.2+ is alright now, as long as v5 is kept maintained for critical fixes (but I really don't expect much of that, and I can do it myself if needed), and most likely the next jump we would do would be PHP 8.1+ to be able to use Symfony components in v6. I don't expect this to happen for another year or so tho as it would leave too many people stranded right now, there's quite a bit of PHP 7.4 usage still :/

soyuka:
My opinion on this is that the Draft 6 version of this package should move on at least php 7 or 8 version, I don't understand why supporting older versions would be a benefit (as it's a maintenance burden anyway). You could always use the `4.x` version of this package with older php versions right? 

To follow @llupa's enhancement here, you should target `master` for this patch, where `master` would be implementing Draft 6. If you want Draft 4, just install 4.x or 5.x version of this package right?

erayd:
> You could always use the 4.x version of this package with older php versions right?

v4 of this library has been unmaintained for ***many*** years. It should be avoided; there are a number of known significant problems with it. For users who require an older version of PHP, and for whom draft-04 is acceptable, then the v5.x.x release branch is usually the right choice.

For users who require draft-06, regardless of PHP version, v6.x.x (master branch) is currently their only option.

llupa:
@erayd I moved the PR to master, but GitHub Actions did not pick up my force push. I wanted to have `Continuous Integration` at least run once. Should I close this PR and open a new one to trigger it? Add a fake commit?

DannyvdSluijs:
@llupa thanks for your efforts and sticking strong with all the rebasing and other comments. In an attempt to revive this library I'm doing some triage on the issues and pull requests. With regards to your commit not being picked up by GitHub workflow I can recommend doing a `git commit --amend --no-edit` which will create a new commit hash which you can then do a `git push --force-with-lease` on. That should do the trick for the workflow.

@erayd I see you mentioning:
>The master branch already implements draft-06, and should be essentially feature-complete. This PR caught something that both we and the unit tests missed, but is an exception rather than the rule.

Does this perhaps mean that the json schema test suite is really do for some updating? Also since the version `1.2.0` which is set in `composer.json` [explains](https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/1.2.0?tab=readme-ov-file#coverage) that is has full coverage on draft3 and draft 4. Version 2.0.0 [mentions](https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/2.0.0?tab=readme-ov-file#coverage) draft 6 and 7 as well.

llupa:
@DannyvdSluijs funny you wrote today. I was thinking about this PR over the weekend 😅 

Workflows waiting to be started 👍

DannyvdSluijs:
@llupa thanks for you initial work. Quite some time has passed since which is on us not on you. 

In #834 the test suite was updated which effectively means we can now run tests for other drafts then just `draft-03` or `draft-04`.

I'll close this PR now as the draft PR #835 is about adding support for `draft-06` to this library.  The proof of concept already support the exclusive minimum and exclusive maximum as you can see [here](https://github.com/jsonrainbow/json-schema/pull/835/files#diff-0f92d4e0aba1938dbbed91c33d94af740a3588c234d4336af3d02524da316f3e) and [here](https://github.com/jsonrainbow/json-schema/pull/835/files#diff-4031c3d1e2296d47e24fc554d0d09ad864cd840a38db7948eb73fde319896002)

Discussions about supporting the newer draft can be found [here](https://github.com/jsonrainbow/json-schema/discussions/820)

At this point I don't have any idea about timeline but I hope to push this along in the coming period.

Again thanks for putting in the effort.

PR #710, opened 2024-02-12
refactor: Remove dead block of code

DannyvdSluijs:
Based on the original PR (#641) from Markus Staab it was found the incorrect order of arguments to be in a dead piece of code. The `strpos`-call was added in ffc856db on 18/02/2016 whereas in 6474265b on 03/03/2016 the calling code of the `validateDateTime` with an `format` argument containing an `u` was removed making this check void. So removing the code block would be best, hence this PR.

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -184,14 +184,6 @@ class FormatConstraint extends Constraint
             return true;
         }
 
-        // handles the case where a non-6 digit microsecond datetime is passed
-        // which will fail the above string comparison because the passed
-        // $datetime may be '2000-05-01T12:12:12.123Z' but format() will return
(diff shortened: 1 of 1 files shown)


dafeder (commented):
@DannyvdSluijs doing my best to review this as I get familiar with the code. It looks to me like the removed case would be covered by the Rfc3339 class, and that all cases are covered in the FormatTest. So I would think if tests pass this is good. However, I don't see a test result for this PR; does CI run for PRs on this repo?

DannyvdSluijs:
>However, I don't see a test result for this PR; does CI run for PRs on this repo?

I’ve seen this on other PRs in this repo as well. I’ll keep an eye out for this issues. For now I’ve amended the commit, creating a new commit hash which has started the pipeline.

dafeder:
Well again a little insecure still about approving being so new to the codebase itself but lgtm?

PR #719, opened 2024-05-13
PHP 8.4 deprecation warnings

sam-bee:
### Introduction

To avoid PHP 8.4 deprecation warnings, introduce explicitly nullable types where null is used as a default value.

### Background

PHP 8.4 will introduce a new deprecation  warning, where `null` is used as a default value, but the typehint is not explicitly nullable. See the [PHP 8.4 approved RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types).

This is causing deprecation warnings in downstream projects - see this [failing build in Composer](https://github.com/composer/composer/actions/runs/9061749203/job/24894182613?pr=11978).

### Solution

Use nullable types in method signatures, wherever `null` is used as a default value.

### Note on PHP versions

The nullable type feature was introduced in PHP 7.1. This pull request proposes a new version of `justinrainbow/json-schema` which explicitly requires PHP 7.1+. Users of PHP 5.3 - 7.0 would not get this version of the project.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
+        "php": ">=7.1.0",
         "marc-mabe/php-enum":"^2.0 || ^3.0 || ^4.0",
         "icecave/parity": "1.0.0"
     },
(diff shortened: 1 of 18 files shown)


sam-bee:
Duplicate of https://github.com/justinrainbow/json-schema/pull/717 ?

Issue #658, opened 2021-02-23
New maintainer required


erayd:
Unfortunately, I no longer have the time available to do this project justice, and have not for some time now. I've been dealing with things as I have the capacity to address them, but the PRs are coming in faster than I have time for, especially recently. I'm also the only active maintainer left, which is resulting in some pretty long delays in resolving things. This isn't good, given our significant downstream dependents.

I think we need to add another maintainer to take over the day-to-day stuff (PR reviews, merges, backports etc). *If you are reading this and are keen to be involved in such a capacity, or would like to suggest a candidate, please comment on this issue*.

---

@bighappyface - do you still have a comms channel with @justinrainbow? We will need him to deal with any new people added to the repo, as neither of us (nor @shmax, who has now moved on entirely) have the necessary admin permissions. If possible, it would be really good to get another person with admin permissions on the project (maybe me for now?) to reduce the bus factor here.

Also important that we be fairly careful about who we bring on board - we need to take into account the security implications if this project were to be compromised.

/cc @Seldaek - would appreciate your input on this; you're the most critical dependent we have.

Seldaek:
Hey, I'd be happy to be added as an admin too here, because indeed Composer depends on it and it's the only dependency we have that is not under our control or part of symfony (which has a fairly decent maintenance outlook).

I however have close to no knowledge of the internals of this lib, and limited json-schema knowledge, so I am not likely to be able to do a lot of maintenance unless absolutely necessary and worth the time to dive in.

erayd:
Thanks @Seldaek - I appreciate that. I think that adding you as an admin would be a good idea if you're OK with it. Even if you won't actually be maintaining it, you're a highly trusted party with an interest in the outcomes, and it would help to reduce the bus factor somewhat.

erayd:
@robocoder How familiar are you with the internal workings of this library? I'm not seeing any previous contribution history from you.

stof:
I don't think you can grant admin permissions to other people on a personal repository. So it might make sense to migrate the repository to an organization if @justinrainbow is not active anymore.

back-2-95:
I was started a new Symfony project and need a JSON validation library. From the memory I came here as my colleagues have used this library for validating JSON schemas.

Now as there is this unfortunate situation, I'd like to just point alternative(s) to this thread so ppl like me can then decide which way to go. My (short) research found 2 alternatives:

- https://github.com/swaggest/php-json-schema
- https://github.com/opis/json-schema 

[...]

PR #723, opened 2024-05-21
Fix: Composer.json homepage

florisluiten:
I was notified by one of my securityscanners that the project URL has been moved (this is indeed correct, as I read in https://github.com/jsonrainbow/json-schema/issues/658). 

I took the liberty to update the homepage in the composer.json to point to the new URL.

### DIFF
--- a/src/JsonSchema/Constraints/FormatConstraint.php
+++ b/src/JsonSchema/Constraints/FormatConstraint.php
@@ -184,14 +184,6 @@ class FormatConstraint extends Constraint
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


DannyvdSluijs (approved):

DannyvdSluijs:
Thanks for the fix. And indeed the repo has moved in the past couple of days so the report was correct. 

Will wait for the pipeline and merge somewhere in the next couple of days.

Seldaek:
The README also needs to be updated so leaving this open as a reminder.

PR #728, opened 2024-05-28
Update README.md to remove more justinrainbow references

e0ipso:
I noticed the reference to `justinrainbow` in the Composer section. I found some more on the links.


DannyvdSluijs:
These were skipped for a reason on the previous PR. On Packagist this library is still called jusyinrainbow/json-schema.  The links would become invalid as well. Is there any specific reason for all off this proposed changes?

DannyvdSluijs:
Closing PR as no feedback was provided after comments. Feel free to reopen if you believe these PR still holds something that should be part of th repo.

PR #682, opened 2022-01-30
Fix ObjectIterator for PHP8

R4c00n:
Hi, with PHP8.1 the iterator methods of ObjectIterator must be covariant with the PHP core interfaces.

I didn't know if theres a specific return type for `::current`, so I went with the suppress annotation here. Happy to change this to a return type, if there is one?

```
Deprecated:  Return type of JsonSchema\Iterator\ObjectIterator::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 42

Deprecated:  Return type of JsonSchema\Iterator\ObjectIterator::next() should either be compatible with Iterator::next(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 52

Deprecated:  Return type of JsonSchema\Iterator\ObjectIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 61

Deprecated:  Return type of JsonSchema\Iterator\ObjectIterator::valid() should either be compatible with Iterator::valid(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 71

Deprecated:  Return type of JsonSchema\Iterator\ObjectIterator::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 81

Deprecated: Return type of JsonSchema\Iterator\ObjectIterator::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [...]/vendor/justinrainbow/json-schema/src/JsonSchema/Iterator/ObjectIterator.php on line 90
```

PR #744, opened 2024-08-02
Extend pipeline for newer versions php

DannyvdSluijs:
This PR updates the pipeline to include PHP >= 8.0 and no longer include PHP <= 7.1. In addition the PR contains the following changes to have green results.
- Update to PHPUnit 8.5 to resolve invocations of deprecated/removed method `each()`, see https://www.php.net/manual/en/function.each.php. Which resulted in:
  - Add `void` return types to `::setup()` methods
  - Explicitly include `phpspec/prophecy`
  - Replace `::setExpectedException()` and `@ExpectedException` for `::expectException`
  - Replace `::assertInternalType()` with `assertIs*()`  methods
  - Replace `::getMock()` with `::createMock()`
- Improve test assertions in case of warning or error during test run.
- Port #682 to fix deprecations from `ObjectIterator`
- Update `icecave/parity` to `^3,0` to resolve invocations of deprecated/removed method `each()`
- Correct styles reported by GHA [workflow](https://github.com/jsonrainbow/json-schema/blob/master/.github/workflows/style-check.yml)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -26,7 +26,10 @@ jobs:
           - "7.2"
           - "7.3"
           - "7.4"
-#          - "8.0"
+          - "8.0"
+          - "8.1"
+          - "8.2"
+          - "8.3"
         dependencies: [highest]
         experimental: [false]
         include:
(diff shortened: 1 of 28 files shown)


DannyvdSluijs:
Has been replaced with #746

PR #746, opened 2024-08-05
Extend pipeline for newer versions php

DannyvdSluijs:
This PR updates the pipeline to include PHP >= 8.0 and no longer include PHP <= 7.1. In addition the PR contains the following changes to have green results.
- Update to PHPUnit 8.5 to resolve invocations of deprecated/removed method `each()`, see https://www.php.net/manual/en/function.each.php. Which resulted in:
  - Add `void` return types to `::setup()` methods
  - Explicitly include `phpspec/prophecy`
  - Replace `::setExpectedException()` and `@ExpectedException` for `::expectException`
  - Replace `::assertInternalType()` with `assertIs*()`  methods
  - Replace `::getMock()` with `::createMock()`
- Improve test assertions in case of warning or error during test run.
- Port #682 to fix deprecations from `ObjectIterator`
- Port #717 to fix implicit nullable deprecations
- Update `icecave/parity` to `^3,0` to resolve invocations of deprecated/removed method `each()`
- Correct styles reported by GHA [workflow](https://github.com/jsonrainbow/json-schema/blob/master/.github/workflows/style-check.yml)

### DIFF
--- a/src/JsonSchema/Uri/UriResolver.php
+++ b/src/JsonSchema/Uri/UriResolver.php
@@ -28,7 +28,7 @@ class UriResolver implements UriResolverInterface
      */
     public function parse($uri)
     {
-        preg_match('|^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?|', $uri ?: '', $match);
+        preg_match('|^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?|', (string) $uri, $match);
 
         $components = array();
         if (5 < count($match)) {
--- a/src/JsonSchema/Uri/UriResolver.php
+++ b/src/JsonSchema/Uri/UriResolver.php
@@ -28,7 +28,7 @@ class UriResolver implements UriResolverInterface
      */
     public function parse($uri)
     {
-        preg_match('|^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?|', $uri, $match);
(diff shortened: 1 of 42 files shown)


src/JsonSchema/Uri/UriResolver.php
```diff
@@ -28,7 +28,7 @@ class UriResolver implements UriResolverInterface
      */
     public function parse($uri)
     {
-        preg_match('|^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?|', $uri, $match);
+        preg_match('|^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?|', $uri ?: '', $match);
```
Seldaek, line 31: This to me looks it would be better to add a string type hint to the uri parameter. But I'm not sure if it would cause further problems in caller code.

Seldaek (approved):
Looking good !

src/JsonSchema/Uri/UriResolver.php
DannyvdSluijs: Thanks this helps. I’ll take a look once I’m back from my holidays. Maybe an explicit string cast could help as well with the benefit of not breaking the function signature (yet).

DannyvdSluijs: This has been done, once the pipeline is green I'll go ahead and merge it.

@Seldaek in the meantime if you can find the time (assuming you're having permissions) could you take a look a the branch protection and change it so that only `master`, `5.x.x` and `6.0.x` have the branch protection rules? I think two inclusion patterns (`master` and `[1-9].*`) would do the trick?
This would help with the rebasing of feature branches which I tend to do to keep a clean history in the feature branch.

Seldaek: Ah sorry about that, fixed it now 👍🏻

PR #761, opened 2024-11-05
Fix PHP 8.4 deprecation warnings in 6.x

W0rma:
From https://github.com/jsonrainbow/json-schema/discussions/726#discussioncomment-9976508:

> I think we could do the same there, tag 6.0 then work on 6.1 bumping PHP and doing all the changes

This PR uses a similar strategy as in https://github.com/jsonrainbow/json-schema/pull/731 but for 6.x:
* bump minimum PHP version to 7.1
* fix PHP deprecation warnings

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -18,6 +18,7 @@ $config
         'no_superfluous_phpdoc_tags' => false,
         'no_useless_else' => true,
         'no_useless_return' => true,
+        'nullable_type_declaration_for_default_null_value' => true,
         'ordered_imports' => true,
         'phpdoc_no_package' => false,
         'phpdoc_order' => true,
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
+        "php": ">=7.1",
         "marc-mabe/php-enum":"^2.0 || ^3.0 || ^4.0",
         "icecave/parity": "1.0.0"
     },
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -39,12 +39,12 @@ class BaseConstraint
     /**
      * @param Factory $factory
      */
-    public function __construct(Factory $factory = null)
+    public function __construct(?Factory $factory = null)
     {
         $this->factory = $factory ?: new Factory();
     }
 
-    public function addError(ConstraintError $constraint, JsonPointer $path = null, array $more = array())
+    public function addError(ConstraintError $constraint, ?JsonPointer $path = null, array $more = array())
     {
         $message = $constraint ? $constraint->getMessage() : '';
         $name = $constraint ? $constraint->getValue() : '';
--- a/src/JsonSchema/Constraints/CollectionConstraint.php
+++ b/src/JsonSchema/Constraints/CollectionConstraint.php
@@ -23,7 +23,7 @@ class CollectionConstraint extends Constraint
     /**
      * {@inheritdoc}
      */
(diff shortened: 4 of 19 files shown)


DannyvdSluijs:
Hi @W0rma 

Thanks for the contribution. I was curious why you're targeting the `6.0.x` branch? As that branch is for patch release for version`6.x` and we should not drop support of dependency versions in a patch release.

The changes (besides the one in `.php-cs-fixer.dist.php`) are already part of the `master` branch. So all it needs is work towards a release of a new mayor version. The discussion you linked has a todo list on what (ideally) needs to be done.

Also if you think the changes in `.php-cs-fixer.dist.php` are needed can you create a PR adding those changes to the `master` branch?

W0rma:
> So all it needs is work towards a release of a new mayor version.

@DannyvdSluijs Will the next release be a major (`7.0`) or a minor (`6.1`) version?

I wasn't sure if `master` already contains BC breaks and my aim was to fix the deprecation warnings for the next `6.x` minor version.

W0rma:
I'm happy to assist as far as I can. Could you create a `6.x` branch so I can prepare a PR which ports the fixes for the PHP 8.4 deprecation warnings?

DannyvdSluijs:
Ideally we get @Seldaek to rename the current `6.0.x` branch to `6.x` (I don't have permissions to do so) so we have a proper branch for version 6. And all the work done in master is going to be 7.0 (which I think releasing sooner than later would be the best option).

This would also help #776 move forward for other people.

erayd:
@DannyvdSluijs For what it's worth, in my opinion you should be set as a project owner. You're the main person driving this library forward, and frankly the only trusted party who has time. And you're clearly in it for the long haul; you've been working on this for quite some time now. It's been wonderful to see some progress 😁 

@Seldaek What do you think?

Seldaek:
Yeah that makes total sense. I've granted you owner access on the organization @DannyvdSluijs :)

DannyvdSluijs:
@W0rma if you could revert the changes is `composer.json` we could merge this one. After which I'll rename the targetted branch (doing so before would close this PR) and make a new 6.x release

W0rma:
> if you could revert the changes is composer.json we could merge this one.

@DannyvdSluijs Sure, I can do that.

Just to make sure that there is no misunderstanding: The version bump is necessary because nullable types require at least PHP 7.1 (see https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types).

Do you plan to add the change again after this PR is merged?

DannyvdSluijs:
@W0rma I didn't think about that implication, thanks for pointing it out. Raising the PHP level is only something we want to do in the next mayor version (already done in the `master`-branch). Which makes the need for having a 7.x version even bigger to support PHP 8.4. for people that are using this library on **PHP 8.4**

As I would see it now we would have:
- `5.x` only fixes (bugs & security).
- `6.x` is there for >= PHP 5.3 and <= PHP 8.3.
- `7.x` is for PHP >= 7.1

This means we can't add the nullable types natively. We could fallback to the older PHPDoc versions and leave the native type unspecified. This way we could still support PHP 8.4 for version `6.x`

Seldaek:
IMO this should be merged as 6.1 (like 5.3.0 dropped php <7.1) or 7.0 if you rather, so that leaves 6.0.x for php 5.3-7.1 users.. Using this these days really is not a reasonable thing anymore. It's not worth spending extra effort supporting these versions.

Up to you of course if you rather do it feel free.. but doesn't seem worth it to me.

W0rma:
> But your contribution leads to the release of the next version so you're making an impact

No worries, that's what I wanted to achieve :smiley:

DannyvdSluijs:
Tag `6.1.0` has been [released](https://github.com/jsonrainbow/json-schema/releases/tag/6.1.0)

PR #769, opened 2024-11-26
fix: correctly set the schema ID when passing it as assoc array

### LINKED ISSUE #767: Validator::validate() doesn't correctly apply schema ID when the schema is an assoc array (works when an object)
Update: this is due to the validator not applying the schema ID correctly, it only handles an object, not an assoc array.

I have this schema:

```json
{
    "id": "tests/schemas/suggest/empty.schema.json",
    "allOf": [
        {
            "$ref": "baseline.schema.json"
        }
    ],
    "properties": {
        "items": {
            "maxItems": 0
        }
    }
}
```

With the update to 6.x, this fails with

```
      file_get_contents(internal://provided-schema/tests/schemas/suggest/baseline.schema.json): Failed to open stream: No such file or directory (JsonSchema\Exception\ResourceNotFoundException)
```

I guess the schemaless ref is now interpreted as internal in some way?

If I modify my schema as so

```json
{
    "id": "tests/schemas/suggest/empty.schema.json",
    "allOf": [
        {
            "$ref": "file://tests/schemas/suggest/baseline.schema.json"
        }
    ],
    "properties": {
        "items": {
            "maxItems": 0
        }
    }
}
```

Now it says

```
      file_get_contents(file://tests/schemas/suggest/baseline.schema.json): Failed to open stream: no suitable wrapper could be found (JsonSchema\Exception\ResourceNotFoundException)
```

which means it understands it needs to read the file locally. For some reason, it doesn't read the file even though it exists.

If I add 

```php
$uri = substr($uri, 7);
```

to strip `file://` before here, it works. 

https://github.com/jsonrainbow/json-schema/blob/377231495a8f76945645ee0ebfb5dbf9fd438a02/src/JsonSchema/Uri/Retrievers/FileGetContents.php#L36

Should this happen conditionally, if there's a `file://` prefix, strip it? There's some `file://` special handling a little later in the function, but it's already too late because an error was raised.

dkarlovi:
Closes #767.

### DIFF
--- a/src/JsonSchema/Validator.php
+++ b/src/JsonSchema/Validator.php
@@ -60,6 +60,8 @@ class Validator extends BaseConstraint
         // add provided schema to SchemaStorage with internal URI to allow internal $ref resolution
         if (is_object($schema) && property_exists($schema, 'id')) {
             $schemaURI = $schema->id;
+        } elseif (is_array($schema) && array_key_exists('id', $schema)) {
+            $schemaURI = $schema['id'];
         } else {
             $schemaURI = SchemaStorage::INTERNAL_PROVIDED_SCHEMA_URI;
         }
--- a/tests/ValidatorTest.php
+++ b/tests/ValidatorTest.php
@@ -18,6 +18,17 @@ class ValidatorTest extends TestCase
         $this->assertFalse($validator->isValid(), 'Validation succeeded, but should have failed.');
     }
 
+    public function testValidateWithAssocSchemaWithRelativeRefs(): void
+    {
+        $schema = json_decode(file_get_contents(__DIR__.'/fixtures/relative.json'), true);
+        $data = json_decode('{"foo":{"foo": "bar"}}', false);
+
+        $validator = new Validator();
+        $validator->validate($data, $schema);
+        
+        $this->assertTrue($validator->isValid(), 'Validation failed, but should have succeeded.');
+    }
+
     public function testBadAssocSchemaInput(): void
     {
         if (version_compare(phpversion(), '5.5.0', '<')) {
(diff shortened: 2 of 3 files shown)


DannyvdSluijs:
@dkarlovi I think this PR could be accepted. If you could just do a rebase and add a changelog entry in `CHANGELOG.md`?

Maybe good to know we are seeing multiple issues being reported with associative arrays, which is a concept that doesn't live in JSON. With those two facts we're seeing if we should [deprecate the associative array support](https://github.com/jsonrainbow/json-schema/discussions/788#drop-support-for-associative-arrays) in the next mayor version `7.0.0`

src/JsonSchema/Validator.php
```diff
@@ -60,6 +60,8 @@ public function validate(&$value, $schema = null, $checkMode = null)
         // add provided schema to SchemaStorage with internal URI to allow internal $ref resolution
         if (is_object($schema) && property_exists($schema, 'id')) {
             $schemaURI = $schema->id;
+        } elseif (is_array($schema) && array_key_exists('id', $schema)) {
+            $schemaURI = $schema['id'];
         } else {
             $schemaURI = SchemaStorage::INTERNAL_PROVIDED_SCHEMA_URI;
         }
```
DannyvdSluijs, line 67: I believe I've learned something more about the codebase today. These changes should be done respecting the `CHECK_MODE_TYPE_CAST`. Doing either a loose type check or a strict type check. I believe the following would do so (but requires some testing as well)

```suggestion
        $schemaURI = SchemaStorage::INTERNAL_PROVIDED_SCHEMA_URI;
        if ($this->factory->getTypeCheck()::propertyExists($schema, 'id')) {
            $schemaURI = $this->factory->getTypeCheck()::propertyGet($schema, 'id');
        }
```

DannyvdSluijs:
This PR has been replaced by #794

PR #772, opened 2025-01-10
Remove obsolete check if FILTER_FLAG_EMAIL_UNICODE is defined

W0rma:
`FILTER_FLAG_EMAIL_UNICODE` is available in PHP 7.1 or higher https://www.php.net/manual/en/filter.constants.php#constant.filter-flag-email-unicode and this package does no longer support PHP 7.1 or lower.

Therefore the check can be safely removed.


DannyvdSluijs (commented):
Nice cleanup @W0rma!

Two things that is needed before this can be merged. 
1. The pipeline is providing some feedback about an ignored error in PHPStan. Could you remove it from the`phpstan-baseline.neon` which can be done by running `./vendor/bin/phpstan --generate-baseline`
2. Could you add an entry i the `CHANGELOG.md` (link)[https://github.com/jsonrainbow/json-schema/blob/master/CHANGELOG.md] under changed?

W0rma:
@DannyvdSluijs Thank you for the review - Done!

DannyvdSluijs (approved):

PR #779, opened 2025-02-04
Use phpstan's int-mask-of<T>

jack-worman:
We are getting phpstan errors when using multiple options of the checkmode (i.e. `Constraint::CHECK_MODE_VALIDATE_SCHEMA | Constraint::CHECK_MODE_EXCEPTIONS`)


DannyvdSluijs:
@jack-worman great improvement, could you also add an entry to the changelog under `## [Unreleased]` following the format from the other lines.
And the PHPStan workflow seems to highlight some issues that might need some looking into.

jack-worman:
CI should be fixed now.

DannyvdSluijs (approved):

PR #781, opened 2025-02-06
More-phpstan-typing

jack-worman:
Just some more low hanging phpstan error fixes

### DIFF
--- a/src/JsonSchema/Constraints/BaseConstraint.php
+++ b/src/JsonSchema/Constraints/BaseConstraint.php
@@ -39,18 +39,15 @@ class BaseConstraint
      */
     protected $factory;
 
-    /**
-     * @param Factory $factory
-     */
     public function __construct(?Factory $factory = null)
     {
         $this->factory = $factory ?: new Factory();
     }
 
-    public function addError(ConstraintError $constraint, ?JsonPointer $path = null, array $more = [])
+    public function addError(ConstraintError $constraint, ?JsonPointer $path = null, array $more = []): void
     {
-        $message = $constraint ? $constraint->getMessage() : '';
-        $name = $constraint ? $constraint->getValue() : '';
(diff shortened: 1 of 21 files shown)


src/JsonSchema/Constraints/BaseConstraint.php
```diff
@@ -89,6 +92,12 @@ public function addErrors(array $errors)
         }
     }
 
+    /**
+     * @param int $errorContext
```
DannyvdSluijs, line 96: I would prefer the native type hint over a docblock

DannyvdSluijs (commented):
Thanks on these type of contributions. I'm unclear if there is a specific reason for adding the types as DocBlock version instead of native ones.  The `master` branch supports PHP 7.2 and up, so we can use most of the native typehints. We only can't use typed properties, mixed return type, union types or Enums.

Would you be okay with improving the PR to use more of the native type hints instead of adding their respective. DocBlock versions. Obvious the `@phpstan-param` can remain to reduce the type further.

jack-worman:
Adding native typehints would be a BC break for people extending the class. Is this okay?

DannyvdSluijs:
> Adding native typehints would be a BC break for people extending the class. Is this okay?

I don't think this is a big problem. We are just strengthening the code base as we are making more explicit what the return types are.

src/JsonSchema/Constraints/BaseConstraint.php
jack-worman: added native typehints for some the stuff i already touched (and a bit more)

DannyvdSluijs (approved):

DannyvdSluijs:
Thanks for all the additions. This is a great step forward into type completion for JSON Schema and bringing it back into modern PHP standards.

Issue #806, opened 2025-03-17
DeepComparer returns false when compares null with null


forrest79:
Hi,

I found out, that this code returns `false`: `var_dump(DeepComparer::isEqual(null, null));`.

I used temporary fix in our repo:

```php
        $isLeftScalar = is_scalar($left) || is_null($left);
        $isRightScalar = is_scalar($right) || is_null($right);
```

DannyvdSluijs:
This was fixed in #807 and version [6.3.1](https://github.com/jsonrainbow/json-schema/releases/tag/6.3.1) was just released containing this fix.

PR #807, opened 2025-03-18
Fix null comparison

nicosp:
Fixes null not equal to null.

Issue: https://github.com/jsonrainbow/json-schema/issues/806


DannyvdSluijs:
@nicosp great work 🥳. Pipelines are green and changes look all good.

One tiny thing, could you add an entry to the change log? If not I'll add it tonight and create a new release.

DannyvdSluijs (approved):
Love it

Issue #816, opened 2025-04-04
Version 6 is broken when using a 32bits build of PHP


stof:
see https://github.com/symfony/symfony/actions/runs/14262540647/job/39977141591#step:5:23 which was a CI job of Symfony running on 32bits with Composer 2.8.7 (which migrated to v6 of this library).

the root cause is that the value of `Validator::ERROR_ALL` is bigger than `PHP_INT_MAX` on a 32bits build of PHP (as PHP integers are signed integers) and so becomes a float.

stof:
Note that `shivammatur/setup-php` supports a way of using a 32bits build of PHP in CI on Ubuntu. See https://github.com/shivammathur/setup-php#multi-arch-setup and https://github.com/shivammathur/setup-php/issues/302#issuecomment-705528321

It could make sense to add a job running on 32bits (as this package is used as part of Composer, it cannot really drop support for such builds)

PR #817, opened 2025-04-04
Fix support for 32bits PHP

nicolas-grekas:
Fix https://github.com/jsonrainbow/json-schema/issues/816

### DIFF
--- a/src/JsonSchema/Validator.php
+++ b/src/JsonSchema/Validator.php
@@ -27,10 +27,10 @@ class Validator extends BaseConstraint
 {
     public const SCHEMA_MEDIA_TYPE = 'application/schema+json';
 
-    public const ERROR_NONE                    = 0x00000000;
-    public const ERROR_ALL                     = 0xFFFFFFFF;
-    public const ERROR_DOCUMENT_VALIDATION     = 0x00000001;
-    public const ERROR_SCHEMA_VALIDATION       = 0x00000002;
+    public const ERROR_NONE                    = 0;
+    public const ERROR_ALL                     = -1;
+    public const ERROR_DOCUMENT_VALIDATION     = 1;
+    public const ERROR_SCHEMA_VALIDATION       = 2;
 
     /**
      * Validates the given data against the schema and returns an object containing the results
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fix support for 32bits PHP ([#817](https://github.com/jsonrainbow/json-schema/pull/817))
 
 ## [6.4.0] - 2025-04-01
 ### Added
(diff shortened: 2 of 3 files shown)


DannyvdSluijs:
@nicolas-grekas I won't expect these changes to work as the flags are part of a bitmask. Wouldn't it suffice to lower the `ERROR_ALL` to `0x7FFFFFFF` (32 bits signed int max in hex)

nicolas-grekas:
I promise this works. The binary representation is all ones.

src/JsonSchema/Validator.php
```diff
@@ -27,10 +27,10 @@ class Validator extends BaseConstraint
 {
     public const SCHEMA_MEDIA_TYPE = 'application/schema+json';
 
-    public const ERROR_NONE                    = 0x00000000;
-    public const ERROR_ALL                     = 0xFFFFFFFF;
-    public const ERROR_DOCUMENT_VALIDATION     = 0x00000001;
-    public const ERROR_SCHEMA_VALIDATION       = 0x00000002;
+    public const ERROR_NONE                    = 0;
+    public const ERROR_ALL                     = -1;
```
Seldaek, line 31: `self::ERROR_DOCUMENT_VALIDATION | self::ERROR_SCHEMA_VALIDATION` also should work here no? Or just hardcode `3` because frankly this is very unlikely to change, so if a new type is added with 4 then you just gotta update the ERROR_ALL.. Doesn't sound that hard 🤷🏻 


```suggestion
    public const ERROR_ALL                     = self::ERROR_DOCUMENT_VALIDATION | self::ERROR_SCHEMA_VALIDATION;
```

Seldaek:
@DannyvdSluijs I know you might be busy with other stuff and that's fine if that's the case, but if you are able to get this into a release in the next couple hours then I could also do a Composer one :)

src/JsonSchema/Validator.php
nicolas-grekas: It makes more sense to me to use -1 for an E_ALL level. Like `error_reporting(-1)` is a common way to turn on all error levels for the engine...

Seldaek: Yup, works for me with -1 too.. Not sure why PHPStan complains though, but I tried on my machine and wasn't able to quiet it down. Seems like the 0xFFFFFFFF makes it give up on some checks.

DannyvdSluijs:
I'll go ahead and trust @nicolas-grekas and @Seldaek on this one (both being high valued members in the PHP community).
Pipeline is green, will merge and do a patch release in the next 15 minutes.

DannyvdSluijs:
[Version 6.4.1](https://github.com/jsonrainbow/json-schema/releases/tag/6.4.1) has been released

nicolas-grekas:
Thank you!

Seldaek:
Thanks! And just to prove this so https://3v4l.org/VbvJ3 ;)

src/JsonSchema/Validator.php
staabm: Phpstan complains because it doesn’t support int-ranges from binary & and | operator yet.

Seldaek: Ok thanks, i thought i saw this work but must remember it wrong then :)

PR #823, opened 2025-05-02
Update UPGRADE-6.0.md

### LINKED ISSUE #822: Breaking change missing in 5 to 6 upgrade
Seems the signature of `BaseConstraint` has changed:

```
public function addError(?JsonPointer $path, $message, $constraint = '', ?array $more = null)

public function addError(ConstraintError $constraint, ?JsonPointer $path = null, array $more = []): void
```

Is that intentional? Guess the fact it isn't optional anymore is the problem there.

bbrala:
Add missing breaking change to upgrade guide

Fixes #822


DannyvdSluijs (approved):

bbrala:
Added changelog entry

PR #830, opened 2025-05-28
feat: add return types to loose type check constraint

shakaran:
This add typing following this rules:

ReturnTypeFromStrictFluentReturn
ReturnTypeFromStrictNativeCall
ReturnTypeFromStrictNewArray

### DIFF
--- a/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
+++ b/src/JsonSchema/Constraints/TypeCheck/LooseTypeCheck.php
@@ -6,14 +6,14 @@ namespace JsonSchema\Constraints\TypeCheck;
 
 class LooseTypeCheck implements TypeCheckInterface
 {
-    public static function isObject($value)
+    public static function isObject($value): bool
     {
         return
             is_object($value) ||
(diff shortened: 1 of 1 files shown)


shakaran:
@Seldaek I agree with you — this could indeed be considered a potential BC break. That said, how can we move forward with adding proper typing to Composer and related projects, and eventually support PHP 8.4?

Are we destined to keep dragging PHP 7.2 (and even EOL versions) along for several more years?

Is there no way we could start tagging new versions that clearly indicate such BC changes?

DannyvdSluijs:
> Adding return types on public methods is a possible BC break for anyone overriding that method.

@Seldaek I'm kind of on the fence with this. I completely get your point and in a library which offers interfaces and has the intention for other developers to implement their own adding such a thing breaks stuff. In this library I would like to be more flexible where possible. 
This doesn't change that we should remain compatible with PPH 7.2 in order to support Composer in it's needs. @shakaran this is one of the main reasons we are "stuck" (for good reasons) on supporting older PHP versions. I hope this helps and explains some. I'll keep this PR on my todo list and see if this can be reused/applied at a later time.

PR #835, opened 2025-06-25
feat: add proof-of-concept implementation for strict fully validating using draft-06 schema

DannyvdSluijs:
⚠️ This is a draft PR as some blockers have been resolved breaking other things in order to get a proof of concept of the approach for supporting multiple drafts.

This PR will:
- Introduce the `JsonSchema\Constraints\Drafts\Draft06` namespace
- A constraint per keyword
- Enable the `draft6` in the test, locally I have 633 passing test, 312 failing tests and 157 ignored tests.

PR #838, opened 2025-08-13
Allow for time-numoffset with and without colon separator

### LINKED ISSUE #837: Possible invalid datetime in FormatTest::getValidFormats
Hi, I see that your working towards supporting draft-06, so I thought I'd try to help.

The first test failure I had was:

```
1) JsonSchema\Tests\Constraints\FormatTest::testValidFormat with data set #10 ('2000-05-01T12:12:12+0100', 'date-time')
Failed asserting that an array is empty.
```

I tracked this down to:

```
        yield ['2000-05-01T12:12:12+0100', 'date-time'];

```

in the `getValidFormats` method, which appears to be an invalid format for RFC3339.

I've opened a pull request to remove that line, please do close this issue (and I'll close the pull request), if I am wrong.

Best!

matason:
closes #837


DannyvdSluijs:
Thanks for the fix. I've cherry picked fc994a73ffa63e5653d608e15632f8f786bb9d84 into #835 

You're correct that offset without a colon is invalid. However this has become part of the expected behaviour of existing implementations so I would rather not break those. I've added it to the list of deprecations in the next mayor version.

With the commit cherry picked this PR can be closed. Once again thanks for the work done 👍🏼

PR #840, opened 2025-08-27
Fix PHP 8.5 deprecation

andypost:
https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_the_http_response_header_predefined_variable

related to #798

PR #841, opened 2025-08-29
Fix PHP 8.5 $http_response_header deprecation

DannyvdSluijs:
Replaces #840 (due to blocked write access on fork)


DannyvdSluijs:
> Any chance this deprecation could be backported to 5.x ? I'm using `behatch/contexts` (which is sadly abandoned) and it requires ^5.0. :(

Hi @gnutix 

There is a 5.x.x branch for which you can create a port based on this PR. I can do the merging and releasing.

PR #853, opened 2025-11-28
Allow underscore and tilde in URI hostnames as per RFC 3986

longwave:
The implementation of [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) in the URI validator is more strict than the specification allows. While the RFC says

> URI producers should use names that conform to the DNS syntax, even when use of DNS is not immediately apparent

the host component grammar actually allows more characters than are allowed in DNS names.

This is a problem for the Drupal project which uses this package in conjunction with custom stream wrappers with module names in the host component. Module names conform to PHP function name standards, not DNS standards - that is, they allow underscores but not dashes.

The relevant RFC grammar is
```
      host        = IP-literal / IPv4address / reg-name

      reg-name    = *( unreserved / pct-encoded / sub-delims )

      unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

This PR widens the `host` validation to allow all characters in the `unreserved` set. While `pct-encoded` and `sub-delims` are also technically allowed, this is the smallest change that will help us out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Fixed
-- Allow more characters in URI hostnames as per RFC 3986 ([#853](https://github.com/jsonrainbow/json-schema/pull/853))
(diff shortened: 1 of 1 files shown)


DannyvdSluijs:
Thank you both. @longwave you’ve played a great part in this as well. Creating a clear and minimal PR, resolving the the feedback points very quickly have helped a lot.

PR #864, opened 2025-12-27
Backport of PHP 8.5 patch to 5.x.x

### LINKED ISSUE #798: Add PHP 8.5 to the build matrix
Add PHP 8.5 to the build matrix. Since this version is not released it should allow failure. I've used this in the JsonMapper project as well.

https://github.com/JsonMapper/JsonMapper/blob/develop/.github/workflows/build.yaml#L84
https://github.com/JsonMapper/JsonMapper/blob/develop/.github/workflows/build.yaml#L111

### LINKED ISSUE #863: curl_close in 5.x.x
https://github.com/jsonrainbow/json-schema/commit/93e0ed79307db19bcfea23c272c559e2b6673b5b wrapped the `curl_close` call in a version check in main.

5.x.x still has this unguarded at https://github.com/jsonrainbow/json-schema/blob/5.x.x/src/JsonSchema/Uri/Retrievers/Curl.php#L54

As other PHP 8.5 changes are being made/backported as per https://github.com/jsonrainbow/json-schema/releases/tag/5.3.1 can we get this one done too?

I don't know if the whole linked commit above should be backported, or if we just want to make a partial backport (happy to do this).

reedy:
ci: Add PHP 8.5 to pipeline, ignoring dependencies and as experimental (#842)

This PR will
- resolve deprecations for:
  - Reflection*::setAccessible
  - curl_close

Fixes #798
Fixes #863


DannyvdSluijs:
@reedy thanks for the PR and I’m willingly get this merged into 5.x.x. This might take some while as I’m enjoying the holidays atm. 

Looks like the CI pipeline indeed doesn’t work so I’ll have to jump on that before we merge things. Do you already found something or have a idea on why it isn’t running them?

reedy:
I'm pretty sure it's because there's no `.github/workflow` files in the branch, and I'm not trying to merge into `main`, so it doesn't know to fallback to anywhere. I haven't seen any docs suggest that there's any sort of fallback behaviour (at least by default, you may be able to define it somehow)

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpull_requestpull_request_targetbranchesbranches-ignore

>When using the pull_request and pull_request_target events, you can configure a workflow to run only for pull requests that target specific branches.

The easiest answer is probably copying a version of the relevant CI files you want to run back into that branch, and it'll probably work fine... I guess that should be easily proved, will do a patch.

No rush on my part, I'm just a little bored, and upgrading an ancient version of composer (while running on PHP 8.5) locally noticed another PHP 8.5 cleanup that needed doing in MediaWiki and various vendored libs, such as this one :)

DannyvdSluijs:
Thanks for all the hard work @reedy. I've taken the liberty, since there was no feedback, to re-author the PR. It has been replaced by https://github.com/jsonrainbow/json-schema/pull/900

PR #912, opened 2026-05-27
fix: annotate resolved schemas with dialect-aware id keyword

### LINKED ISSUE #911: UriRetriever annotates resolved $refs with draft-04 "id" even when configured for Draft-6+
> Couldn't find an existing issue for this — closest is discussion #820 (general "supporting newer drafts"), please point me at a duplicate if I missed one.

## Summary

`UriRetriever::retrieve()` unconditionally stamps every resolved schema with `$jsonSchema->id = $resolvedUri;`. The library's default dialect was bumped to Draft-6 in `Factory::$defaultDialect` and Draft-6 [renamed `id` to `$id`](https://json-schema.org/draft-06/json-schema-release-notes#backwards-incompatible-changes), but `UriRetriever` was never updated to match. As a result, a `Validator` configured for Draft-6/7 still emits a Draft-4 `id` keyword on every URI-resolved schema.

Internally this is harmless — `SchemaStorage::findSchemaIdInObject()` already reads either `id` or `$id` ([SchemaStorage.php:236-243](https://github.com/jsonrainbow/json-schema/blob/master/src/JsonSchema/SchemaStorage.php#L236-L243)). But anyone consuming the resolved schema externally — serializing it to JSON, sending it to a strict Draft-7 validator like Ajv, etc. — gets a schema with an unknown keyword:

```
Error: NOT SUPPORTED: keyword "id", use "$id" for schema ID
```

## Minimal reproduction

```php
use JsonSchema\Constraints\Factory;
use JsonSchema\DraftIdentifiers;
use JsonSchema\SchemaStorage;

$factory = new Factory();
$factory->setDefaultDialect(DraftIdentifiers::DRAFT_7);
/** @var SchemaStorage $storage */
$storage = $factory->getSchemaStorage();
$storage->addSchema('file:///tmp/example.json', (object) [
    '$defs' => (object) [
        'thing' => (object) ['type' => 'object'],
    ],
]);
$resolved = $storage->resolveRefSchema((object) ['$ref' => 'file:///tmp/example.json#/$defs/thing']);
var_dump(property_exists($resolved, 'id'));   // bool(true)  ← draft-04 keyword leaked
var_dump(property_exists($resolved, '$id'));  // bool(false)
```

## Offending code

[`src/JsonSchema/Uri/UriRetriever.php` (line 188-190)](https://github.com/jsonrainbow/json-schema/blob/master/src/JsonSchema/Uri/UriRetriever.php#L188-L190):

```php
if ($jsonSchema instanceof \stdClass) {
    $jsonSchema->id = $resolvedUri;
}
```

No dialect check.

## Proposed fix

Make the annotation dialect-aware. Minimal, BC-preserving — Draft-3/4 callers keep getting `id`; Draft-6+ callers get `$id`.

penyaskito:
## Summary

- `UriRetriever::retrieve()` now reads the root schema's `$schema` keyword to determine whether to annotate with `id` (Draft-3/4) or `$id` (Draft-6+)
- Schemas without a `$schema` keyword default to `$id`, matching the library's Draft-6 default dialect
- Updated `testRetrieveSchemaFromPackage` to assert structure instead of a brittle md5 hash

## Context

`UriRetriever::retrieve()` unconditionally stamped `$jsonSchema->id = $resolvedUri` on every resolved schema. Draft-6 [renamed `id` to `$id`](https://json-schema.org/draft-06/json-schema-release-notes#backwards-incompatible-changes), but the annotation was never updated. This leaks a Draft-4 keyword into schemas resolved under Draft-6/7, causing errors in strict validators (e.g. Ajv: `NOT SUPPORTED: keyword "id", use "$id" for schema ID`).

The dialect is detected from the root schema *before* pointer resolution, so sub-schemas inherit the correct keyword from their parent document.

Internally safe — `SchemaStorage::findSchemaIdInObject()` already reads both `id` and `$id`.

## Test plan

- [x] Full test suite passes (6928 tests, 0 failures)
- [x] Verified Draft-04 meta-schema → `id`, Draft-07 → `$id`, no `$schema` → `$id`

Fixes #911

### DIFF
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -185,8 +185,8 @@ class UriRetriever implements BaseUriRetrieverInterface
 
         // Detect dialect from the root schema's $schema keyword before
         // resolvePointer() may walk into a sub-schema.
-        $dialect = isset($jsonSchema->{'$schema'}) ? $jsonSchema->{'$schema'} : null;
-        $usesDollarId = !in_array($dialect, [DraftIdentifiers::DRAFT_3, DraftIdentifiers::DRAFT_4], true);
+        $dialect = isset($jsonSchema->{'$schema'}) ? rtrim($jsonSchema->{'$schema'}, '#') : null;
+        $usesDollarId = !in_array($dialect, [DraftIdentifiers::DRAFT_3()->withoutFragment(), DraftIdentifiers::DRAFT_4()->withoutFragment()], true);
 
         // Use the JSON pointer if specified
         $jsonSchema = $this->resolvePointer($jsonSchema, $resolvedUri);
--- a/src/JsonSchema/Uri/UriRetriever.php
+++ b/src/JsonSchema/Uri/UriRetriever.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 
 namespace JsonSchema\Uri;
 
+use JsonSchema\DraftIdentifiers;
 use JsonSchema\Exception\InvalidSchemaMediaTypeException;
 use JsonSchema\Exception\JsonDecodingException;
 use JsonSchema\Exception\ResourceNotFoundException;
@@ -182,11 +183,17 @@ class UriRetriever implements BaseUriRetrieverInterface
 
         $jsonSchema = $this->loadSchema($fetchUri);
 
+        // Detect dialect from the root schema's $schema keyword before
+        // resolvePointer() may walk into a sub-schema.
+        $dialect = isset($jsonSchema->{'$schema'}) ? $jsonSchema->{'$schema'} : null;
+        $usesDollarId = !in_array($dialect, [DraftIdentifiers::DRAFT_3, DraftIdentifiers::DRAFT_4], true);
+
         // Use the JSON pointer if specified
         $jsonSchema = $this->resolvePointer($jsonSchema, $resolvedUri);
 
         if ($jsonSchema instanceof \stdClass) {
-            $jsonSchema->id = $resolvedUri;
+            $idKeyword = $usesDollarId ? '$id' : 'id';
+            $jsonSchema->{$idKeyword} = $resolvedUri;
         }
 
         return $jsonSchema;
--- a/tests/Uri/UriRetrieverTest.php
+++ b/tests/Uri/UriRetrieverTest.php
@@ -341,6 +341,7 @@ EOF;
         // check that the schema was loaded & processed correctly
         $this->assertEquals('object', $schema->type);
         $this->assertObjectHasAttribute('$id', $schema);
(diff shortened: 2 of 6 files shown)


tests/Uri/UriRetrieverTest.php
```diff
@@ -339,7 +339,8 @@ public function testRetrieveSchemaFromPackage(): void
         $this->assertNotFalse($schema);
 
         // check that the schema was loaded & processed correctly
-        $this->assertEquals('454f423bd7edddf0bc77af4130ed9161', md5(json_encode($schema)));
+        $this->assertEquals('object', $schema->type);
+        $this->assertObjectHasAttribute('$id', $schema);
```
penyaskito, line 343: ⚠️ I'm not sure why an md5 was used here. Maybe it should just update this to the new md5 now that it uses `$id` instead of `id`?

src/JsonSchema/Uri/UriRetriever.php
```diff
@@ -182,11 +183,17 @@ public function retrieve($uri, $baseUri = null, $translate = true)
 
         $jsonSchema = $this->loadSchema($fetchUri);
 
+        // Detect dialect from the root schema's $schema keyword before
+        // resolvePointer() may walk into a sub-schema.
+        $dialect = isset($jsonSchema->{'$schema'}) ? $jsonSchema->{'$schema'} : null;
+        $usesDollarId = !in_array($dialect, [DraftIdentifiers::DRAFT_3, DraftIdentifiers::DRAFT_4], true);
```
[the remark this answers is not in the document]
penyaskito: Good catch — fixed in 60f167c. The comparison now normalizes both sides: `rtrim($dialect, '#')` on the input and `withoutFragment()` on the `DraftIdentifiers` constants. Added 5 tests covering Draft-04 and Draft-07 with and without the trailing `#`, plus the no-`$schema` default.

tests/Uri/UriRetrieverTest.php
DannyvdSluijs: This was from a 9 years ago commit by a maintainer which is no longer active. I belief this to be a quick way to check for the expected result. These test now no longer confirm the `id` or `$id` value which would be even better. Could you add that an assertion for that?

DannyvdSluijs (commented):
Thanx for the issue report and the PR. I think it would be best to improve the tests ever so slightly.

tests/Uri/UriRetrieverTest.php
penyaskito: Good point — added value assertions ("id" or "$id" based on the active Draft) in 85bfc72.

DannyvdSluijs (approved):

penyaskito:
🚀 thanks!

Issue #925, opened 2026-07-17
Value-aware static applicability gating on constraints (follow-up to #920)


DannyvdSluijs:
## Description

#920 fixed unconditional constraint instantiation in `Draft06Constraint`, `Draft07Constraint`, and `Draft2019Constraint` by adding a `KEYWORD_SCHEMA_PROPERTIES` map in each dispatcher class, gating `Factory::createInstanceFor()` on **schema-property presence** before instantiating a constraint. That fix does not account for whether the keyword is even applicable to the **value's type**.

## Opportunity

Most individual constraint classes have a second internal guard beyond `property_exists($schema, ...)`, checking the value's type (`is_object`, `is_array`, `is_string`, `is_numeric`, etc.) before doing real work. For `Draft2019` specifically, 26 of 34 constraint classes (~76%) carry this second guard.

A schema that legally declares keywords applicable to multiple instance types (e.g. both `properties` and `pattern` on the same subschema) will, under the #920 fix, still instantiate a constraint that's guaranteed to no-op based on the value's type alone.

Example: validating a bare string against a schema declaring both `properties` (object-applicable) and `pattern` (string-applicable) still instantiates `PropertiesConstraint` needlessly.

## Proposed approach

Expose a static method on each constraint class (e.g. `public static function isApplicable($value, $schema): bool`) that encapsulates both the schema-property and value-type applicability checks, colocated with the class that actually has that domain knowledge. `Draft*Constraint::checkForKeyword()` would call this statically — avoiding instantiation entirely rather than just deferring it — before ever calling `Factory::createInstanceFor()`.

## Why this is a separate, larger change (not part of #920)

- **Blast radius**: ~98 constraint classes across `Draft06`/`Draft07`/`Draft2019` would each need this method added. For ~90% of them (where the keyword name already equals the schema property name), it's a one-line restatement of the guard that already exists in `check()`.
- **Backward-compatibility / breaking change**: `ConstraintInterface` is public, and consumers can substitute their own constraint classes via `Factory::setConstraintClass()`. Adding a *required* static method to the interface would break any consumer-authored constraint that doesn't implement it. This needs to be either an opt-in contract (checked via `method_exists()`, falling back to today's always-instantiate behavior) or shipped as an explicit breaking change with a major version bump.
[...]

PR #926, opened 2026-07-17
Avoid unconditional constraint instantiation in Draft06/07/2019 dispatch

### LINKED ISSUE #920: Draft06Constraint, Draft07Constraint, and Draft2019Constraint instantiate all constraint objects unconditionally per schema node
## Description

`Draft06Constraint::check()`, `Draft07Constraint::check()`, and `Draft2019Constraint::check()` each call `checkForKeyword()` for every registered keyword on every schema node, regardless of whether that keyword is present in the schema. Each call invokes `$this->factory->createInstanceFor($keyword)`, creating a fresh constraint object that immediately returns after a `property_exists` guard inside its own `check()`:

```php
// same pattern in all three draft constraint classes
protected function checkForKeyword(string $keyword, $value, $schema = null, ...): void
{
    $validator = $this->factory->createInstanceFor($keyword); // always instantiated
    $validator->check($value, $schema, $path, $i);           // immediately returns if keyword absent
}
```

Keyword counts per draft:
- `Draft06Constraint`: 27 unconditional instantiations per node
- `Draft07Constraint`: 28 unconditional instantiations per node
- `Draft2019Constraint`: 30 unconditional instantiations per node

For a typical schema with 3–5 keywords, 22–27 constraint objects are instantiated and discarded per schema node. In a deeply nested document this multiplies across every node visited.

## Affected Files

- `src/JsonSchema/Constraints/Drafts/Draft06/Draft06Constraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft07/Draft07Constraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft2019/Draft2019Constraint.php`

## Suggested Fix

Guard the `createInstanceFor` call with a `property_exists` check before instantiating:

```php
protected function checkForKeyword(string $keyword, $value, $schema = null, ?JsonPointer $path = null, $i = null): void
{
    if (!\property_exists($schema, $keyword)) {
        return;
    }

    $validator = $this->factory->createInstanceFor($keyword);
}
```

**Note:** the keyword name passed to `checkForKeyword` must match the schema property name for this guard to work. Most do (`'required'`, `'type'`, `'pattern'`, etc.), but two do not:
- `'ref'` maps to schema property `'$ref'`
- `'ifThenElse'` maps to schema property `'if'`

These two would need special handling or a keyword→schema-property mapping table.

DannyvdSluijs:
## Summary

- `Draft06Constraint`, `Draft07Constraint`, and `Draft2019Constraint` previously called `checkForKeyword()` for every registered keyword on every schema node, unconditionally instantiating a constraint object (via `Factory::createInstanceFor()`) even when the keyword was absent from the schema — the constraint's own `check()` would just immediately return via its internal `property_exists($schema, ...)` guard.
- `checkForKeyword()` now skips `createInstanceFor()` entirely when the relevant schema property is absent, using a small per-class keyword→property map to handle the three keywords whose name doesn't match the schema property they guard on (`ref`→`$ref`, `ifThenElse`→`if`, `content`→`contentMediaType`/`contentEncoding`).
- No behavior change: the guard exactly mirrors each constraint's own existing early-return.

Fixes #920.

## Verification

- Full official JSON-Schema-Test-Suite (`composer test`): identical failure set before and after (verified via git-stash A/B diff of failing test names) — zero regressions.
- `composer phpstan` (level 8): clean.
- `composer style-check`: clean.
- 16 new unit tests (`tests/Constraints/Drafts/Draft{06,07,2019}/`) assert `createInstanceFor` is skipped for absent keywords and invoked for present ones.
- Benchmarked on a representative nested schema: constraint instantiations for one recursive validation dropped from 40 to 9; isolated dispatch throughput went from ~267 to ~864 checks/sec (~3.2x). No measurable change in peak memory — the avoided objects were already being freed immediately via refcounting, so this is a CPU/allocator-churn win rather than a footprint win.

## Follow-up

A related, larger idea came up during review: gating could go further by also checking the *value's* type (not just schema-property presence), since ~76% of Draft2019 constraints already guard on value type internally. That's a bigger, potentially breaking change (touches ~98 constraint classes, would need an opt-in addition to `ConstraintInterface` to stay backward compatible for consumers using `Factory::setConstraintClass()`), so it's tracked separately in #925 rather than folded into this PR.

## Test plan

- [x] `composer test`
- [x] `composer phpstan`
- [x] `composer style-check`

### DIFF
--- a/src/JsonSchema/Constraints/Drafts/Draft06/Draft06Constraint.php
+++ b/src/JsonSchema/Constraints/Drafts/Draft06/Draft06Constraint.php
@@ -66,6 +66,10 @@ class Draft06Constraint extends Constraint
         $this->checkForKeyword('pattern', $value, $schema, $path, $i);
     }
 
+    private const KEYWORD_SCHEMA_PROPERTIES = [
+        'ref' => ['$ref'],
+    ];
+
     /**
      * @param mixed $value
      * @param mixed $schema
@@ -73,9 +77,15 @@ class Draft06Constraint extends Constraint
      */
     protected function checkForKeyword(string $keyword, $value, $schema = null, ?JsonPointer $path = null, $i = null): void
     {
-        $validator = $this->factory->createInstanceFor($keyword);
-        $validator->check($value, $schema, $path, $i);
+        foreach (self::KEYWORD_SCHEMA_PROPERTIES[$keyword] ?? [$keyword] as $property) {
+            if (property_exists($schema, $property)) {
+                $validator = $this->factory->createInstanceFor($keyword);
+                $validator->check($value, $schema, $path, $i);
 
-        $this->addErrors($validator->getErrors());
+                $this->addErrors($validator->getErrors());
+
+                return;
+            }
+        }
     }
 }
--- a/src/JsonSchema/Constraints/Drafts/Draft07/Draft07Constraint.php
+++ b/src/JsonSchema/Constraints/Drafts/Draft07/Draft07Constraint.php
@@ -68,6 +68,12 @@ class Draft07Constraint extends Constraint
         $this->checkForKeyword('content', $value, $schema, $path, $i);
     }
 
+    private const KEYWORD_SCHEMA_PROPERTIES = [
+        'ref' => ['$ref'],
+        'ifThenElse' => ['if'],
+        'content' => ['contentMediaType', 'contentEncoding'],
+    ];
+
     /**
      * @param mixed $value
      * @param mixed $schema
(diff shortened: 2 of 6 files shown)


src/JsonSchema/Constraints/Drafts/Draft06/Draft06Constraint.php
```diff
@@ -66,16 +66,26 @@ public function check(&$value, $schema = null, ?JsonPointer $path = null, $i = n
         $this->checkForKeyword('pattern', $value, $schema, $path, $i);
     }
 
+    private const KEYWORD_SCHEMA_PROPERTIES = [
```
DannyvdSluijs: **nitpick**: This is one exception we can handle inline instead of making this a class constant  without any context.

src/JsonSchema/Constraints/Drafts/Draft06/Draft06Constraint.php
```diff
@@ -66,16 +66,26 @@ public function check(&$value, $schema = null, ?JsonPointer $path = null, $i = n
...
      * @param mixed $value
      * @param mixed $schema
      * @param mixed $i
      */
     protected function checkForKeyword(string $keyword, $value, $schema = null, ?JsonPointer $path = null, $i = null): void
     {
-        $validator = $this->factory->createInstanceFor($keyword);
-        $validator->check($value, $schema, $path, $i);
+        foreach (self::KEYWORD_SCHEMA_PROPERTIES[$keyword] ?? [$keyword] as $property) {
+            if (property_exists($schema, $property)) {
```
DannyvdSluijs: **nitpick**: We can flip the `if`-statement and apply early continue

Issue #907, opened 2026-05-13
Add support for `unevaluatedProperties` for more efficient useage of `allOf` schemas


fulstadev:
When you have an `allOf` JSON schema of two objects with the [unevaluatedProperties](https://json-schema.org/draft/2020-12/json-schema-core#name-unevaluatedproperties) property, let's say: 

```
{
    "type": "object",
    "unevaluatedProperties" : false,
    "allOf": [
        {
            "type": "object",
            "properties": {
                "hello" : {
                    "type" : "string"
                }
            },
            "required" : ["hello"]
        },
        {
            "type": "object",
            "properties": {
                "world": {
                    "type" : "string"
                }
            },
            "required" : ["world"]
        }
    ]
}
```

The validator should treat the validation equivalently to:

```
{
    "type": "object",
    "properties": {
        "hello" : {
            "type" : "string"
        },
        "world": {
            "type" : "string"
        }
    },
    "additionalProperties" : false,
    "required" : ["hello","world"]
}
```

Some tests on my side seem to ignore it. If you submit an object with additional properties (other than `hello` and `world`), validation does not fail as expected. 

I suppose that the `unevaluatedProperties` thing is from a more recent draft, but it has been introduced for a [reason](https://stackoverflow.com/questions/79247662/exact-definition-of-unevaluatedproperties-in-jsonschema) and makes the usage of the very core `allOf` feature much more efficient, as it effectively allows you to reuse object schemas defined throughout the project in a much stricter way.

Do you think that it's possible to introduce this keyword ?

DannyvdSluijs:
Hi @fulstadev 

`unevaluatedProperties ` was added in `Draft-2019-09` which is a draft currently not yet support (nor is `Draft-2020-12`). In #885 I'm working on adding initial support for `Draft-2019-09` but I'm planning to skip `unevaluatedProperties` due to the complexity.

You're free to work on a PR that targets either main once the PR is merged or target the PR and we can see if it can get merged.  It might be quite a large amount of work due to the fact currently there is no keeping track of properties that have been evaluated.

tomatotomata:
[...]

PR #931, opened 2026-08-06
feat: support unevaluatedProperties in Draft 2019-09

tomatotomata:
## Summary

This adds a focused proof of concept for `unevaluatedProperties` in Draft 2019-09, following the scope discussed in #907.

The implementation derives evaluated property names from `properties`, `patternProperties`, and direct `allOf` branches. That covers the motivating composition while keeping the first patch small. When `unevaluatedProperties` is a schema, each remaining property is validated against it. When it is `false`, the validator reports the property with a dedicated constraint error.

The validator does not yet carry annotations across every applicator, so this intentionally does not claim complete Draft 2019-09 semantics for `anyOf`, `oneOf`, conditionals, or referenced schemas. I would treat this as a proof of concept for the project to shape further.

## Tests

- `php -d extension=mbstring -d extension=openssl vendor/bin/phpunit --filter UnevaluatedPropertiesTest --testdox` passes locally (2 assertions-backed cases, 2 expected skips for the assoc/type-cast variants).
- PHPStan passes for the new constraint.
- `git diff --check` passes.
- The pre-existing full-suite run reaches the suite but has an unrelated Windows path-separator failure; the original CI run also exposed the missing Draft 2019 strict-mode flag in this new test, which is fixed in the follow-up commit.

This patch was prepared with AI assistance and then reviewed and tested in this checkout. The repository's contribution notice asks contributors to have authored 100% of the content, so please let me know if the project needs a different treatment of that requirement before considering the patch.

Updates #907


DannyvdSluijs:
@tomatotomata ill be back from
Holidays next week. Hoping I can provide a review late next week. Early quick scan already enabled me to start the workflows which have been picked up by you already. Thnx.

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
```diff
@@ -0,0 +1,112 @@
...
+
+namespace JsonSchema\Constraints\Drafts\Draft2019;
+
+use JsonSchema\ConstraintError;
+use JsonSchema\Constraints\ConstraintInterface;
+use JsonSchema\Entity\ErrorBagProxy;
+use JsonSchema\Entity\JsonPointer;
+
+/**
+ * Proof-of-concept support for unevaluatedProperties.
```
DannyvdSluijs, line 13: **suggestion**: Before merging we should remove the proof on concept comment.

```suggestion
```

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * The current validator does not carry annotations between applicators, so this
+ * first implementation derives the evaluated property names from properties,
+ * patternProperties, and allOf branches in the current schema.
```
DannyvdSluijs, line 17: suggestion: The comment doesn't need to inform about this being the first implementation
```suggestion
 * The current validator does not carry annotations between applicators, so this
 * constraint derives the evaluated property names from properties,
 * patternProperties, and allOf branches in the current schema.
```

tests/Constraints/UnevaluatedPropertiesTest.php
DannyvdSluijs, line 1: nitpick: I think this file would belong in `test/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraintTest.php`

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
```diff
@@ -0,0 +1,112 @@
...
+     * @param object $value
+     *
+     * @return array<int, string>
+     */
+    private function collectEvaluatedProperties($schema, object $value): array
+    {
+        if (!is_object($schema)) {
+            return [];
+        }
+
```
DannyvdSluijs, line 78: suggestion: Since `$object` was already verified as being of type object on line 43 we can safely narrow the type of the function to `object` and skip the check as well as the doc types since we use the native typing. This requires a additional check on line 91 where we do a recursive call.
```suggestion
    /**
     * @return array<int, string>
     */
    private function collectEvaluatedProperties(object $schema, object $value): array
    {
```

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
DannyvdSluijs, line 1: **issue**: With this keyword now supported we should alos enable the test in https://github.com/jsonrainbow/json-schema/blob/bbe268a154d7b4d0418dfb828ab1efd5c9c1dd13/tests/JsonSchemaTestSuiteTest.php#L206-L245

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
```diff
@@ -0,0 +1,112 @@
...
+    public function check(&$value, $schema = null, ?JsonPointer $path = null, $i = null): void
+    {
+        if (!is_object($schema) || !property_exists($schema, 'unevaluatedProperties') || !is_object($value)) {
+            return;
+        }
+
+        if ($schema->unevaluatedProperties === true) {
+            return;
+        }
+
```
DannyvdSluijs, line 41: issue: This should also evaluate the availability and value of `additionalProperties`, see the [tests](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/6648e8194c69697b2e1a15fe76a06a480b183a51/tests/draft2019-09/unevaluatedProperties.json#L134-L162) in the test suite and the [spec](https://json-schema.org/draft/2019-09/json-schema-core#rfc.section.9.3.2.4).

src/JsonSchema/Constraints/Drafts/Draft2019/UnevaluatedPropertiesConstraint.php
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                }
+            }
+        }
+
+        if (isset($schema->allOf) && is_array($schema->allOf)) {
+            foreach ($schema->allOf as $branch) {
+                $evaluated = array_merge($evaluated, $this->collectEvaluatedProperties($branch, $value));
+            }
+        }
```
DannyvdSluijs, line 100: **question**: This should include (either now or at a later state) `anyOf`, `oneOf`, `if/then/else` and more keywords right? Curious to learn what would be your suggested approach?

DannyvdSluijs (commented):
Really nice work. I didn't think about this solution direction. Refreshing to see your approach.
I did add some review comments could you perhaps take a look and see if you can contribute?

tomatotomata:
I pushed the reviewer-requested follow-up in `5e0a3b3`.

The collector is now branch-aware for `$ref`, `allOf`, `anyOf`, `oneOf`, conditionals, `dependentSchemas`, and `additionalProperties`, and the focused test is under the Draft 2019 namespace. I also kept the existing `$ref` sibling behavior and made nested error paths relative to the current property.

Validation on the current head:
- Draft 2019 focused tests: 16 tests, 27 assertions, with 8 expected skips for unsupported type-cast cases
- official unevaluatedProperties conformance slice: 125 tests, 125 assertions
- focused PHPStan: clean
- `git diff --check`: clean

The recursive reference case remains explicitly skipped because this branch does not implement `$recursiveRef` yet. Let me know what you think.

tomatotomata:
Thanks, I worked through the remaining points together so the implementation and coverage line up.

- Removed the proof-of-concept wording and moved the focused test to the Draft 2019-09 test directory.
- Narrowed the collector to object schemas and guarded recursive collection where boolean schemas are valid but cannot contribute property names.
- Added `additionalProperties` to the evaluated-property collection and extended coverage across `allOf`, `anyOf`, `oneOf`, `if`/`then`, and `dependentSchemas`.
- Enabled the three single-cyclic-reference cases in the official suite. The `$recursiveRef` case remains skipped because this validator does not support recursive references yet.

The focused constraint tests pass (18 tests, 30 assertions; 9 expected mode skips), and the official `unevaluatedProperties` subset passes (128 tests). PHPStan on the changed source and the targeted style check also pass. The full PHPUnit suite is not green on my PHP 8.5.8 Windows setup (149 errors and 204 failures, including missing `json-schema-test-suite/remotes/node` fixtures), so I’m only claiming the focused results above.

I pushed these changes as `6dbce11`. Let me know if you’d like any part of the collector approach adjusted.

Issue #918, opened 2026-06-08
RequiredConstraint::incrementPath is duplicated across drafts instead of living on JsonPointer


DannyvdSluijs:
## Description

`incrementPath` is a path manipulation helper that is copy-pasted into the `RequiredConstraint` of every draft (Draft06, Draft07, Draft2019) because those constraints implement `ConstraintInterface` directly rather than extending `Constraint`, and therefore don't inherit the `incrementPath` method available on the base `Constraint` class.

## Affected Files

- `src/JsonSchema/Constraints/Drafts/Draft06/RequiredConstraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft07/RequiredConstraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft2019/RequiredConstraint.php`

The Draft2019 version has a `@todo` comment acknowledging this:

```php
/**
 * @todo refactor as this was only copied from UndefinedConstraint
 */
protected function incrementPath(?JsonPointer $path, $i): JsonPointer
{
    $path = $path ?? new JsonPointer('');

    if ($i === null || $i === '') {
        return $path;
    }

    return $path->withPropertyPaths(array_merge($path->getPropertyPaths(), [$i]));
}
```

## Suggested Fix

Move `incrementPath` onto `JsonPointer` itself as an instance method, since it exclusively operates on a `JsonPointer`. All three `RequiredConstraint` files (and any other constraint that needs it) could then call `$path->withAppendedPath($i)` without needing to duplicate the helper.

Alternatively, a static utility method on `JsonPointer` or a dedicated path utility class would also centralise the logic without touching the constraint hierarchy.

PR #942, opened 2026-08-30
Report additional property errors at the property path

### LINKED ISSUE #919: AdditionalPropertiesConstraint reports incorrect error path when additionalProperties is a schema
## Description

When `additionalProperties` is a schema object (rather than `false`), each additional property value is validated against that schema — but the path is not incremented with the property key. Any validation errors from the sub-schema check are reported at the parent object's path instead of the specific property's path.

## Affected Files

- `src/JsonSchema/Constraints/Drafts/Draft06/AdditionalPropertiesConstraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft07/AdditionalPropertiesConstraint.php`
- `src/JsonSchema/Constraints/Drafts/Draft2019/AdditionalPropertiesConstraint.php`

All three have a `@todo` comment acknowledging this:

```php
$schemaConstraint->check($additionalPropertiesValue, $schema->additionalProperties, $path, $i); // @todo increment path
```

## Steps to Reproduce

```json
{
  "properties": { "foo": {} },
  "additionalProperties": { "type": "integer" }
}
```

Validating `{"foo": "ok", "bar": "not-an-integer"}` produces a type error at path `` (root) instead of `/bar`.

## Expected Behaviour

The error path should point to the specific additional property that failed — e.g. `/bar` — not to the parent object.

## Suggested Fix

Increment the path with the property key before passing it to the sub-schema check:

```php
$schemaConstraint->check($additionalPropertiesValue, $schema->additionalProperties, $path->withAppendedPath($key), $i);
```

This depends on a path utility being available on `JsonPointer` — see issue #918 which tracks the same missing utility in `RequiredConstraint`.

Amoifr:
Closes #919 and #918.

Two corrections to the diagnosis in #919 came out of reproducing it, so the change is not the one-liner the issue suggests.

**The suggested fix could not compile.** `JsonPointer` has no `withAppendedPath()`. The equivalent lives as a `protected incrementPath()` copy-pasted into the `RequiredConstraint` of all three drafts, which is exactly what #918 asks to move onto `JsonPointer`. Fixing #919 on its own would have meant writing that helper a fourth, fifth and sixth time, so doing #918 first makes this the smaller change rather than a wider one. `JsonPointer::withAppendedPath()` is added, the three duplicates are gone.

**Incrementing the path alone changes nothing.** `AdditionalPropertiesConstraint` never propagates the sub-schema errors: it calls `check()`, reads `isValid()` as a boolean, and throws the errors away. So the reported symptom does not happen either. Reproducing the example from the issue in `CHECK_MODE_STRICT`, which is where the draft constraints are actually used:

```
before   pointer=''      constraint=additionalProp
after    pointer='/bar'  constraint=type
```

There was no type error at the root; there was no type error at all. The reason for the failure was discarded and replaced by an `additionalProperties` error pointing at the parent.

That second error is wrong on its own terms. When `additionalProperties` is a schema, the property **is** permitted, so "additional property found" is not what happened. `PropertiesConstraint` already does the right thing three files away, ending its loop with `$this->addErrors($schemaConstraint->getErrors())`. This aligns `AdditionalPropertiesConstraint` with it, and keeps the `additionalProperties` error for the `additionalProperties: false` case where it belongs.

### Verification

Nine tests added, three per draft: the sub-schema error lands at `/bar`, a property matching the sub-schema reports nothing, and `additionalProperties: false` still reports the property. Each fails without the change.

Full suite green at 3169 tests, including the official JSON-Schema-Test-Suite for draft06, draft07 and draft2019 in strict mode, which arbitrates the validity verdicts. PHPStan clean.

One caveat on style: `php-cs-fixer` pins itself to PHP 8.0 and I only have 8.5 here, so with `PHP_CS_FIXER_IGNORE_ENV` it proposes changes on 120 files, all of them the same bogus `&$value` to `& $value` rewrite. The only thing it says about the files touched here is that rewrite, on a signature line I did not modify, so I left it alone rather than commit noise.

### DIFF
--- a/tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
+++ b/tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JsonSchema\Tests\Constraints\Draft06;
+
+use JsonSchema\Constraints\Drafts\Draft06\AdditionalPropertiesConstraint;
+use JsonSchema\Tests\Constraints\VeryBaseTestCase;
+
+class AdditionalPropertiesConstraintTest extends VeryBaseTestCase
+{
+    public function testSubSchemaErrorsAreReportedAtThePropertyPath(): void
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": {"type": "integer"}}');
+        $value = json_decode('{"foo": "ok", "bar": "not-an-integer"}');
+
+        $constraint->check($value, $schema);
+
+        $errors = $constraint->getErrors();
+        $this->assertCount(1, $errors);
+        $this->assertSame('/bar', $errors[0]['pointer']);
+        $this->assertSame('type', $errors[0]['constraint']['name']);
+    }
+
+    public function testAPropertyMatchingTheSubSchemaIsNotReported(): void
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"additionalProperties": {"type": "integer"}}');
+        $value = json_decode('{"bar": 1}');
+
+        $constraint->check($value, $schema);
+
+        $this->assertSame([], $constraint->getErrors());
+    }
+
+    public function testAdditionalPropertiesFalseStillReportsTheProperty(): void
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": false}');
+        $value = json_decode('{"foo": "ok", "bar": 1}');
+
+        $constraint->check($value, $schema);
+
+        $errors = $constraint->getErrors();
+        $this->assertCount(1, $errors);
+        $this->assertSame('additionalProp', $errors[0]['constraint']['name']);
+    }
+}
--- a/tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
+++ b/tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
@@ -44,6 +44,7 @@ class AdditionalPropertiesConstraintTest extends VeryBaseTestCase
 
         $errors = $constraint->getErrors();
         $this->assertCount(1, $errors);
+        $this->assertSame('/bar', $errors[0]['pointer']);
         $this->assertSame('additionalProp', $errors[0]['constraint']['name']);
     }
 }
--- a/tests/Constraints/Draft07/AdditionalPropertiesConstraintTest.php
+++ b/tests/Constraints/Draft07/AdditionalPropertiesConstraintTest.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JsonSchema\Tests\Constraints\Draft07;
+
+use JsonSchema\Constraints\Drafts\Draft07\AdditionalPropertiesConstraint;
+use JsonSchema\Tests\Constraints\VeryBaseTestCase;
+
+class AdditionalPropertiesConstraintTest extends VeryBaseTestCase
+{
+    public function testSubSchemaErrorsAreReportedAtThePropertyPath(): void
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": {"type": "integer"}}');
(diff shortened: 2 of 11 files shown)


tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": false}');
+        $value = json_decode('{"foo": "ok", "bar": 1}');
+
+        $constraint->check($value, $schema);
+
+        $errors = $constraint->getErrors();
+        $this->assertCount(1, $errors);
+        $this->assertSame('additionalProp', $errors[0]['constraint']['name']);
```
DannyvdSluijs, line 48: **nitpick**: Can you include a assertion that the pointer is `/bar`?

tests/Constraints/Draft07/AdditionalPropertiesConstraintTest.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": false}');
+        $value = json_decode('{"foo": "ok", "bar": 1}');
+
+        $constraint->check($value, $schema);
+
+        $errors = $constraint->getErrors();
+        $this->assertCount(1, $errors);
+        $this->assertSame('additionalProp', $errors[0]['constraint']['name']);
```
DannyvdSluijs, line 48: **nitpick**: Same, can you include a assertion that the pointer is `/bar`?

tests/Constraints/Draft2019/AdditionalPropertiesConstraintTest.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $constraint = new AdditionalPropertiesConstraint();
+        $schema = json_decode('{"properties": {"foo": {}}, "additionalProperties": false}');
+        $value = json_decode('{"foo": "ok", "bar": 1}');
+
+        $constraint->check($value, $schema);
+
+        $errors = $constraint->getErrors();
+        $this->assertCount(1, $errors);
+        $this->assertSame('additionalProp', $errors[0]['constraint']['name']);
```
DannyvdSluijs, line 48: **nitpick**: Here as well, can you include a assertion that the pointer is `/bar`?

DannyvdSluijs (changes requested):
Nice progress 👍🏼 If you could just add a few assertion on the pointer that would be great.

tests/Constraints/Draft06/AdditionalPropertiesConstraintTest.php
Amoifr: Added, and the assertion promptly failed, which turns your nitpick into a bug report: the pointer was `''`, not `/bar`.

This PR fixed the sub-schema branch (`additionalProperties` as an object) to report at the property path, but left the `additionalProperties: false` branch emitting on the parent `$path`, so the two halves of the same method disagreed. Fixed in e3ae15c for all three drafts:

```php
$this->addError(ConstraintError::ADDITIONAL_PROPERTIES(), ($path ?? new JsonPointer(''))->withAppendedPath($key), ['found' => $key]);
```

One behaviour change to flag, since you may want to veto it: the two `error message contains property name not value` cases in `AdditionalPropertiesTest` now expect `property => extraProp` and `pointer => /extraProp` instead of empty strings. Those cases assert on the message, so the pointer was incidental there, but it is visible to anyone reading `pointer`. Happy to revert to reporting at the parent and assert `''` in these three tests instead, just say so.

Full suite green either way: 3169 tests, 9821 assertions, PHPStan clean.

tests/Constraints/Draft07/AdditionalPropertiesConstraintTest.php
Amoifr: Done here too. See the Draft06 thread for why this needed a code change rather than just the assertion: the pointer really was `''` for the `additionalProperties: false` branch.

tests/Constraints/Draft2019/AdditionalPropertiesConstraintTest.php
Amoifr: Done here too. See the Draft06 thread for why this needed a code change rather than just the assertion: the pointer really was `''` for the `additionalProperties: false` branch.

DannyvdSluijs (approved):

Issue #953, opened 2026-09-13
JsonSchemaTestSuiteTest silently runs zero tests since the suite added tests/v1/


Amoifr:
### What happens

`JsonSchemaTestSuiteTest` yields no test at all. Its data provider throws while building the data sets:

```
1) Warning
The data provider specified for JsonSchema\Tests\JsonSchemaTestSuiteTest::testTestCaseValidatesCorrectly is invalid.
InvalidArgumentException: v1 is not a valid constraint name.
/src/JsonSchema/DraftIdentifiers.php:53
/tests/JsonSchemaTestSuiteTest.php:98

WARNINGS!
Tests: 1, Assertions: 0, Warnings: 1.
```

PHPUnit 8.5 treats an invalid data provider as a warning, not a failure, and no `failOnWarning` is set in `phpunit.xml.dist`. The command exits **0**, so CI stays green while the entire official test suite is skipped.

### Why

`json-schema/json-schema-test-suite` is required as `dev-main`, and `composer.lock` is gitignored while CI runs `composer update`, so every run picks up the current upstream tip. The suite now ships a `tests/v1/` directory, and the provider only skips three drafts:

```php
$skippedDrafts = ['draft2020-12', 'draft-next', 'latest'];
```

`v1` therefore reaches `DraftIdentifiers::fromConstraintName('v1')`, which knows drafts 3, 4, 6, 7, 2019-09 and 2020-12 only.

Reproduced on `main` with `composer install` on PHP 8.4.

### One caveat before fixing it

Adding `'v1'` to `$skippedDrafts` does restore the suite, but it will not come back green. Once the class actually runs, 361 cases fail, because the skip list in `shouldNotYieldTest()` was tuned against an older revision of the upstream suite and has been drifting unnoticed for as long as the provider has been throwing. Worth knowing before the one line change turns a quiet CI into a loud one.

Deciding what `v1` should mean for this library is a separate question: it is the draft the suite now publishes as the stable release, so it may eventually deserve a `DraftIdentifiers` entry rather than a skip.

Pinning the suite to a tag, or committing `composer.lock`, would also keep a new upstream directory from muting the whole test class again.

### Why it matters

The failure mode is silent in both directions: the suite does not report a failure, it reports having nothing to run, and the build passes. A green CI does not currently mean the drafts are being exercised.

Thanks for maintaining this library, and sorry to be the bearer of a slightly awkward discovery. 🙂

PR #954, opened 2026-09-13
Only treat schema bearing positions as identifier locations

### LINKED ISSUE #949: Schema ids are discovered in annotation data and unknown keywords
## Description

`SchemaStorage::scanForSubschemas()` walks every object it reaches and treats any `id`/`$id` it finds as a schema identifier. It has no notion of which keywords are *schema-valued*, so identifiers are picked up from positions that hold plain data.

## Reproduce

```php
$storage = new SchemaStorage();
$storage->addSchema('http://example.org/root/', json_decode('
    {"default": {"id": "leaked.json", "type": "string"}}
'));
// registers http://example.org/root/leaked.json
```

The same happens under `examples` and under any unknown keyword:

```php
{"someUnknownKeyword": {"id": "leaked.json", "type": "string"}}
// registers http://example.org/root/leaked.json
```

## Cause

`enum` and `const` are special-cased by name, which is the only place the scanner distinguishes data from schemas. Everything else — `default`, `examples`, unknown keywords — is walked as if it were a schema. There is a leftover marker for this in the code history: a `// $id in unknown keywords is not valid` comment alongside an empty `in_array($propertyName, [])`, i.e. the check was stubbed out and never filled in.

A registered identifier that came from annotation data can shadow a real one, and a `$ref` to that URI then resolves to the data rather than to the intended schema.

## Related skipped tests

`JsonSchemaTestSuiteTest::shouldNotYieldTest()` skips these, which are the same problem:

- `[draft6/unknownKeyword.json]: $id inside an unknown keyword is not a real identifier: type matches second anyOf, which has a real schema in it`
- `[draft6/unknownKeyword.json]: $id inside an unknown keyword is not a real identifier: type matches non-schema in third anyOf`
- `[draft7/unknownKeyword.json]` — both of the above

## Suggested approach

Replace the name-based `enum`/`const` special case with a keyword classification the scan carries as it descends, distinguishing:

- schema-valued keywords (`items`, `additionalProperties`, `not`, `if`/`then`/`else`, …)
- maps of schemas (`properties`, `patternProperties`, `definitions`, `$defs`, `dependencies`, `dependentSchemas`) — already listed as `SCHEMA_MAP_KEYWORDS`
- arrays of schemas (`allOf`, `anyOf`, `oneOf`, tuple `items`)
- everything else, which is data and must not be descended into for identifiers

The classification is draft-dependent, so it likely belongs near `DraftIdentifiers`/`Factory` rather than hardcoded in `SchemaStorage`. Removing the four `unknownKeyword.json` skips is the acceptance criterion. Re-run the suite with `shouldNotYieldTest()` disabled before and after to catch both regressions and newly passing cases.

## Context

Raised in review on [#947](https://github.com/jsonrainbow/json-schema/pull/947). It reproduces identically on `main`, so it is not a regression from that PR, and fixing it properly is a larger design change than that PR's scope.

Amoifr:
Fixes #949

### Problem

`SchemaStorage` walks a document looking for `id` / `$id` so it can register subschemas under their canonical URI. It descends into every member it meets, so an `$id` written in a position that holds *data* rather than a schema gets registered as though it identified a schema. When that value collides with a real `$id` elsewhere in the document, the fake one can win, and a `$ref` to that URI then resolves to the wrong schema.

That is exactly what `unknownKeyword.json` in the official suite pins down: an `$id` buried under an unknown keyword must not shadow the real one.

### Cause

`scanMembersForSubschemas()` skipped `enum` and `const` by name. That covers two data positions out of many: `default`, `examples` and any unknown keyword were still descended into.

### Fix

Instead of listing the positions to avoid, which is open ended, the scanner now only descends into positions that actually hold schemas. Two classification lists sit beside the existing `SCHEMA_MAP_KEYWORDS`:

- `SCHEMA_KEYWORDS`, whose value is a schema (`not`, `if`, `additionalProperties`, and so on)
- `SCHEMA_ARRAY_KEYWORDS`, whose value is a list of schemas (`allOf`, `anyOf`, `prefixItems`, and so on)

Everything else is data, and an `id` inside it identifies nothing.

Members of a schema map (`properties`, `$defs`, `definitions`, ...) remain schemas whatever their name, which is what the existing `$parentProperty` check already guaranteed: a subschema legitimately named `enum` is still traversed.

The classification is deliberately draft agnostic, since `SchemaStorage` is never told which dialect it is reading. A keyword counts if it holds schemas in any draft. That is also the safer direction: treating a keyword unknown to the current draft as a schema position is simply the behaviour every keyword had before.

### Tests

Six cases added to `subschemaRegistrationProvider()`. Four fail without the source change (`default`, `examples`, an unknown keyword holding a list, an unknown keyword holding a map). The remaining two guard against over narrowing in the other direction, checking that an `id` under `not` and under an `items` tuple is still registered.

The six now obsolete `unknownKeyword.json` entries in `shouldNotYieldTest()` are removed.

### Notes

`expandRefs()` carries the same `enum` / `const` special case. I left it untouched: the fixture does not exercise it, and measuring confirmed the scanner change alone is sufficient. Glad to align the two walkers in a follow up if you would prefer they agree.

One caveat on how far I could verify this. `JsonSchemaTestSuiteTest` currently yields no tests at all, because the upstream suite added a `tests/v1/` directory that `DraftIdentifiers::fromConstraintName()` rejects, and PHPUnit reports that as a warning while still exiting 0. I opened #953 about it. To exercise the `unknownKeyword` cases here I skipped `v1` locally, which is **not** part of this PR. With the suite unblocked that way, this change takes the failure count from 364 down to 354, repairing 6 suite cases plus the 4 new unit cases, and breaking none.

PHPStan reports no errors. Thanks for the library, and for the very readable `SchemaStorage`, which made this one pleasant to chase. 🙂


DannyvdSluijs:
Hello @Amoifr 

Thanks for the input on #953, I've just merged #956 which fixes the warning in the PHPUnit workflow. Skipping the `v1` folder and fixing the broken tests already in main. Downside is this PR now has conflicts. Could you rebase/recreate the PR?

Amoifr:
Thanks for the quick fix in #956, and for the heads-up! Rebased on main. One thing worth mentioning: the two `unknownKeyword.json` skips this PR removed are now listed under `optional/`, so I removed those entries instead. Without that, the fix would have had no test covering it. The 6 cases pass now, and they fail if I revert the change.

